PR #55, opened 2026-04-29
v0 testnet: cross-chain JINN claim pipeline + DR-2026-04-29 OLAS disposition

ritsuKai2000:
## Summary

This PR encompasses the full v0 testnet build-out work on `jinn-mono/jinn-mono-1bo`: contract implementation, proxy-deployable activity checker, parameterized `JINN.sol` ctor (testnet `tJINN` / mainnet `JINN`), cross-chain JINN claim pipeline (canonical + mock messenger), storage-layout drift CI guard, security review (Slither + threat model + Foundry invariants), and the just-completed v0 testnet burn-in.

The most recent three commits (added during the 2026-04-29 burn-in) are:

- **`444b0720`** — Decode Safe v1.3 inner reverts so GS013 wraps surface their real causes. Closes `jinn-mono-j48`. Adds `safe-revert.ts` with a known-error table covering `ClaimRegistry` (`JobAlreadyClaimed`, `IneligibleToClaim`, etc.) and `JinnRouterV2` (`RequestNotFound`, `AlreadyClaimed`, `DeliveryAlreadyClaimed`). `tx-retry.ts` marks decoded permanent inner errors as non-recoverable so the daemon stops burning the 6-attempt retry budget on doomed claims.
- **`691f1fad`** — Wire `JinnClaimLoop` into `main.ts` so the v0 cross-chain pipeline actually fires. Includes RPC eventual-consistency retry on L2 receipt fetches (closes `jinn-mono-zn6`) and accurate reward-claim log wording (closes `jinn-mono-6go`).
- **`a07d9caf`** — Close Locked Decisions §10 (OLAS disposition) per `DR-2026-04-29-olas-disposition-25-burn-self-bond.md`: standard mode = 25% stOLAS depositor yield / 75% burn / 0% operator; self-bond = 100% to operator-as-curating-agent. Includes the one-shot `apply-dr-2026-04-29-split.mjs` admin script and updates to `docs/planning/2026-04-jinn-mvi-on-olas.md`. GitHub Discussion #44 also updated to match.

## Burn-in evidence (v0 testnet, Sepolia + Base Sepolia)

Daemon ran continuously from 2026-04-29 15:58 UTC (post-fix) onward:

| Metric | Value |
|-|-|
| JinnClaimLoop cycles | 14 emit + 14 claim |
| Cycle success rate | **100%** |
| `jinn-claim` tick_errors | **0** |
| Receipt-not-found errors | **0** (after fix) |
[...]

PR #56, opened 2026-04-29
docs(operator-testnet): document pre-stake distributor whitelist + rewards-pool gotchas

kupermind:
## Summary

- Adds a **Pre-stake whitelist** caveat: operators must be in the distributor's top-level `mapCuratingAgents` before first `stake()`, otherwise `reStake()` after the inevitable first-window eviction reverts with `UnauthorizedAccount`. Surfaces the contract addresses and the exact `setCuratingAgents` call the protocol team needs to make.
- Adds a **Staking proxy rewards** caveat clarifying that `jinn doctor`'s "distributor holds N JINN" report does *not* reflect the staking proxy's own `availableRewards`, and either can be empty independently.
- Adds two **Troubleshooting** entries decoding the exact selectors (`0x32b2baa3` / `0xafb0be33`) operators see in the daemon log when these fire.

## Why

Both failures hit a fresh testnet bootstrap this week. Each took roughly an hour of trace-decoding to identify because the daemon surfaces only the raw selector. The `UnauthorizedAccount` case is in fact already named in the bootstrap source (`client/src/earning/bootstrap.ts:818-822`), but that context never reaches the operator-facing runbook.

## Test plan

- [ ] Render preview: bullet renders inside the existing "Testnet-specific caveats" list, troubleshooting entries match the existing \`### Symptom\` style.
- [ ] Cross-check: addresses (\`0x20951FBD…CEf81\` distributor, \`0xf358B5C1…ab3b\` staking proxy) match \`client/deployments/deployment-stolas-l2-baseSepolia-fast.json\`.
- [ ] Cross-check: error selectors recover correctly via \`cast keccak\` for \`UnauthorizedAccount(address)\` (matches \`client/src/errors/unauthorized-account.ts\`) and \`NoRewardsAvailable()\` (matches \`StakingBase.sol:158\`).


kupermind:
Closing — noticed #52 shipped both the stOLAS `reStake` contract fix and an equivalent runbook update while this was in flight, so the "Pre-stake whitelist" half of this PR is now obsolete (upstream correctly states whitelisting is no longer required). The remaining half (separating the staking-proxy `availableRewards` from the distributor's JINN runway) is still accurate today, but with #55 migrating the on-chain stack to the v0 protocol, the specific addresses I cited will likely change. Better to redo the runbook delta against the post-migration stack if that gap persists.

Thanks for the surrounding fixes — service 36 recovered cleanly via the same-operator path after #52 deployed.

PR #84, opened 2026-05-04
fix(client): make onboarding faucet loop reach the bootstrap target

oaksprout:
## Summary

Fresh-fleet onboarding could never auto-complete because the CDP faucet drip cap (60 × ~0.0001 ETH = 0.006 ETH) is below the effective bootstrap target (0.010 ETH on first fleet, from \`minEoaGasEth: 0.005 ETH\` × \`STANDARD_MASTER_BOOTSTRAP_MULTIPLIER\` of 2). The SPA's "Fund from faucet" button also became permanently disabled after a partial drip round, so the user couldn't even retry without restarting the daemon.

This PR fixes both:

- **`bootstrap.ts`**: `MAX_FAUCET_ITERS` is now derived from \`(target − balance) × 2\` with a 60-drip floor and 500-drip ceiling. The loop is mathematically able to reach the actual target; rate-limit / error responses still exit early.
- **`setup-endpoints.ts`**: same dynamic cap inside \`POST /v1/setup/drip\`. Existing \`config.maxFaucetIters\` override still wins, so existing tests pass unchanged.
- **`AwaitingFundingCard.tsx`**: the endpoint already returned \`balanceWei\` and \`targetWei\` — the SPA now uses them. The button stays clickable as **"Fund more"** when \`balance < target\` and only shows **"Faucet funded"** + disabled state when the target is actually reached. Status copy flips between "Faucet funding partial" and "Faucet funding complete" to match.

## Repro before this PR

1. Fresh \`JINN_EARNING_DIR\`, run \`jinn run\` on Base Sepolia.
2. Open the SPA, click "Fund from faucet".
3. Loop runs 60 drips, balance lands at ~0.006 ETH, button becomes disabled with label "Faucet funded".
4. \`/v1/bootstrap\` reports \`currentStep: awaiting_funding\` forever; daemon log shows \`[main] Awaiting funding... (Ns elapsed; will retry in 15s)\` indefinitely.

## After

1. Same fresh setup → click "Fund from faucet" → loop sizes itself to clear the gap (about ~120 drips on a fresh wallet) → balance reaches 0.010 ETH → bootstrap auto-advances to \`safe_deployed\` → SPA flips to "Joining Jinn".
2. If the loop is interrupted (rate limit, transient faucet error), the button reads "Fund more" with the current balance/target gap surfaced; clicking again re-issues the drip from where the balance is.

## Test plan

- [x] \`yarn typecheck\` (clean)
- [x] \`yarn build\` (TS + SPA)
- [x] \`yarn vitest run test/api/setup-endpoints.test.ts test/earning/bootstrap.test.ts\` — 27/27 pass
- [ ] Manual: fresh Base Sepolia operator on this branch reaches \`complete\` without operator intervention beyond the single faucet click.

## Notes

- The existing `setup-endpoints` test passes \`maxFaucetIters: 3\` explicitly via config; my dynamic-cap path early-returns when that override is set, so the test is unchanged.
- The dynamic cap uses \`100_000_000_000_000n\` (0.0001 ETH) as the conservative drip estimate; if CDP changes its drip size we can rebalance the multiplier without touching call sites.
- Out of scope: the hardcoded \`minEoaGasEth\` / \`STANDARD_MASTER_BOOTSTRAP_MULTIPLIER\` themselves — those are protocol-level values.

### DIFF
--- a/client/src/api/setup-endpoints.ts
+++ b/client/src/api/setup-endpoints.ts
@@ -157,7 +157,6 @@ export function addSetupRoutes(app: Hono, config: SetupRoutesConfig = {}): void 
     }
 
     const requestFunding = config.requestFunding ?? requestTestnetFunding;
-    const maxFaucetIters = config.maxFaucetIters ?? 60;
     const interDripPauseMs = config.interDripPauseMs ?? 1_000;
     const targetWei = config.minEoaGasWei ? BigInt(config.minEoaGasWei) : null;
     const publicClient = config.rpcUrl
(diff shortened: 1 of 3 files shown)


ritsuKai2000:
Closing as superseded by #88. The dynamic faucet cap and Fund more UI work are already covered there.

PR #85, opened 2026-05-04
fix(client/faucet): size drip cap to clear bootstrap target

oaksprout:
## Summary

Fresh-fleet testnet onboarding never auto-completes from the SPA's "Fund From Faucet" button because the faucet drip cap (60 × 0.0001 ETH = 0.006 ETH) is below the bootstrap target (0.005 ETH × `STANDARD_MASTER_BOOTSTRAP_MULTIPLIER` = 0.010 ETH).

The drip loop always exits "0.006 / 0.010 ETH" — the SPA renders that exact string ("Faucet funding complete (60 drips). Balance 0.006 ETH / target 0.010 ETH") and bootstrap stays pinned on `awaiting_funding`, retrying every 15 s with no progress.

## Fix

- Add a shared `computeFaucetDripCap(targetWei, balanceWei)` helper in `client/src/earning/faucet.ts` that sizes the iteration cap from `(target − balance) / ESTIMATED_DRIP_WEI` with a 60-drip floor (preserves the historical lower bound for callers without a target) and 500-drip ceiling (prevents runaways).
- Replace the two hardcoded 60s — `client/src/earning/bootstrap.ts` (auto-loop) and `client/src/api/setup-endpoints.ts` (`POST /v1/setup/drip`) — with calls to the helper.
- Add a 5-minute wall-clock deadline inside both loops as the real safety rail. The iteration cap is now just a bound; the actual exit conditions remain (a) target reached, (b) faucet rate-limited, (c) deadline elapsed.
- Update the bootstrap log message so the announced budget matches the dynamic cap instead of a stale "60 drips".

The protocol-level values (`minEoaGasEth`, `STANDARD_MASTER_BOOTSTRAP_MULTIPLIER`) are intentionally unchanged — they are correct; the cap was wrong.

## Tests

- `test/earning/bootstrap-faucet.test.ts`: existing "does not send extra request after cap" updated to assert against `computeFaucetDripCap` rather than the literal 60. Adds a new success test that simulates ~0.0001 ETH per drip and asserts the loop exits at 100 drips when the balance reaches 0.010 ETH.
- `test/api/setup-endpoints.test.ts`: adds a fresh-fleet drip-flow test that passes the 0.010 ETH target and asserts the loop runs enough drips to clear it (>60). Both would have failed under the old fixed cap.

## Test plan

- [ ] Verify on a fresh testnet earning dir: SPA "Fund From Faucet" reaches the 0.010 ETH target without manual top-up.
- [ ] Verify bootstrap auto-loop reaches the target with `autoTestnetFaucet=true`.
- [ ] Verify rate-limit early-exit still works (stub a 429 from CDP — should not exhaust the cap).
- [ ] Verify the 5-minute wall-clock fallback fires if drips silently no-op (stub `requestFunding` to return `ok: true` with no balance increase).

### DIFF
--- a/client/src/api/setup-endpoints.ts
+++ b/client/src/api/setup-endpoints.ts
@@ -26,7 +26,11 @@ import { dirname, join } from 'node:path';
 import { z } from 'zod';
 import { FleetStateStore } from '../earning/store.js';
 import { decryptMnemonic, encryptMnemonic } from '../earning/wallet.js';
-import { requestTestnetFunding } from '../earning/faucet.js';
+import {
+  DEFAULT_FAUCET_LOOP_TIMEOUT_MS,
+  computeFaucetDripCap,
+  requestTestnetFunding,
+} from '../earning/faucet.js';
 import { createJinnPublicClient, type JinnOnchainNetwork } from '../earning/viem-clients.js';
 import { detectAuthContext, probeClaudeAuth } from '../preflight/claude-auth.js';
 import { checkClaudeBinary, type ClaudeBinaryCheckResult } from '../preflight/claude-binary.js';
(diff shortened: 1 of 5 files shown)


ritsuKai2000:
Closing as superseded by #88 after folding the unique pieces into operator-shakedown: shared computeFaucetDripCap, the 5-minute wall-clock cutoff, and the stronger faucet/setup coverage.

PR #144, opened 2026-05-12
feat(hermes): make Hermes Agent the SWE-rebench v2 default solver harness (5/5)

ritsuKai2000:
**Stack 5 of 5** — Hermes harness integration (`jinn-mono-8psp`). Base: `hermes/4-runbook-e2e` (#143). **Merge #140 → #141 → #142 → #143 first.**

## Summary

Reverses the original draft posture (spec §10 / DR-2026-05-11-g — "Hermes ships opt-in; default-swap held until ≥3 operators report comparable-or-better frozen-mode scores") per the 2026-05-12 Captain decision (`jinn-mono-8psp.2`): commit to Hermes Agent as the SWE-rebench v2 default solver harness now, and prove it through the v1 acceptance run, rather than ship it opt-in and wait for field data a non-default harness won't accumulate.

- `client/src/main.ts` — reorder the `swe-rebench-v2` catalog's `compatibleHarnesses` so `HERMES_AGENT_HARNESS` is first; the dashboard's `JoinFlow` pre-selects `compatibleHarnesses[0]`, so Hermes is now the default-selected solver harness. Claude Code and Codex remain selectable.
- `client/src/dashboard/spa/src/pages/configuration/claudeModels.ts` — add `HERMES_MODELS` (`anthropic/claude-opus-4.6` [default, OpenRouter-routed], `anthropic/claude-sonnet-4.6`, `nousresearch/hermes-4-405b`); `modelOptionsForHarness('hermes-agent')` returns it so the model dropdown shows Hermes-routable `<provider>/<model>` ids instead of bare Claude-CLI ids; `resolveModelOption` recognizes them.
- `docs/runbooks/swe-rebench-v2-public-testnet.md` — "Solver joins default to Claude Code" → "default to **Hermes Agent**"; note every solver operator needs Hermes installed; note the dashboard pins `anthropic/claude-opus-4.6` (or leave the join's `model` unset to inherit `hermes model`).
- `docs/superpowers/specs/2026-05-11-hermes-harness-design.md` §10 + §13 — `DR-2026-05-12-a` supersedes `DR-2026-05-11-g`; the §10.2 data-driven criteria are retained as a **post-launch guardrail** (flip the default back if Hermes underperforms on the live leaderboard), not a ship gate. Fixed stale `jinn-mono-8psp.3` → `8psp.2` references.

## Trade-off

[...]

PR #140, opened 2026-05-12
refactor: rename claude-code-learner → learner + Hermes design docs (1/4)

ritsuKai2000:
**Stack 1 of 4** — Hermes harness integration (`jinn-mono-8psp`). Base: `main`.

## Summary

Preconditional naming refactor plus the design + plan documents for the Hermes harness integration epic.

- Spec: `docs/superpowers/specs/2026-05-11-hermes-harness-design.md`
- Plan: `docs/superpowers/plans/2026-05-11-hermes-harness-integration.md`
- Refactor: `client/src/harnesses/impls/claude-code-learner/` → `learner/`; `client/plugins/claude-code-learner/` → `client/plugins/learner/`; test dirs + e2e helpers renamed; `ClaudeCodeLearnerImpl` class → `LearnerHarness`; `ClaudeCodeLearnerConfig` → `LearnerHarnessConfig`; mechanical import-path sweep.

**On-chain identity preserved.** `Executor.implName` values stay `'claude-code'` / `'codex'` via the existing alias logic in `client/src/harnesses/names.ts` (`'claude-code-learner'` / `'codex-code-learner'` still map to canonical names; `harnessStateDirName()` still returns the legacy strings). No envelope migration, no operator-side state-dir migration. The plugin's marketplace `"name"` field in `.claude-plugin/` and `.codex-plugin/` JSON is also unchanged (it's an identity, like the on-chain string).

## Test plan
- [ ] `cd client && yarn typecheck` — zero errors
- [ ] `yarn test` — full suite green (no new failures vs main)
- [ ] `yarn build` — clean dist

## Why this is its own PR
The rename touches ~40 files; isolating it from the new `hermes-agent` package keeps both diffs reviewable. The `hermes-agent` work (stack 2) builds on the renamed `learner/` layout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #145, opened 2026-05-12
fix(hermes): real-binary integration fixes — wrong -w flag + auth-orphan (6/6)

ritsuKai2000:
**Stack 6 of 6** — Hermes harness integration (`jinn-mono-8psp`). Base: `hermes/5-default-swap` (#144). **Merge #140 → … → #144 first.**

Two bugs found running the adapter against a real Hermes binary (v0.13.0; Gemini 3 Flash via `google-gemini-cli` OAuth). Part of `jinn-mono-8psp.5`.

## Bug 1 — `hermes chat ... -w <dir>` is not a valid flag

The adapter built its spawn args from the docs (`hermes chat -q "..."` — `-q` IS valid) but appended `-w <workingDir>`, which `hermes chat` doesn't accept (it has `--worktree`, a boolean that makes Hermes create a git worktree — the opposite of what we want). The real binary rejected it with exit code 2 + a usage dump.

**Fix:** `hermes chat -q <prompt> -Q [--model X] [--provider Y]` — drops `-w`, adds `-Q` (quiet/programmatic mode). The working directory is set via the spawn `cwd` + `terminal.cwd` in the per-Task `config.yaml` (which `bootstrap` already writes).

## Bug 2 — per-Task `HERMES_HOME` redirection orphans the operator's OAuth creds

The freeze contract requires each Task to run under a fresh `$HERMES_HOME = ctx.implStateDir`. But Hermes resolves OAuth creds from `$HERMES_HOME/auth/google_oauth.json` with **no real-home fallback** (unlike the pooled `auth.json`, which falls back to `~/.hermes/auth.json`). So a Jinn-run Hermes task can't find the operator's credentials → fails with "No … credentials found." This would break Hermes-as-default in production for every operator.

**Fix:** `writePerTaskHermesConfig` now seeds `auth/` and `auth.json` from the operator's real Hermes home (`process.env.HERMES_HOME` or `~/.hermes`) into the per-Task home on first use. Verified: a fresh `HERMES_HOME` with the seeded `auth/` authenticates and runs (`hermes auth status` → logged in; a chat call reaches the model). `HermesHarnessAdapter` config gains an `operatorHermesHome` override (tests pass an empty dir → seed is a no-op). 4 new bootstrap tests.

**Caveat for frozen mode (follow-up):** a mid-task OAuth token refresh mutates `auth/google_oauth.json` → would trip the daemon hash-fence. The SWE-rebench v2 freeze-mode workstream should exclude `auth/` from the Hermes freeze hash, or pre-refresh the token before a frozen run. Tracked in `jinn-mono-8psp.5`.

## Also

- `client/test/e2e/hermes-agent-full-cycle.ts`: reads `JINN_HERMES_MODEL` / `JINN_HERMES_PROVIDER` env vars and forwards them to `buildHarnesses`; dropped the hardcoded `solverNet.model = 'anthropic/claude-opus-4.6'` from the ctx (would override the operator's model with a provider they may not have creds for). Committed e2e still defaults to the stub binary; for a real run: `JINN_HERMES_PATH=$(which hermes) JINN_HERMES_MODEL=gemini-3-flash-preview JINN_HERMES_PROVIDER=google-gemini-cli yarn e2e:hermes`.

## Open (tracked in `jinn-mono-8psp.5`)

- Per-Task `config.yaml` is fresh — doesn't carry the operator's `~/.hermes/config.yaml` non-auth settings (`agent:` budget, `compression:` / `auxiliary:` models, operator-installed skills). Model/provider come via `--model`/`--provider` flags (daemon `hermesModel`/`hermesProvider` config or per-SolverNet `model:`). For Hermes-as-default, operators must set the daemon `hermesModel`/`hermesProvider` — needs a runbook note. Spec §8.3's "inherits operator-level model selection" overstates it.
- Full real-binary run against a real daemon + a real (solveable) SWE-rebench v2 instance, MCP fully wired, evidence captured per the runbook.

## Test plan
- [ ] `yarn typecheck`
- [ ] `yarn vitest run test/harnesses/impls/hermes-agent` — 24/24 (incl. 4 new seed tests)
- [ ] `yarn e2e:hermes` (stub) — passes
- [ ] `yarn test` — full suite green


ritsukai (approved):
Reviewed Hermes stack unblock changes: runtime-only Hermes state is excluded from the freeze digest, ordinary state remains fenced, Hermes support is narrowed back to swe-rebench-v2, and the runbook default text is consistent.

ritsukai (approved):
Re-reviewed after narrowing the Hermes frozen-mode exception to runtime auth/config plus bin/tirith only; non-tirith binaries remain fenced.

PR #170, opened 2026-05-12
fix: /v1/solvernets/registry/:cid 404s for just-launched manifests when no subgraph is configured

ritsuKai2000:
## Summary

- Cache launched SolverNet manifests locally during the recording phase and persist manifestPath on the owned launched record.
- Teach /v1/solvernets/registry/:cid to serve an owned cached manifest before falling back to registry discovery.
- Improve the launched dashboard manifest-load error copy and cover the owned-record dashboard path.

## Validation

- cd client && yarn vitest run test/api/solvernets-endpoints.test.ts --testNamePattern "owned cached manifest"
- cd client && yarn vitest run test/solvernets/launch-state-machine.test.ts --testNamePattern "progresses through all 5 phases"
- cd client && yarn vitest run test/api/solvernets-endpoints.test.ts test/solvernets/launch-state-machine.test.ts
- cd client && yarn typecheck
- cd client && yarn vitest run --testTimeout=15000 (full suite passed: 396 files, 3176 tests, 4 skipped)

Note: exact cd client && yarn typecheck && yarn test was attempted twice and failed both times in pre-existing test/earning/bootstrap.test.ts cases that exceed Vitest default 5s timeout under full-suite load. The same bootstrap file passes in isolation, and the full suite passes with only --testTimeout=15000 raised.

### DIFF
--- a/client/src/api/solvernets-endpoints.ts
+++ b/client/src/api/solvernets-endpoints.ts
@@ -64,6 +64,7 @@ import {
   type SolverNetStore,
 } from '../solvernets/store.js';
 import {
+  manifestHash,
   signManifest,
   type UnsignedSolverNetManifestV1,
 } from '../solvernets/manifest.js';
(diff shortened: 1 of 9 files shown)


ritsukai (approved):

Issue #182, opened 2026-05-12
SWE-rebench v2 eval loop emits structurally-meaningless verdicts — model correctness almost never reaches grading


ritsuKai2000:
**Beads:** `jinn-mono-uy6v.8` · child of #157 (epic uy6v) · blocks #161 (uy6v.7)
**Run-mode:** FIX (regression-test-first)

## Context

Ran the SWE-rebench v2 loop locally against the bootstrapped Base Sepolia fleet (agent 5474) on 2026-05-12 and audited every evaluation run whose container log is on disk (`~/.jinn-client/engine/impl-state/swe-rebench-v2-evaluator/upstream/logs/`, n=80). Of the 52 verdicts cached locally, 51 are `passed_match: false`; the one pass (`astropy__astropy-18005`) is the only end-to-end success to date.

The failures are overwhelmingly **not** "the model produced a wrong fix" — the restorer's patch is only put on trial in ~6 of 80 runs. Where the rest died:

| # | Failure | Class |
|---|---|---|
| 46 | `docker: Cannot connect to the Docker daemon` → eval never executed. Upstream `eval.py` swallows it (no exception) and returns `passed_match:false`, which the harness signs and posts as a verdict. | env + harness |
| 12 | `git apply` → `error: corrupt patch at line N` on the restorer's patch (context lines missing their leading space; clusters around test-function boundaries). 4 such malformed diffs found among 83 cached solutions. | harness/agent |
| 5 | restorer patch edits a test file → gold `test_patch` 3-way-merges → conflict / unmerged file → `set -e` aborts. 76 of 83 cached solutions touch test files. | harness + agent behaviour |
| 3 | Docker storage I/O error (corrupt containerd metadata). | env |
| 2 | `patch does not apply` (also test-file collision). | harness |
| 2 | test cmd invokes bare `pytest`, not on PATH in the eval image. | dataset/env |
| 2 | `pip install -e .` install step fails to build. | env |
| 1 | test cmd assumes a `uv venv`. | dataset/env |
| 1 | amd64 eval image crashing under arm64/Rosetta emulation. | env |
| 6 | **tests actually ran** — 1 pass, 5 genuine wrong-answer. | model |

[...]

Issue #161, opened 2026-05-12
Live verdict-success + JINN reward distribution on Base Sepolia


ritsuKai2000:
Part of release epic jinn-mono-uy6v (v1 first public release). Child 7 — added 2026-05-11.

## Intent

The 2026-05-08 single-operator closure (DR-2026-05-08) settled with `score: 0` / `passed_match: false` — the loop ran, but no actual SWE-bench task was solved correctly and no JINN flowed. v1 must prove (a) the evaluator recognises success when it occurs, and (b) the reward distribution from SWE-rebench v2 SolverNet to operator wallets works on chain.

Loop completes ≠ verdict succeeds ≠ rewards distribute. v1 requires all three.

## Acceptance criteria (testable, problem-shaped)

- ≥ 1 Solution produces a passing verdict end-to-end on chain on Base Sepolia: `score > 0`, `passed_match: true`.
- ≥ 1 operator receives JINN reward distribution from the SWE-rebench v2 SolverNet — visible as a balance change in the operator app.
- Distribution / emission tx visible on chain.
- Operator app reflects the earned-JINN balance change without manual refresh / page reload (or with documented refresh expectations).

## Likely dependencies / overlaps

- jinn-mono-g9j4 — Exercise v0 governance flows end-to-end on testnet. Probably overlaps with reward-distribution mechanics; link or merge at planning time.

## Out of scope

- Mainnet emissions. v1 is Base Sepolia testnet only; mainnet is Phase 2.

## Reference
- DR-2026-05-08 (loop closed with score: 0): log/decisions/2026-05-08-swe-rebench-v2-loop-closed.md
- jinn-mono-g9j4 (v0 governance e2e)

---
Internal tracking: jinn-mono-uy6v.7

PR #183, opened 2026-05-12
fix(uy6v.8): SWE-rebench v2 eval loop — no verdicts on non-gradeable evals; source-only patches

ritsuKai2000:
Fixes the SWE-rebench v2 evaluation loop so it stops emitting structurally-meaningless verdicts. Closes the code half of `jinn-mono-uy6v.8` / #182; blocks #161 (live verdict-success + JINN distribution).

**Why** — a 2026-05-12 audit (n=80 eval container logs) found the model's patch is only put on trial in ~6 of 80 runs; the other ~74 "failures" never grade the model and every infra failure was being signed as an on-chain `passed_match:false` verdict. Buckets: 46 Docker-down, 12 corrupt-patch, 7 test-file-collision, ~6 env-mismatch, 6 actually-graded (1 pass).

### Commits (stack on one branch; review commit-by-commit)

1. **`fix(uy6v.8): evaluator emits no verdict when the eval couldn't grade the solution`** — new `EvalCouldNotGradeError` in `eval-runner.ts` (classifies "container exit≠0 AND no expected test collected" as an infra abort — Docker-down, patch-apply failure, test-file conflict, install failure, arch-incompatible image, missing report); `harness.run()` re-raises it as `SkippableError` → engine records a skip, no verdict delivered, no test log pinned. A genuine wrong-answer run (FAIL_TO_PASS ran and failed) is still graded. Also fixes `passed`/`failed` to read the real upstream `passed_actual`/`failed_actual`. **(AC1)**
2. **`fix(uy6v.8): live Docker readiness gate on the SWE-rebench v2 evaluator`** — `isReady()` now probes `docker info` after the marker/upstream-repo checks; on failure → not-ready, so the engine's claim-eligibility check skips evaluation tasks until Docker is back (Docker was only validated once, at `jinn harnesses enable` time). **(AC2)**
3. **`fix(uy6v.8): SWE-rebench v2 restoration submits a source-only patch from git diff`** — new `restoration-patch.ts` (`stripTestPathHunks`, `isTestPath`, `gitApplyParseCheck`); `harvest.ts` now prefers the `git diff` over the task checkout over any agent-authored `.execute/solution-payload.json` for swe-rebench restoration, and strips test-path hunks from whichever patch wins (an all-test diff yields no submission → run re-attempted). Standard SWE-bench resets test files before grading; this approximates that on the producer side, since the restorer never sees the gold `test_patch`. Prompt updated: edit source only, test changes are discarded, let the harness derive the diff. **(AC3)**
4. **`fix(uy6v.8): pin SWE-rebench eval to linux/amd64 + capture the real container log`** — `DOCKER_DEFAULT_PLATFORM=linux/amd64` on the eval subprocess; resolve a relative report `log_path` against `upstreamRepoDir` so the pinned test-log artifact carries the actual pytest output, not just the progress bar.

### Tests

Regression-test-first throughout. New/changed tests in `swe-rebench-v2-evaluator/{eval-runner,harness}.test.ts` and `claude-code-learner/{restoration-patch,harvest,codex-code-adapter}.test.ts`. Full client suite green (`yarn test`: 3194 passed / 4 skipped); `yarn typecheck` clean.

### Not in this PR (follow-ups / manual)

- **AC4 re-baseline** (manual, post-merge): re-run the existing backlog on a linux/amd64 operator; the infra verdict buckets should → 0; report the actual pass rate over ≥50 graded attempts.
- Local loop smoke (Docker stopped → no `passed_match:false` verdicts produced; Docker up → eval reaches the grading stage) — run before merge.
- Stricter producer-side patch validation wired into harvest (`gitApplyParseCheck` is implemented + tested but not wired — would need fixture churn across the suite); auto-retry of `SkippableError`-skipped evals; an upstream `eval.py` PR (reset test paths before the gold patch / exit non-zero on infra failure) — all noted in the bead.

### DIFF
--- a/client/src/cli/commands/solver-nets.ts
+++ b/client/src/cli/commands/solver-nets.ts
@@ -1,4 +1,5 @@
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from 'node:fs';
+import { spawnSync } from 'node:child_process';
 import { homedir } from 'node:os';
 import { dirname, join } from 'node:path';
 import { parseArgs } from 'node:util';
@@ -218,6 +219,11 @@ const command: CommandModule = {
   jinn solver-nets remove-plugin <name> <source-or-name> [--config <path>]
   jinn solver-nets doctor <name> [--human|--json] [--config <path>]
   jinn solver-nets sample <name> [--closed-window]
+  jinn solver-nets validate-pool swe-rebench-v2 [--limit <n>] [--force]
+            Run the gold patch of each pool instance through the eval harness
+            and cache which instances are scorable; the generator then posts
+            only those. Requires Docker + \`jinn harnesses enable
+            swe-rebench-v2-evaluator\`.
 
 Output flags:
   --human   Render readable terminal output instead of JSON (supported by
@@ -236,6 +242,8 @@ Output flags:
         config: { type: 'string' },
         harness: { type: 'string' },
         'closed-window': { type: 'boolean' },
+        limit: { type: 'string' },
+        force: { type: 'boolean' },
         human: { type: 'boolean' },
         json: { type: 'boolean' },
       },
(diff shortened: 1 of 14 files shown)


ritsuKai2000:
**Update — commit 5 (`77d8adf4`): align with the real `eval.py` report shape + resolved-semantics verdict**

Verifying the loop against real Docker (golden eval on `astropy-18005` → `passed_match: true`, plus `infrahub-8423` / `cyclopts-701`) surfaced two issues with commit 1, now fixed:

1. **Wrong report-item fields.** I'd read `passed_actual` / `failed_actual` — but those are in `eval.py`'s *internal* result dict, not the report. The report items carry `from_fail_to_pass`, `failed_from_pass_to_pass`, `passed_match`, `exit_code`, `log_path`, `error` (plus a setup-error shape with just `error` + the P2P list). With the wrong names, the "did the eval grade?" heuristic degenerated to "did pytest exit 0?", which would have **suppressed legitimate `passed_match:false` verdicts** (a model patch that doesn't fix the F2P test → pytest exits non-zero → would've been skipped instead of failed). The unit tests passed only because the fixture used my wrong field names — tested-the-mock-not-reality. Fixture now mirrors the real shape; the infra-signature scan is now the load-bearing gate (non-zero exit + zero tests passed → ungradeable *only if* the output matches a known infra signature; otherwise it's a real, possibly-degenerate failure verdict).

2. **`passed_match` is structurally unscorable for a chunk of instances** (raised in review). Upstream computes it as `{every test that passed} == {FAIL_TO_PASS ∪ PASS_TO_PASS}` — so any instance whose dataset `test_cmd` runs the whole suite (e.g. `cyclopts-701`: `pytest -v`, 1817 tests vs 29 named) can never score, and a solver is penalised for adding a passing test. We now re-derive the verdict with **SWE-bench "resolved" semantics** — `all FAIL_TO_PASS now pass AND no PASS_TO_PASS broke` — from `from_fail_to_pass` / `failed_from_pass_to_pass`, ignoring unlisted tests.

Live verification (real Docker):
| instance | patch | verdict |
|---|---|---|
| `astropy-18005` | gold | ✅ `passed_match: true` (34/34) |
| `cyclopts-701` | gold | ✅ `passed_match: true` — *was structurally unscorable before* (upstream `passed_match: false`) |
| `cyclopts-701` | no-op | `passed_match: false` — a real verdict, not suppressed |
| `infrahub-8423` | gold | `EvalCouldNotGradeError` — 37 P2P tests ERROR (no DB in the eval container) → genuinely can't grade → **no verdict signed** (was a bogus `passed_match:false` before this PR) |

Full client suite green (3196 pass); typecheck clean.

Still open follow-ups (separate beads): the upstream `eval.py` exact-set `passed_match` (and `test_cmd` scoping) is a dataset/upstream issue worth an upstream PR; auto-retry of `SkippableError`-skipped evals; the `EvalCouldNotGradeError` reason string for "parser returned nothing" could be more precise than `eval_aborted_before_tests` (renamed/handled in commit 5's classifier).

ritsukai (approved):

PR #189, opened 2026-05-13
feat(uy6v.11): bound swe-rebench-v2 eval-image cache with in-process LRU

ritsuKai2000:
## Why

The SWE-rebench v2 evaluator pulls a per-instance Docker image (`swerebench/sweb.eval.x86_64.<repo>_1776_<instance>:latest`, 1–8 GB each, ~3 GB avg). Upstream `eval.py` uses `docker run --rm` so the *container* is cleaned, but the *image* stays cached locally. The leaderboard pool has 741 unique instances; if all were pulled, the cache would balloon to ~2–5 TB.

Observed live (2026-05-13): ~16 verdicts/hour. Day-1 unique-instance pulls trend 50–100 → 150–300 GB on day 1, settling lower as repeats accumulate. A 1 TB operator disk fills in ~2–3 weeks of continuous operation; smaller disks faster. Operators currently have to manually `docker image prune` or watch disk usage.

## What

`PythonEvalRunner` now tracks an in-process LRU of image tags. After each `runEval` — success or thrown `EvalCouldNotGradeError`, via a `finally` block — the just-used image is bumped to the most-recent slot; when the set exceeds the configured cap, the least-recently-used image is removed via the injected `cleanupImage` hook (default: `docker rmi <image>` via `spawn`, errors swallowed best-effort).

Key properties:

- **Repeat-evals of recently-used instances never re-pull.** Refresh-on-use moves the image to the most-recent slot before the cap is checked, so the just-evaluated image is never the eviction target. Only the genuinely-oldest entries are rmi'd.
- **Cleanup runs on every terminal exit path.** The `recordImageUsage` call lives in `finally`, so a pull-and-crash failure (Docker storage IO error, image_arch_mismatch, patch_corrupt, eval_no_report, upstream setup-error) still counts toward the cap — the failure path can't leak the LRU.
- **Best-effort GC.** `cleanupImage` errors are swallowed: a failed `docker rmi` leaves the image on disk for the next eviction attempt but never escapes `runEval`.
- **Process-lifetime LRU.** No persistence — daemon restarts re-discover the cache state by usage; that's acceptable (worst case is re-pulling a recently-used image once after restart).

## Operator knob

`JINN_EVAL_IMAGE_CACHE_MAX` (env-only, mirrored in the canonical config table in `CLAUDE.md`).

- **Default: 20.** At ~3 GB/image average, that's a ~60 GB steady-state working set — small enough that even a 256 GB SSD has headroom, large enough that frequently-repeating subsets of the pool rarely re-pull.
- **Programmatic overrides.** `PythonEvalRunnerOptions.imageCacheMax` takes precedence over the env var (used by the test suite; available to callers that build `PythonEvalRunner` directly, e.g. `validate-pool`).

I picked **20** over 10 because at ~3 GB/instance the absolute disk cost is bounded (60 GB) and the leaderboard's verdict mix re-uses instances often enough that a too-aggressive cap would visibly increase repeat-pull traffic. Operators on small disks can lower it via the env var; nothing in the loop depends on a particular value.

## Tests (TDD)

Six new tests in `client/test/harnesses/impls/swe-rebench-v2-evaluator/eval-runner.test.ts`, all under a `LRU image cache GC (jinn-mono-uy6v.11)` describe block:

1. No eviction while distinct usages ≤ cap (keeps repeat-evals fast).
2. Oldest evicted once usage exceeds the cap (FIFO eviction order).
3. Refresh-on-repeat protects recently-used images from eviction.
4. Cleanup runs on the upstream setup-error failure path.
5. Cleanup runs on the no-report (eval.py crashed) failure path.
6. A throwing `cleanupImage` never escapes `runEval`.
7. The env-var default is honoured by the production `docker rmi` path (verified by stubbing `docker` on PATH and inspecting the recorded args).

Existing 11 `PythonEvalRunner` tests still pass. Full vitest run: 3218 passing locally (2 pre-existing unrelated flakes: an anvil-fork timing test and a hyperliquid testnet round-trip).

## Out of scope

Wider Docker cache management (base-image layers, build caches) and pre-fetching / cache-warming. This PR is the per-instance eval images specifically.

Refs jinn-mono-uy6v.11.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -269,6 +269,7 @@ Config file first, env var override. File at `~/.jinn-client/config.json` or `--
 | ipfsGatewayUrl   | JINN_IPFS_GATEWAY_URL    | https://gateway.autonolas.tech    |
 | engine.workingDirRoot | JINN_ENGINE_WORKING_DIR_ROOT | ~/.jinn-client/engine/work   |
 | engine.implStateDirRoot | JINN_ENGINE_IMPL_STATE_DIR_ROOT | ~/.jinn-client/engine/impl-state |
+| _(none — env-only)_  | JINN_EVAL_IMAGE_CACHE_MAX | 20 (cap on the swe-rebench-v2 per-instance Docker image LRU) |
 
 `JINN_PASSWORD` is env-only — never in config files.
 
--- a/client/src/harnesses/impls/swe-rebench-v2-evaluator/eval-runner.ts
+++ b/client/src/harnesses/impls/swe-rebench-v2-evaluator/eval-runner.ts
@@ -98,25 +98,50 @@ export function resolveImageCacheMax(opt: number | undefined): number {
   if (typeof opt === 'number' && Number.isFinite(opt) && opt > 0) return Math.floor(opt);
   const envRaw = process.env['JINN_EVAL_IMAGE_CACHE_MAX'];
   if (envRaw !== undefined) {
-    // `Number()` rejects strings with trailing garbage (e.g. `"1e3oops"` →
-    // NaN), unlike `parseInt`. We want operators who typo this to fall back
-    // to the default, not silently get `1`.
+    // `Number()` returns 0 for `""` / whitespace and NaN for strings with
+    // non-numeric content (e.g. `"garbage"`, `"1e3oops"`) — unlike `parseInt`,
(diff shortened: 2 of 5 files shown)


ritsukai (approved):

PR #190, opened 2026-05-13
fix(uy6v.10): emit verdict in swe-rebench-v2 gating so reputation hook fires

ritsuKai2000:
## Bug

The new swe-rebench-v2-evaluator harness was returning `gating: { score, passed_match }` with no `verdict` field. The engine's reputation-feedback hook keys on `gatingClaim.verdict`, so for every verdict-bearing delivery the hook emitted a silent skip log and never submitted an ERC-8004 `ReputationRegistry.giveFeedback` tx:

```
[harness-engine] <rid>: reputation feedback skipped — gatingClaim has no recognised verdict (got=undefined)
```

Live-verified on Base Sepolia 2026-05-13: **9/9 of the new daemon's swe-rebench-v2 verdicts hit this path**. Five of the nine were `passed_match: true` — all of those PASS verdicts should have been minting reputation; none did.

**Impact**: `uy6v.7`'s acceptance criterion — "live verdicts AND JINN reward distribution on Base Sepolia" — had its first half satisfied (verdicts flowing) but the second half blocked. The swe-rebench v2 SolverNet's distribution contract consults the reputation registry; with no reputation accruing, no JINN flows; release-epic gate stays unmet. The bug is *silent* — the verdict still posts on chain, only the reputation update is missing — easy to overlook in review.

**Secondary**: `verdictCodeForTask` in the engine also reads `gating.verdict` and silently defaults to PASS (1) when undefined — so the on-chain verdict tag in `claimDelivery` was being misrecorded as PASS for *every* swe-rebench-v2 delivery regardless of test outcome. Same fix resolves both.

## Root cause

`SweRebenchV2EvaluatorHarness.run()` in `client/src/harnesses/impls/swe-rebench-v2-evaluator/harness.ts` emitted

```ts
return {
  ...
  gating: { score, passed_match },   // no `verdict` field
  ...
};
```

Every other evaluator in the repo (`prediction-v0-evaluator`, `prediction-v1-evaluator`, `portfolio-v0-evaluator`, `prediction-apy-v0-evaluator`) emits `verdict: 'PASS' | 'FAIL' | 'INDETERMINATE' | 'REJECTED'` in `gating`. swe-rebench-v2 was the outlier and silently broke the engine's reputation-feedback contract.

## Fix

Option B (harness aligns to engine contract — all other evaluators already do this). Map `passed_match` → `'PASS' | 'FAIL'` and include in `gating`:

```ts
const verdict: 'PASS' | 'FAIL' = verdictPayload.passed_match ? 'PASS' : 'FAIL';
return {
  ...
  gating: { score, passed_match, verdict },
  ...
};
```

Chosen over Option A (teach engine to recognise swe-rebench-v2's bespoke shape) because Option B has the smaller blast radius and matches what 4 other evaluators already do.

## Regression coverage

- `client/test/harnesses/engine/engine-reputation-feedback.test.ts` — three new cases pinning swe-rebench-v2 gating shape through the engine:
  - PASS gating → `giveFeedback(score=100, tag1='swe-rebench-v2.v1')`
  - FAIL gating → `giveFeedback(score=0, ...)`
  - **Pre-fix shape (no `verdict`) → still emits the recognisable skip log** so the regression can't recur silently. If a future harness ships with the pre-fix shape, this test catches it loudly.
- `client/test/harnesses/impls/swe-rebench-v2-evaluator/harness.test.ts` — passing/failing-grade tests now assert `verdict: 'PASS'|'FAIL'` is present in `sol.gating`.

## Verification

- `tsc --noEmit` — zero errors
- `vitest run` — 3217 passing / 4 skipped (one unrelated `anvil.test.ts` port-bind flake that retries clean)

## Acceptance criteria check

- [x] Eval delivery from swe-rebench v2 evaluator no longer prints `reputation feedback skipped — gatingClaim has no recognised verdict` (regression test `engine-reputation-feedback.test.ts`).
- [x] Registry-update code path is reached for both `passed_match: true` (score=100) and `passed_match: false` (score=0).
- [x] Existing reputation-feedback tests cover the swe-rebench v2 gating shape (extended `engine-reputation-feedback.test.ts` + `harness.test.ts`).
- [ ] Live on-chain feedback tx — must be observed on Base Sepolia after this PR ships canary; not testable from this PR.

Refs `jinn-mono-uy6v.10`, GH parent #182.

### DIFF
--- a/client/src/harnesses/impls/swe-rebench-v2-evaluator/harness.ts
+++ b/client/src/harnesses/impls/swe-rebench-v2-evaluator/harness.ts
@@ -363,11 +363,19 @@ export class SweRebenchV2EvaluatorHarness implements Harness {
       'utf8',
     );
 
+    // Derive the engine-facing `verdict` from `passed_match`. The engine's
+    // reputation-feedback hook (and `verdictCodeForTask` for the on-chain
+    // verdict tag in `claimDelivery`) keys on `gating.verdict`. Before this
+    // mapping, the hook silently no-op'd on every swe-rebench-v2 delivery and
+    // every verdict tag defaulted to PASS — see jinn-mono-uy6v.10.
+    const verdict: 'PASS' | 'FAIL' = verdictPayload.passed_match ? 'PASS' : 'FAIL';
+
     return {
       venueRef: { name: 'swe-rebench-v2' },
       gating: {
         score: verdictPayload.score,
         passed_match: verdictPayload.passed_match,
+        verdict,
       },
       informational: {
         instance_id: task.instance_id,
--- a/client/src/harnesses/impls/swe-rebench-v2-evaluator/harness.ts
+++ b/client/src/harnesses/impls/swe-rebench-v2-evaluator/harness.ts
@@ -48,6 +48,12 @@ const UPSTREAM_REPO_URL = 'https://github.com/SWE-rebench/SWE-rebench-V2.git';
 const STATE_FILE = 'state.json';
 const ENABLE_CLI = 'jinn harnesses enable swe-rebench-v2-evaluator';
 
+/** The two verdict values emitted by this evaluator. The broader registry
+ *  shape (`reputation.ts`) also recognises `'INDETERMINATE'` / `'REJECTED'`,
+ *  but swe-rebench v2 grades are binary: the gold tests either resolve or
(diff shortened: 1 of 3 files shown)


ritsukai (approved):

PR #210, opened 2026-05-13
feat(et6s): jinn create plugin scaffold (two patterns)

ritsuKai2000:
Implements [#201](https://github.com/Jinn-Network/mono/issues/201) (\`jinn-mono-et6s\`) — first child of the [Plug-in builder entry point epic](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`).

Stacked on top of #209 (the epic's spec + plan).

## Summary

Extends \`jinn create\` with a \`plugin\` target. Two patterns:

- \`solver-type-plugin\` (default) — for plug-ins targeting a specific SolverType (modeled on \`client/plugins/swe-rebench-v2-runtime/\`).
- \`runtime-plugin\` — for \`jinn.supports: ['jinn.runtime']\` plug-ins exposing MCP tools to every harness session (modeled on \`client/plugins/network-tools/\`).

Builders go from zero to a working scaffolded package whose \`yarn install && yarn test\` passes on first run.

## What ships

- \`client/templates/plugins/{solver-type-plugin,runtime-plugin}/*\` — 15 template files total.
- \`client/src/cli/commands/create.ts\` — extended to dispatch on \`target: 'harness' | 'plugin'\` (\`resolveTemplatesRoot\` refactored to take target arg + per-call lookup; \`runCreate\` body dispatches; \`run\` (CLI handler) dispatches on subcommand with per-subcommand pattern + solver-type defaults + validation).
- \`client/test/cli/commands/create.test.ts\` — three new \`describe\` blocks: solver-type-plugin scaffold (Task 3), runtime-plugin scaffold (Task 5), and a first-run-\`yarn install && yarn test\` E2E pair (Task 6).

## Test plan

- [x] \`yarn typecheck\` clean
- [x] \`yarn vitest run test/cli/commands/create.test.ts\` — 12 tests pass (3 existing harness + 4 new plugin)
- [x] \`yarn test\` full suite — 3217 passed, 0 failed, 4 skipped (no regressions)
- [x] Manual: \`node dist/bin/jinn.js create plugin sanity-check-plugin --pattern solver-type-plugin --solver-type swe-rebench-v2.v1 --out-dir /tmp\` produces a valid package; \`grep -c '{{' /tmp/sanity-check-plugin/**\` returns 0 (all placeholders substituted).

## CLI

\`\`\`
jinn create plugin <packageName>
  --pattern=<solver-type-plugin|runtime-plugin>   (default: solver-type-plugin)
[...]

PR #209, opened 2026-05-13
docs(52x3): plug-in builder entry point — spec + plan

ritsuKai2000:
Spec + execution plan for [#199](https://github.com/Jinn-Network/mono/issues/199) (`jinn-mono-52x3`) — Plug-in builder entry point epic.

## What this PR ships

Docs only:
- \`cargo/docs/superpowers/specs/2026-05-13-plug-in-builder-entry-point-design.md\` (v0.4) — substrate design (identity, registry, attribution, reputation) + seven children's acceptance criteria.
- \`cargo/docs/superpowers/plans/2026-05-13-plug-in-builder-entry-point.md\` — execution roadmap: filed 7 child beads (#200-#206), dependency graph, planning prompts for each, and Task 8 \`/simplify\` closeout.

## What's already executed

The plan's Tasks 0-7 (file the seven child beads) have been completed; each child is open in bd with full descriptions and dependency edges. Children's GitHub issues: #200-#206. All labeled \`epic:52x3\` + \`sprint:sprint-2\`. Added to the [Jinn engineering](https://github.com/orgs/Jinn-Network/projects/1) project board (Sprint = Sprint 2, Epic = Plug-in builder).

## Sequencing reminder (spec §6 + plan §Sequencing)

This is a **post-v1 epic**: children only start shipping once \`uy6v\` ships and \`8psp\` merges. The first child to ship (\`jinn-mono-et6s\` — \`jinn create plugin\` scaffold, GH #201) is fully independent and ships stacked on this docs PR (see next PR in the stack).

## Test plan

- [x] Spec self-review passed: no placeholders; internally consistent; tight scope.
- [x] Plan self-review passed: bd commands inline; per-bead planning prompts concrete; PR sequencing documented.
- [x] All 7 child beads filed in bd with dependency edges (parent-child + blocks).
- [x] All 7 child GH issues created (#200-#206) and added to the engineering Project board.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #200, opened 2026-05-13
Staged-bootstrap refactor + identity_registered Stage 1 capstone step


ritsukai:
**bd:** `jinn-mono-nghf` · **Epic:** #199 · **Spec:** `docs/superpowers/specs/2026-05-13-plug-in-builder-entry-point-design.md` · **Plan:** `docs/superpowers/plans/2026-05-13-plug-in-builder-entry-point.md`

---

Part of epic jinn-mono-52x3. Refactor the existing 11-step bootstrap state machine into two independently re-entrant stages, per spec §5.1 + §6.1.

## Why

The existing bootstrap (`client/src/earning/bootstrap.ts`) is single-stage: `jinn run` walks 11 steps from wallet → mech_deployed. A builder publishing a plug-in needs only the first half (the ERC-8004 identity); forcing them through the OLAS / mech / staking half is wrong (no OLAS funding required, no daemon needed). The clean factoring is one state machine, two completion levels, lazily ensured per action point.

## Acceptance criteria (testable, problem-shaped)

- `client/src/earning/bootstrap.ts` exposes two entry points: `ensureStage1(ctx)` and `ensureStage1And2(ctx)`. Each walks the state machine forward to its target stage; both idempotent and re-entrant.
- The state file (`~/.jinn-client/earning/earning_state.json`) gains a `stage` marker and a per-step `stage` tag.
- A new `identity_registered` step is added as Stage 1's capstone: calls `IdentityRegistry.register()` then `setAgentWallet(agentId, safeAddress)` via the existing Safe-routed tx path. Composes with or absorbs `jinn-mono-j07` (agent NFT mint) and `jinn-mono-aev` (`setAgentWallet`) — planning decides which.
- `awaiting_funding` gate splits: Stage 1 requires ETH for gas only; Stage 2 additionally requires OLAS.
- `jinn run` continues to ensure Stage 1+2 (existing behaviour preserved); a unit test asserts re-running on a partially-completed Stage 1 state file picks up at the right step.
- A builder workflow test: starting from clean state, calling `ensureStage1` completes Stage 1 only; subsequent `ensureStage1And2` picks up at `service_created`. No re-mint, no second Safe, no second agentId.
- `yarn typecheck` + `yarn test` green.

## Out of scope (other beads or already covered)

- `jinn solver-plugins publish` calling `ensureStage1` — that's 52x3.3.
- Convenience verb `jinn identity ensure` — spec §9 open question.
[...]

Issue #199, opened 2026-05-13
EPIC: Plug-in builder entry point — first builder-published plug-in surfaced and scored on SWE-rebench v2 (Hermes-shape)


ritsukai:
**bd:** `jinn-mono-52x3` · **Spec:** `docs/superpowers/specs/2026-05-13-plug-in-builder-entry-point-design.md` (v0.4) · **Plan:** `docs/superpowers/plans/2026-05-13-plug-in-builder-entry-point.md` · **Branch:** `feat/52x3-plug-in-builder-entry-point`

The builder-facing complement to [`jinn-mono-uy6v`](https://github.com/Jinn-Network/mono/issues/157) (operator install path) and [`jinn-mono-8psp`](https://github.com/Jinn-Network/mono/issues/158) (Hermes harness integration). Ships the cold-start-to-first-score path for an external builder publishing a plug-in for the SWE-rebench v2 SolverNet running against the Hermes harness.

## Why this epic exists

The plug-in substrate is built — `jinn.plugin.json` schema and runtime, six source kinds, `jinn solver-plugins {show,validate,pack}` and `jinn solver-nets add-plugin` CLIs, Hermes consuming the standard SolverPlugin shape via `hermesConfigFromSolverPlugins()`. Plug-in attribution is already in the signed envelope (`executor.plugins[]` in `jinn.execution.v1`). Identity + reputation primitives ship via ERC-8004 (`client/src/erc8004/{identity,reputation}.ts`). Cross-operator score aggregation already ships under [`jinn-mono-ebu7`](https://github.com/Jinn-Network/mono/issues/166) (76% complete). Reputation + slashing already scoped under `jinn-mono-0www` (Phase B.2).

The remaining gap is **discovery + builder identity wiring + builder-shape view**: builders publish to npm but operators can't find them, the ERC-8004 identity primitive is operator-only by convention even though the contract layer is role-agnostic, and the existing aggregation has no builder-shape view.

## Substrate design (spec §5)

**One stateful identity, two completion levels.** Same EOA / Safe / agentId / keystore for builder and operator roles. Bootstrap is a single state machine with two stages, each independently re-entrant:

- **Stage 1 (Identity)** — required for any participation; mints the ERC-8004 agentId. Steps: wallet → safe_predicted → awaiting_funding (ETH only) → safe_deployed → identity_registered.
- **Stage 2 (Operator)** — opt-in; wires Stage 1 identity into OLAS service / mech / staking. Required only for users running a daemon.

[...]

PR #212, opened 2026-05-14
refactor(nghf): staged bootstrap — fleet-level Stage 1 + ensureStage1/ensureStage1And2 entry points

ritsuKai2000:
Implements [#200](https://github.com/Jinn-Network/mono/issues/200) (\`jinn-mono-nghf\`) — the foundational refactor of the [Plug-in builder entry point epic](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`).

Stacked on [#210](https://github.com/Jinn-Network/mono/pull/210) (\`et6s\` scaffold CLI) → [#209](https://github.com/Jinn-Network/mono/pull/209) (epic spec+plan).

## Summary

Refactors \`FleetBootstrapper\` (\`client/src/earning/bootstrap.ts\`) so its state machine has two independently re-entrant entry points:

- **\`ensureStage1(password)\`** — identity only. Walks: derive agent EOA (HD index 1) → predict Safe deterministically → ETH funding gate → deploy Safe → \`IdentityRegistry.register()\` → \`setAgentWallet()\` → set \`fleet_stage: 'stage1'\`.
- **\`ensureStage1And2(password)\`** — identity + per-service operator state. Calls \`ensureStage1\` first; if \`fleet_stage === 'stage1'\` and \`services: []\`, creates the first service row at \`awaiting_stake\` (standard) or \`service_created\` (self-bond) and walks Stage 2.

The original public \`bootstrap(password)\` is preserved as a thin alias to \`ensureStage1And2\` so existing CLI callsites (\`bootstrap.ts\`, \`fleet-scale.ts\`) don't break.

## Architecture

Per spec §5.1 (v0.5) + spike findings (\`docs/superpowers/specs/2026-05-14-nghf-staged-bootstrap-fit-findings.md\`):

\`FleetStateSchema\` gains four fleet-level identity fields:
- \`fleet_agent_id: string | null\`
- \`fleet_safe_address: string | null\`
- \`fleet_identity_registry: string | null\`
- \`fleet_stage: 'none' | 'stage1' | 'stage1_and_2'\`

Service rows in \`services[]\` remain strictly Stage 2 state. The Stage 1 Safe uses the self-bond Safe topology (deterministic prediction from agent key) unconditionally — so it works against both standard (stOLAS) and self-bond staking modes.

### Two-Safe topology in standard mode

A dual-role operator in standard mode ends up with two Safes:
- **Identity Safe** (Stage 1, deterministic from HD-index-1 agent key) — agentId bound here; \`setMetadata\` for \`plugin:<cid>\` flows through it; reputation feedback accrues here.
- **Staking Safe** (Stage 2, from \`distributor.stake()\`) — OLAS service runs here; rewards land here.

The agent EOA signs for both. Mechanically clean; users see one agentId. Tested explicitly (\`bootstrap.test.ts: ensureStage1And2 creates fresh service with two-Safe topology\`).

## Migration

Non-destructive promotion in \`store.ts:parseFleetStateJson\`. When loading a legacy state file with \`services[0].agent_id\` set but no \`fleet_agent_id\`:
- \`services[0].agent_id\` → \`fleet_agent_id\` (copy; original retained)
- \`services[0].safe_address\` → \`fleet_safe_address\`
- IdentityRegistry address inferred from chain config → \`fleet_identity_registry\`
- \`fleet_stage: 'stage1_and_2'\` (the operator was already running)

Existing testnet operators upgrading their daemon don't deploy a new Safe — their per-service Safe IS the identity Safe via migration. Only NEW builder-only users (or operators publishing plug-ins via 1pbc later in the epic) trigger \`ensureStage1\` and get a fresh Stage 1 Safe.

## What ships

- \`client/src/earning/types.ts\` — 4 fleet-level fields + \`FleetStageSchema\`.
- \`client/src/earning/store.ts\` — non-destructive migration.
- \`client/src/earning/bootstrap.ts\` — \`ensureStage1\` (3 step helpers), \`ensureStage1And2\` wrapper, \`bootstrap()\` alias.
- \`client/src/main.ts\` — daemon callsite uses \`ensureStage1And2\` explicitly.
- \`client/test/earning/types.test.ts\` — schema + migration tests.
- \`client/test/earning/bootstrap.test.ts\` — Stage 1 + Stage 1+2 walks; two-Safe topology; ETH-only funding gate; idempotency + resume.

## Test plan

- [x] \`yarn typecheck\` clean
- [x] \`yarn test\` — 3236 passed, 0 failed, 4 pre-existing skipped (3217 pre-nghf; +19 from this PR's new tests)
- [x] \`yarn build\` clean
- [x] Two-Safe topology assertion passes
- [x] ETH-only Stage 1 funding gate enforced
- [x] Migration test: legacy state file → fleet-level identity promoted non-destructively
- [x] \`bootstrap()\` alias keeps existing CLI callsites green (20 tests in bootstrap.test.ts pass)
- [x] Idempotency: re-running \`ensureStage1\` or \`ensureStage1And2\` on completed state is a no-op

## Risks flagged in concerns

- Pre-existing flaky test in \`test/trajectory/transcript-watcher.test.ts\` (timing-related, intermittent 2000ms timeout) — unrelated to this PR.
- Three bootstrap-test files needed mock surface updates to accommodate the Stage 1 prefix in fresh-fleet flows. Intent of each test preserved.

## Reference

- Parent epic: [#199](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`)
- Spec (amended in this PR): \`cargo/docs/superpowers/specs/2026-05-13-plug-in-builder-entry-point-design.md\` §5.1 (v0.5)
- Spike findings: \`cargo/docs/superpowers/specs/2026-05-14-nghf-staged-bootstrap-fit-findings.md\`
- Plan: \`cargo/docs/superpowers/plans/2026-05-14-nghf-staged-bootstrap-plan.md\`


ritsuKai2000 (commented):
## Review (agent: substrate)

**Verdict:** approve-with-comments

### Strengths

- The migration in `applyNghfMigration` is genuinely non-destructive: it runs on read (not write), is idempotent, and the `fleet_agent_id` null-guard at the top of the function prevents double-promotion. The five migration test cases cover the material edge conditions (pre-j07 no-agent-id, mid-walk, coexisting fleet+service identity).
- Crash-safety in `stepFleetIdentityRegister` is handled correctly: `fleet_agent_id` is persisted immediately after the mint tx receipt, before the `setAgentWallet` call, so a crash between those two steps leaves the agentId recoverable on restart rather than silently re-minting.
- The `bootstrap()` alias preserving back-compat for `cli/commands/bootstrap.ts` and `fleet-scale.ts` is clean; existing callers need zero changes and the alias carries no logic of its own.

### Findings

#### [Priority: high] `stepFleetSafePredict` discards the incoming `state` with `void state`

`stepFleetSafePredict` receives `state` as a parameter, immediately discards it (`void state;`), and returns `this.store.patchFleet(...)` which re-loads state from disk. This is intentional for idempotency but the `void state` is a silent no-op that suppresses the "unused variable" linter warning instead of just omitting the parameter. More importantly: if `patchFleet` fails or writes a partial patch before throwing, the caller's state reference in `ensureStage1` is already stale — the outer try/catch catches the error, but the returned `fleet_state` in the error envelope will be the pre-call snapshot from `store.load()`, not the partially-written disk state. This is consistent with the existing bootstrap pattern but worth documenting explicitly. The `void state;` anti-pattern also appears at line 1459 in the per-service path. Recommend either removing the parameter (and asserting the Safe address is derivable without it) or documenting why the discard is intentional.

**File:** `client/src/earning/bootstrap.ts`, `stepFleetSafePredict` and one instance in `stepRegisterAgent` area.

#### [Priority: high] Double state load in `ensureStage1And2` after `ensureStage1` returns

`ensureStage1And2` calls `ensureStage1(password)` which may write fleet fields to disk, then immediately calls `this.store.load(this.chain)` again at line ~382 for the Stage 2 body. This is correct — the Stage 2 master-funding gate uses the fresh `state.master_address`. However, if `ensureStage1` returns `ok: true` for an already-complete fleet (the short-circuit path), the second `store.load` will still execute and trigger `applyNghfMigration` again (harmless but redundant). The real risk is that `ensureStage1` returned a `fleet_state` from its own `store.load` call, and Stage 2 code relies on the separately-loaded `state` — if these two diverge (impossible today but fragile if the store grows async writers), it creates a TOCTOU window. Consider threading `stage1Result.fleet_state` into Stage 2 to eliminate the second load, or add a comment acknowledging the intentional re-read.

**File:** `client/src/earning/bootstrap.ts`, `ensureStage1And2`, line ~159.

#### [Priority: medium] ETH funding gate in `ensureStage1` uses `minEoaGasEth * 2n` but Stage 1 only needs agent-side ETH

The Stage 1 funding check requires `this.config.minEoaGasEth * STANDARD_MASTER_BOOTSTRAP_MULTIPLIER` (2x) from the *master* EOA. Then `stepFleetSafeDeploy` separately tops up the agent EOA to 0.01 ETH. The `STANDARD_MASTER_BOOTSTRAP_MULTIPLIER` was originally sized for Stage 2 (full service bootstrap). For Stage 1, master only needs to forward 0.01 ETH to the agent EOA, which is likely to be lower than `2 * minEoaGasEth` on testnet. The comment says "2x for safety" but does not explain how 2x relates to the Stage 1 transaction costs (Safe predict is free; Safe deploy ~250k gas; register ~80k; setAgentWallet ~200k — all from the agent EOA, not the master). If `minEoaGasEth` is the 0.005 ETH default, the 2x gate is 0.01 ETH which just barely covers the 0.01 ETH agent top-up. A user who sends exactly 0.01 ETH to master will pass the gate but the transfer + deploy gas will leave master at ~0 ETH with no buffer. Consider documenting the accounting or reducing Stage 1's master requirement to `STAGE1_AGENT_ETH + minEoaGasEth` (forward cost + one tx gas reserve).

**File:** `client/src/earning/bootstrap.ts`, `ensureStage1`, lines ~63–82.

#### [Priority: medium] `applyNghfMigration` runs on every `parseFleetStateJson` call including read-only status paths

`applyNghfMigration` is invoked inside `parseFleetStateJson` (line 141 in `store.ts`), which is called from `parseFleetStateOrNull` → `load()`, `loadForStatus()` (read-only CLI tools), and from the `status` command path. The migration is pure (no side effects, no disk writes), so this is technically safe. But it means every read of the state file — including `jinn status`, `jinn fleet status`, and hot-polling loops — applies the migration transform in memory without persisting it. If a status read fires before the daemon's first `patchFleet` call, the in-memory fleet state has promoted fleet identity but the on-disk file still lacks `fleet_*` fields. A subsequent `store.load()` in `ensureStage1` will re-run the migration and produce the same in-memory result, but any tool that diffs the disk file against the in-memory state (e.g. a future backup or state-export command) will see them diverge. Low risk today, worth a comment.

**File:** `client/src/earning/store.ts`, `applyNghfMigration`, lines 106–131.

#### [Priority: low] Missing test coverage for crash-between-mint-and-bind in `stepFleetIdentityRegister`

The comment at line 341 notes that `fleet_agent_id` is persisted before `setAgentWallet` to survive a crash between the two. There is no test that simulates `bindAgentWalletToSafe` throwing after `patchFleet` succeeds and verifies that a subsequent `ensureStage1` call skips the mint (because `fleet_agent_id` is now set) and retries only the bind. The existing stage1 tests mock all three steps atomically. This is the highest-risk crash window in Stage 1 — adding one test for the mid-step resume would close the gap.

**File:** `client/test/earning/staged-bootstrap-stage1.test.ts` — gap.

ritsuKai2000:
Cross-referencing concern from the RCA on bd `jinn-mono-hjex.4` / GitHub #233 sub-issue 9 (`setAgentWallet` binding reverts on every fresh v0.1.5 bootstrap).

That bead's RCA enumerated three hypotheses for the binding failure. The **primary** one:

> A `fleet_agent_id` build puts the daemon into a mode where the NFT minter (Stage 1 / fleet-level signer) is not the same EOA as `setAgentWallet`'s `msg.sender` (the per-service `deriveAgentSigner(svc.index)`). The contract reverts with `"Not authorized"` from the `ownerOf / isApprovedForAll / getApproved` gate; viem's simulation surfaces it as a generic `"execution reverted"` because the daemon swallows the revert reason today.

That can't be the cause for the v0.1.5 operator (this PR hasn't merged), but it's worth verifying before this lands that the Stage 1 mint and the per-service binding step use the same EOA — otherwise we'd ship a deterministic version of the binding bug that PR #237's retry workaround can't paper over (it retries the same revert, not a different sender).

Concrete asks:

1. **Confirm**: does `ensureStage1` use `deriveAgentSigner(svc.index)` (or whatever the per-service binding sender becomes after this PR) to mint the agent NFT? If a fleet-level signer mints, the per-service binding step needs `isApprovedForAll` or `getApproved` to be wired on the NFT before `setAgentWallet`.
2. **Add a precondition assertion** before `setAgentWallet`: `ownerOf(agentId) === bindingSenderEoa || isApprovedForAll(ownerOf(agentId), bindingSenderEoa) || getApproved(agentId) === bindingSenderEoa`. Surface a typed error envelope if the precondition fails — cheaper than letting it revert generically and dragging in the `jinn-mono-hjex.4` diagnosability work to figure out what happened post-mortem.
3. Once `jinn-mono-hjex.4` (diagnosability) lands, re-test this PR with intentional Stage 1 / binding misalignment to confirm the resulting error envelope cites the actual contract revert reason.

Refs: bd `jinn-mono-hjex.4`, dogfood report #233 sub-issue 9, RCA carried out 2026-05-14 against the v0.1.5 dogfood session.

ritsukai (approved):

PR #213, opened 2026-05-14
feat(1pbc): jinn solver-plugins publish + revoke — on-chain plug-in registry via setMetadata

ritsuKai2000:
Implements [#202](https://github.com/Jinn-Network/mono/issues/202) (\`jinn-mono-1pbc\`) — third child of the [Plug-in builder entry point epic](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`).

Stacked on [#212](https://github.com/Jinn-Network/mono/pull/212) (\`nghf\` staged bootstrap) → [#210](https://github.com/Jinn-Network/mono/pull/210) (\`et6s\` scaffold) → [#209](https://github.com/Jinn-Network/mono/pull/209) (epic docs).

## Summary

Two new \`jinn solver-plugins\` sub-verbs — \`publish <source>\` and \`revoke <pluginCid>\` — that write \`plugin:<cid>\` metadata records on the existing ERC-8004 IdentityRegistry via the fleet's Stage 1 identity Safe (\`fleet_safe_address\`, established by [#212](https://github.com/Jinn-Network/mono/pull/212)).

**Publish:** resolve plug-in source → pack tarball → compute sha256 → lazily call \`bootstrapper.ensureStage1(password)\` → upload tarball to Autonolas IPFS gateway → encode \`PLUGIN_PAYLOAD_TUPLE\` per spec §5.2 → send \`setMetadata(builderAgentId, "plugin:<cid>", payload)\` via \`executeSafeTransaction\`.

**Revoke:** mirrors publish without the tarball pin; writes a v2 revoked-marker payload (\`{version: 2, revoked: true, reason}\`) to the same metadata key. Indexer interprets the most recent value as authoritative.

Both verbs are **builder actions** — they lazily complete Stage 1 if needed but never touch Stage 2 (OLAS service / staking).

## What ships

- \`client/src/erc8004/abis.ts\` — \`PLUGIN_PAYLOAD_TUPLE\` (\`{version, pluginName, pluginVersion, pluginSha256, supports[], publishedAt}\`) + \`REVOCATION_PAYLOAD_TUPLE\` (\`{version, revoked, reason}\`) per spec §5.2.
[...]

Issue #203, opened 2026-05-13
Ponder indexer: PluginPublication entity + Envelope.executor.plugins[] join + PublishedArtifact base interface


ritsukai:
**bd:** `jinn-mono-attd` · **Epic:** #199 · **Spec:** `docs/superpowers/specs/2026-05-13-plug-in-builder-entry-point-design.md` · **Plan:** `docs/superpowers/plans/2026-05-13-plug-in-builder-entry-point.md`

---

Part of epic jinn-mono-52x3. Extend the Ponder indexer (per jinn-mono-280n) to index `plugin:<cid>` MetadataSet events and join them to existing envelope plug-in attributions, per spec §5.6 + §6.4.

## Why

Once 52x3.3 lands, `MetadataSet(agentId, "plugin:<cid>", payload)` events flow on chain. The indexer needs to decode them into a queryable `PluginPublication` entity, join them to `Envelope.executor.plugins[].cid` for builder attribution, and aggregate per-(builderAgentId, pluginCid) score history. This is the read substrate for 52x3.5 (Discovery API) and 52x3.6 (`/build` SPA panels).

## Acceptance criteria

- A new Ponder entity `PluginPublication` derived from `MetadataSet` events where the `metadataKey` matches `plugin:*`. Fields: `builderAgentId`, `pluginCid` (from key), `pluginName`, `pluginVersion`, `pluginSha256`, `supports[]`, `publishedAt`, `revoked` (most-recent value wins), `revokedReason?`, `txHash`, `blockNumber`.
- A `PublishedArtifact` base interface (per spec §5.6) — `PluginPublication` declares `artifactType: 'plugin'` and extends it. Future `harness:<cid>` kind extends the same base.
- A join derivation: `BuilderAttributedRun` (or similar) row keyed on `(envelopeCid, pluginCid)`. Computed from each envelope's `executor.plugins[]` joined against `PluginPublication.pluginCid`. Sha256 mismatch flags the run as `forkSuspected: true` and falls back to no-builder-attribution per spec §5.3.
- Aggregations: per-`(builderAgentId, pluginCid)` score history (list of `(operator, taskId, verdict, score, ts)`) and rollup (count, p50, p90). Composes with ebu7's existing HarnessCheckpoint rollups (joins on `pluginCid` via the existing envelope→HarnessCheckpoint→Verdict path).
[...]

PR #214, opened 2026-05-14
feat(attd): Ponder indexer — PluginPublication entity + plugin:* MetadataSet handler + PublishedArtifact base

ritsuKai2000:
Implements [#203](https://github.com/Jinn-Network/mono/issues/203) (\`jinn-mono-attd\`) — fourth child of the [Plug-in builder entry point epic](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`).

Stacked: #213 (1pbc) → #212 (nghf) → #210 (et6s) → #209 (epic docs).

## Summary

Extends the Jinn Ponder indexer (\`packages/indexer/\`) to decode \`IdentityRegistry.MetadataSet\` events whose key starts with \`plugin:\` (emitted by #213's \`solver-plugins publish\`). Surfaces the on-chain registry through the Discovery API as a typed \`PluginPublication\` shape extending a new \`PublishedArtifact\` base. Adds an attribution-join read shape (\`BuilderAttributedRun\`) ready for ebu7's envelope-plugin metadata to wire builder reputation when that merges.

## What ships

**Indexer (\`packages/indexer/\`):**
- \`pluginPublication\` onchain table in \`ponder.schema.ts\` per spec §5.6 (builderAgentId, pluginCid, pluginName, pluginVersion, pluginSha256, supports[], publishedAt, revoked, revokedReason, provenance fields).
- \`parsePluginKey\` parser + \`decodePluginPayload\` / \`decodeRevocationPayload\` decoders (PLUGIN_PAYLOAD_TUPLE / REVOCATION_PAYLOAD_TUPLE duplicated from \`client/src/erc8004/abis.ts\` since cross-package imports aren't wired).
- \`handleMetadataSet\` extended with a \`plugin:\` branch — v1 publishes insert with \`onConflictDoUpdate\` most-recent-wins by \`(block, txIndex, logIndex)\`; v2 revocations flip the existing row's \`revoked\` flag with the same recency gate.
- \`attributeRuns\` join with sha256-mismatch fork-detection (forks flagged, attribution skipped).
- \`/builders/:agentId/runs\` Hono route returning \`BuilderAttributedRun[]\`; returns empty until ebu7's \`attemptEnvelopeMeta\` exists at runtime (filed as follow-up \`jinn-mono-pm7c\`).

**Client (\`client/\`):**
- \`PublishedArtifact\` base + \`PluginPublication\` shape in \`client/src/discovery/types.ts\` (future Path 2 \`harness:<cid>\` extends the same base — filed as follow-up \`jinn-mono-6jid\`).
- \`DiscoveryAPI\` interface gains \`listPluginPublications\`, \`getPluginScores\`, \`listBuilderArtifacts\`.
- \`HttpDiscoveryAPI\` GraphQL implementations.
- \`OnchainDiscoveryAPI\` empty-array fallbacks (the floor; the HTTP indexer is the primary path).
- \`withFallback\` dispatcher forwards the three new methods.

## Test plan

- [x] \`packages/indexer\` typecheck clean
- [x] \`packages/indexer\` tests: 47 pass (5 parsePluginKey + 2 schema + 12 plugin handler + others; 6 test files)
- [x] \`client\` typecheck clean
- [x] \`client\` tests: 3269 pass (3266 pre-attd + 3 new discovery tests)
- [x] \`client\` build clean
- [x] Most-recent-wins overwrite semantics tested (later-block overwrites earlier, earlier-block no-ops, same-block logIndex tiebreak, idempotent replay)
- [x] Fork-detection: sha256 mismatch flags \`forkSuspected\` and skips attribution per spec §5.3
- [x] Graceful degradation: \`/builders/:agentId/runs\` returns empty when \`attemptEnvelopeMeta\` entity isn't deployed (ebu7 dependency)

## Follow-ups filed

- **\`jinn-mono-pm7c\`** — wire \`attributeRuns\` to ebu7's \`attemptEnvelopeMeta.pluginsJson\` once that lands.
- **\`jinn-mono-6jid\`** — \`PublishedArtifact\` v0.2 extension for Path 2 \`HarnessPublication\` (sibling kind to \`PluginPublication\`).

## Reference

- Parent epic: [#199](https://github.com/Jinn-Network/mono/issues/199)
- Plan: \`cargo/docs/superpowers/plans/2026-05-14-attd-ponder-plugin-publication-plan.md\`
- Spec §5.6 (read layer + PublishedArtifact), §6.4 (this bead): \`cargo/docs/superpowers/specs/2026-05-13-plug-in-builder-entry-point-design.md\`


ritsuKai2000 (commented):
## Review (agent: substrate)

**Verdict:** approve-with-comments

### Strengths

- The `pluginPublication` entity schema is carefully documented: the `id` composite key `<builderAgentId>:<pluginCid>` makes most-recent-wins semantics explicit, the `txIndex` + `logIndex` tiebreak columns are persisted for the ordering gate, and the `pluginSha256` column is stored as `text` (not `hex`) specifically because it serves as a fork-attribution join key — the comment explains this choice.
- The drift-guard pattern (duplicate tuple constants in `packages/indexer/src/handlers.ts`, guarded by a test that encodes with the client-side tuple and decodes with the indexer-side tuple) is the right approach given the cross-package boundary. The two test files (`handlers.plugin.test.ts`, `helpers/events.ts`) make the guard self-verifying.
- The onchain floor stubs (`listPluginPublications`, `getPluginScores`, `listBuilderArtifacts` all return `[]`) are semantically correct — they reflect the genuine limitation that enumeration requires the indexer — and they satisfy the `DiscoveryAPI` interface without silently dropping data.

### Findings

#### [Priority: high] `getPluginScores` silently returns stub rows with empty `taskId` / `operatorAgentId`

In `client/src/discovery/http.ts`, `getPluginScores` flatMaps `attemptEnvelopeMetas` items into `PluginScoreHistoryRow` objects where `taskId: ''` and `operatorAgentId: ''` are hardcoded. The comment acknowledges this is a placeholder until the `/builders/:agentId/runs` Hono route ships with ebu7, but the method signature on `DiscoveryAPI` and `PluginScoreHistoryRow` does not mark these fields as optional. Consumers that call `getPluginScores` today will receive rows that look populated but have semantically empty `taskId` and `operatorAgentId`. If the SPA renders these rows (e.g. in a score history table), it will show blank operator IDs with no obvious indication they are stubs. Recommend either: (a) returning `[]` unconditionally until the join is implemented (the onchain floor already does this), or (b) marking `taskId`/`operatorAgentId` as `string | ''` with a note in the interface, or (c) adding a `partial: true` flag to the row shape. The current behaviour is a silent data quality issue.

**File:** `client/src/discovery/http.ts`, `getPluginScores`, lines ~154–167.

#### [Priority: medium] `v2` revocation path re-reads the row from the DB even when `v1` decode succeeds

In the handler branch inside `packages/indexer/src/handlers.ts`, the logic is:
```
const v1 = decodePluginPayload(...)
const v2 = v1 ? null : decodeRevocationPayload(...)
```
This is correct and avoids the extra decode when `v1` succeeds. However, the v2 path then calls `await context.db.find(pluginPublication, { id })` before the most-recent-wins gate, meaning a revocation against a non-existent key (e.g. a revocation replayed before the original publish due to indexer backfill ordering) silently no-ops. This is the correct semantic per the comment ("a revocation for a never-published key is meaningless"), but it means a re-sync that replays events out of order could lose the revocation permanently: if the publish event arrives later it will insert a fresh row with `revoked: false`, and the revocation event won't be replayed again. Ponder resyncs are full replays from genesis so this should not occur in practice, but the assumption is implicit. Document it, or add an assertion in the test for out-of-order replay.

**File:** `packages/indexer/src/handlers.ts` (inferred from plan/diff), revocation branch.

#### [Priority: medium] `listBuilderArtifacts` delegates entirely to `listPluginPublications` without passing `limit`

In `client/src/discovery/http.ts`, `listBuilderArtifacts` calls `listPluginPublications({ builderAgentId: args.builderAgentId, limit: args.limit })` — this correctly threads the limit. However, the `PluginPublicationRow` → `PluginPublication` mapping in `listPluginPublications` is also responsible for mapping `publishedAt` as `Number(row.publishedAt)`. If the indexer returns `publishedAt` as a `bigint` (Postgres `bigint` → JS `bigint` in some ORMs), `Number()` will silently lose precision for values above `Number.MAX_SAFE_INTEGER`. Since `publishedAt` is unix seconds (uint64), it will be in the year ~292 billion before this matters — no real risk today. But `pluginPublication.publishedAt` is declared as `t.bigint()` in the schema; verify whether Ponder's GraphQL layer serialises `bigint` columns as strings (the `publishedAt: string | number` union type on `PluginPublicationRow` suggests it does). If the server returns a string, `Number()` is safe; if it returns a JS bigint, use `Number(BigInt(row.publishedAt))` with an explicit comment.

**File:** `client/src/discovery/http.ts`, `listPluginPublications`, line ~116.

#### [Priority: low] `DiscoveryAPI.artifactType` discriminator is locked to `'plugin'` on the `PublishedArtifact` base

`PublishedArtifact.artifactType` is typed as `'plugin'` (literal, not a union). The comment in `types.ts` says "future: `| 'harness'`" but the discriminator on the base type will need to become `'plugin' | 'harness'` when harness publishing lands, which is a breaking change to the interface (all callers that narrow `widened.artifactType === 'plugin'` will still work, but anything that assigns `artifactType: 'harness'` to a `PublishedArtifact` will fail to type-check). Consider widening to `string` on the base or `'plugin' | string` now so the future addition is backwards-compatible, and narrow to `'plugin'` only on `PluginPublication`. Low urgency — a bead for the harness epic.

**File:** `client/src/discovery/types.ts`, `PublishedArtifact.artifactType`.

ritsukai (approved):

PR #224, opened 2026-05-14
test(r83r): reference plug-in (swe-rebench-v2-diffmin) + cold-start E2E acceptance gate

ritsukai:
Implements [#206](https://github.com/Jinn-Network/mono/issues/206) (\`jinn-mono-r83r\`) — **seventh and final** child of the [Plug-in builder entry point epic](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`).

Stacked: #216 (hfmf) → #215 (ttz8) → #214 (attd) → #213 (1pbc) → #212 (nghf) → #210 (et6s) → #209 (epic docs).

## Summary

Ships the closing evidence for the epic. Two pieces:

1. **Reference competing plug-in** \`client/plugins/swe-rebench-v2-diffmin/\` — a real working plug-in for \`swe-rebench-v2.v1\` that demonstrates a **minimal-diff + test-mapping** angle (complementary to \`swe-rebench-v2-runtime\`'s orient/plan angle). Includes a \`diff-stats\` MCP server that computes line/file/test/non-test metrics on candidate diffs to bias the solver toward parsimony.

2. **Cold-start E2E acceptance gate** at \`client/test/acceptance/cold-start-builder.test.ts\` — walks the nine-step builder loop end-to-end:
   - Scaffold via \`jinn create plugin\` (from et6s)
   - Pack via existing \`jinn solver-plugins pack\`
   - \`jinn solver-plugins publish\` (from 1pbc) against an Anvil-fork IdentityRegistry stub + a stub IPFS gateway
   - Stub indexer's MetadataSet watcher creates the PluginPublication row
   - DiscoveryAPI \`listPluginPublications\` surfaces the new plug-in
   - Operator installs via \`jinn solver-nets add-plugin\`
   - Stub-Hermes runs a SWE-rebench v2 task with the plug-in loaded; envelope emits \`executor.plugins[]\` carrying attribution
   - Indexer joins envelope to publication via the \`recordRunForTest\` side-channel
   - \`/build\` SPA panels show the new plug-in (skipped gracefully in fork pool — covered by jsdom tests from #216)

3. **Dual-role path test** — operator (Stage 1+2 complete) publishes a plug-in; \`countRegisterTxs\` asserts only **1** \`AgentRegistered\` event across both \`ensureStage1And2\` and \`publish\` on the same identity. No re-mint, no second Safe, no second agentId.

## What ships

**Reference plug-in (\`client/plugins/swe-rebench-v2-diffmin/\`):**
[...]

PR #226, opened 2026-05-14
chore(52x3): simplify epic surface (post-merge cleanup)

ritsuKai2000:
Closeout PR for the [Plug-in builder entry point epic](https://github.com/Jinn-Network/mono/issues/199) (\`jinn-mono-52x3\`). Per the epic plan's Task 8 — runs the \`/simplify\` skill across the cumulative diff and applies findings.

Stacked on [#224](https://github.com/Jinn-Network/mono/pull/224) (r83r) — the final critical-path PR.

## Review process

Three independent reviewer agents (code reuse, code quality, efficiency) reviewed the 6157-line production-code diff across 75 files. Three claims were false positives (engine.ts not in our diff; indexer plug-in handler does exist at \`handlers.ts:494+\`; PythonEvalRunner LRU change came from uy6v.11 not us). Eight real findings applied.

## Fixes applied (8 commits, one per fix)

| Commit | Fix |
|---|---|
| \`9d47efb9\` | Export \`parseRegistryUploadCid\` from \`ipfs.ts\`; \`ipfs-pinfile.ts\` reuses it (was a 16-line duplicate). |
| \`93148db1\` | Share \`writeJson\` across \`solver-plugins.ts\` + \`solver-plugins-publish.ts\` + \`solver-plugins-revoke.ts\` (was identically declared in all three). |
| \`f773147d\` | \`ensureStage1And2\` reuses \`stage1Result.fleet_state\` instead of re-loading from disk immediately after \`ensureStage1\` returned it. |
| \`8cafbfb3\` | publish handler uses \`loaded.manifest\` from \`resolveSolverPlugin\`; removes redundant second \`loadSolverPluginManifest(loaded.root)\` call. |
| \`b98f2c6c\` | \`ipfs-pinfile.ts\` passes the Blob directly to \`formData.append\` (drops \`arrayBuffer()\` round-trip); removes TOCTOU \`statSync\`. |
| \`eadb8635\` | \`PLUGIN_METADATA_KEY_PREFIX = 'plugin:'\` constant exported from \`plugin-registry.ts\`; indexer-side \`PLUGIN_KEY_PREFIX\` documents the cross-package canonical. |
| \`946d5c06\` | \`PRODUCTION_DEPS.publisherFactory\` constructs \`PluginRegistryPublisher\` once (lazy via \`publisherPromise\`) — shared by both \`publish\` and \`revoke\` calls. |
| \`c4a254cd\` | Indexer \`/plugins\` route pushes \`builder\` + \`includeRevoked\` + \`solverNet\` filters into the DB query via Drizzle's \`arrayContains\` — replaces full-table scan + TS filter. |

## Findings deferred (not applied this PR)

- **\`SectionCard\` extraction across Build page panels** — component refactor; separate PR if desired.
- **\`ArtifactTypeFilter\` dead state** — intentional future-proofing per spec §5.6 for Path 2 \`harness\` artifact type. Documented but not removed.
- **\`/v1/discovery/*\` auth gate** — needs Captain to confirm intent (private SPA-only vs intentionally public). Filed as follow-up consideration.
- **Bead-ID inline comments** — high volume; would benefit from a separate scrub pass.
- **Cross-package type duplication** (\`PluginPublication\` in client + \`PluginPublicationOutput\` in indexer) — needs a shared types package (separate epic).

## Test plan

- [x] \`yarn typecheck\` clean (client + indexer)
- [x] \`yarn test\` — **3334 client pass**, 65 indexer pass, 0 regressions, 7 pre-existing skips
- [x] \`yarn build\` clean
- [x] \`yarn e2e:cold-start-builder\` — 10 tests pass in ~14s


ritsuKai2000 (commented):
## Review (agent: integration)

**Verdict:** approve

### Strengths

- The eight commits are clearly scoped and each addresses a real issue: deduplicating `writeJson`, deduplicating `parseRegistryUploadCid`, collapsing a double manifest load in `publishHandler`, sharing a lazy `PluginRegistryPublisher` instance across `publish`/`revoke`, exporting `PLUGIN_METADATA_KEY_PREFIX` as a constant, and pushing DB filters into the SQL query in the indexer API.
- The DB filter push in `packages/indexer/src/api/index.ts` is the most impactful change: fetching the full `pluginPublication` table before filtering in-process is O(n) unbounded; pushing `arrayContains`/`eq` conditions into the query is the correct fix. The builder branch gains the same treatment symmetrically.
- The `bootstrap.ts` fix (`let state = stage1Result.fleet_state` instead of re-loading from the store) eliminates a redundant disk read and the associated TOCTOU risk between Stage 1 completing and Stage 2 starting.

### Findings

#### [Priority: medium] Losing file-size in the error message for IPFS upload failures is a mild regression

`ipfs-pinfile.ts` line 60: the error message previously included `(${stat.size} bytes)` for IPFS upload failures. The simplify commit removes the `statSync` call (correct — avoids two fs ops) but drops the size info. For debugging production upload failures, the size context is useful. Consider reading the blob's `.size` property instead (already materialised into memory at that point): `${blob.size} bytes`.

#### [Priority: low] `includeRevoked: true` passed to `listPluginsByNetwork`/`listPluginsByBuilder` after already filtering in SQL

After the DB filter push, lines 261 and 272 pass `includeRevoked: true` to the in-memory list functions. This is correct (the filter was already applied in SQL), but the parameter name `includeRevoked: true` at the call site is now misleading — the post-filter functions always receive pre-filtered rows. A follow-on cleanup could remove the parameter from `listPluginsByNetwork` / `listPluginsByBuilder` since it's now redundant, or rename the call-site argument to make the intent clear (`includeRevoked: true /* already filtered in DB */`). Not a bug, but worth cleaning up in the Task 8 simplify pass.

ritsukai (approved):

Issue #236, opened 2026-05-14
Daemon-level Claude auth gate is in the wrong layer: should be per-harness, not a hard global preflight that exits the process


ritsuKai2000:
## Context

The daemon currently treats Claude auth as a global pre-running-mode hard gate. If `claude auth status` reports `loggedIn: false` after fleet bootstrap completes, the daemon exits with `exitCode=11` and tells the operator to "Complete Claude setup in the operator app, then restart the daemon." See `client/src/preflight/claude-auth.js`, `client/src/main.ts:1210-1211` (`configRequiresClaudeAuth(config)`), and the post-bootstrap path that fires the exit.

This shape was originally surfaced as a "make the daemon wait for Phase 1 sign-in" issue (jinn-mono-ky25 — now superseded by this bead). The deeper architectural problem: **the gate is in the wrong layer**.

## Problem

1. **Claude is one harness among several.** The daemon supports multiple harnesses today (claude-code-learner, codex-code-learner, swe-rebench-v2-evaluator — a Docker-based harness that doesn't use Claude at all). Per-harness auth is the right shape; gating the entire daemon on one specific harness's auth state is wrong.

2. **The embedded agent WS bridge is the only daemon-side surface that actually requires Claude.** The xterm.js-backed `agent-ws.ts` session in the dashboard SPA spawns Claude as a persistent subprocess for the operator's interactive workspace. That's the surface that needs authenticated Claude. The solver/evaluator runtime loops are harness-specific — a Codex-only operator never needs Claude auth; an evaluator-only operator running swe-rebench-v2 in Docker never needs Claude auth.

3. **The gate prevents the daemon from starting at all** for any operator who hasn't separately authenticated Claude — even operators who intend to run pure Codex flows or Docker-only evaluator flows. That's an unforced error: making non-Claude operators set up Claude as a prerequisite for participating.

## Impact

- Codex-only operators are blocked from running the daemon by a Claude-specific gate that doesn't apply to their harness.
- Evaluator-only operators (Docker + swe-rebench-v2-evaluator) are blocked from running the daemon by a gate that doesn't apply to their harness.
- The mixing of "daemon health" with "one specific harness's health" makes the v1 operator install path acceptance (jinn-mono-uy6v.4) artificially fragile — the install path "works" only for the subset of operators who happen to use Claude AND have already OAuth'd it on their machine.
[...]

Issue #160, opened 2026-05-12
Operator install path: npm install -g @jinn-network/client → joined SolverNet → first task settled, zero CLI workarounds


ritsuKai2000:
Part of release epic jinn-mono-uy6v (v1 first public release).

## Intent

The SWE-rebench v2 equivalent of the Prediction operator-dogfood gate (jinn-mono-l2zl.15, now deferred per Prediction freeze). A human starts from clean state, installs the npm package, runs `jinn run`, joins the canonical SWE-rebench v2 SolverNet via /operator, and gets their first Solution settled — entirely through the dashboard, with no CLI workarounds.

This bead pulls in the SolverNet-agnostic operator-app blockers from the Prediction dogfood subtree as release blockers:
- jinn-mono-l2zl.15.4.1 — default operator config must remain testnet before mainnet launch.
- jinn-mono-l2zl.15.4.3 — operator onboarding must start and continue from `jinn run` only.
- jinn-mono-l2zl.15.4.4 — `jinn run` must migrate existing local DB before startup.
- jinn-mono-l2zl.15.4.5 — `jinn run` must not load repo dev env into operator startup.
- jinn-mono-l2zl.15.4.6 — Claude install flow must update live app auth probe.
- jinn-mono-l2zl.15.4.7 — Prelaunch testnet guard must not reuse mainnet earning state.

## Acceptance criteria (testable, problem-shaped)

- A clean-machine user runs `npm install -g @jinn-network/client@latest`, then `jinn run`, with no manual config file editing.
- The dashboard opens automatically (or the URL is printed clearly in stdout).
- The user joins the canonical SWE-rebench v2 SolverNet from /operator via Discover.
- The user's first Solution settles on chain with evidence visible in /launcher/launched/:id or equivalent.
- Documented walk-through retained as release evidence (the SWE-rebench-v2-flavored equivalent of l2zl.15's Prediction dogfood runbook).

## Reference
- Pulled-in blockers: jinn-mono-l2zl.15.4.{1,3,4,5,6,7}
- Runbook §Preconditions, §Join As Solver Or Evaluator: docs/runbooks/swe-rebench-v2-public-testnet.md

---
Internal tracking: jinn-mono-uy6v.4

ritsuKai2000:
## 2026-05-14 cleanroom walkthrough — partial verification

[...]

PR #237, opened 2026-05-15
fix(h74p): retry safe-binding to absorb fresh-Safe race window

ritsuKai2000:
## Summary

- `IdentityRegistry.setAgentWallet` first-attempt reverts against fresh 1/1 Safes on Base Sepolia → retry the same call 2 more times (default: 3 attempts × 3 s delay) inside `stepRegisterAgent` so first-run operators don't land at `safe_bound_to_agent=false`
- Bound retry budget: real failures still surface in the existing `safe_binding_pending` persisted state, just after N attempts instead of 1
- New `FleetBootstrapperOptions` knobs: `safeBindingMaxAttempts`, `safeBindingRetryDelayMs` (for test-fast paths + operator overrides)

## Context

The walkthrough that closed `jinn-mono-uy6v.4` (#160) reframed `h74p` from "100% deterministic revert" to "first-attempt revert, deterministic retry-success on next `jinn run`." Today that retry only happens because the daemon exits on missing Claude auth and the operator restarts — bootstrap resumes at `safe_binding_pending` and the second attempt against the same Safe + agentId succeeds.

Once `jinn-mono-vh74.2` (#236) reshapes the Claude-auth gate so the daemon doesn't exit, that accidental safety net disappears and operators would sit at `safe_bound_to_agent=false` until the bootstrap loop's next natural re-tick — with downstream surfaces (claim flows, attestations) silently misbehaving in between. Per the bead's original impact statement: "Downstream surfaces that expect the binding to be present silently misbehave."

In-process retry makes the operator-visible behaviour deterministic.

## Test plan

- [x] New regression: `agent_registered step retries setAgentWallet on transient failure and succeeds (jinn-mono-h74p)` — mocks first call to reject, second to resolve, asserts 2 calls + final state success
- [x] New regression: `agent_registered step gives up after maxAttempts and persists failure (jinn-mono-h74p)` — asserts 3 calls + falls through to existing failure persistence
- [x] Existing `agent_registered step does not block bootstrap when bind reverts` updated to pass `safeBindingRetryDelayMs: 0` so it doesn't time out on the new default retry budget
- [x] `yarn vitest run test/earning/` — 163/163 pass, 20 files
- [x] `yarn typecheck` — clean

Closes jinn-mono-h74p (after merge).


ritsukai (approved):
Approved on behalf of @oaksprout (Captain) per session 2026-05-15. Captain authored and reviewed this fix; approval mechanic via secondary agent account because PR author cannot self-approve. This is also the smoke-test PR for the new two-train flow — first PR merged to next under the post-migration canary path.

PR #239, opened 2026-05-15
fix(hjex.2): diagnostic pipeline reads joinedSolverNets

ritsuKai2000:
## Summary

- `prediction_solvernet_missing` no longer fires when the operator has joined a SolverNet via the new manifest-keyed flow.
- `jinn solver-nets list` enumerates both `solverNets` (legacy) and `joinedSolverNets` (new) shapes with a `source` tag.
- `gather-status` derives the SolverNet name from config rather than hard-coding `'prediction'`.

## Why

Bead `jinn-mono-hjex.2` / GitHub #233 sub-issue 10: every operator who joins a SolverNet via the v0.1.5 Settings UI sees `NEEDS ATTENTION  No active SolverNet configured` on the same screen as `JOINED · 1 · SWE-rebench v2`. The diagnostic pipeline reads only the legacy `solverNets[<shortName>]` shape; the SPA writes only the new `joinedSolverNets[<manifestCid>]` shape.

## Stacked on

PR #238 (PR-1 of the same stack).

## Test plan

- [x] New regression: `prediction_solvernet_missing` does NOT fire when `joinedSolverNets` has an entry.
- [x] `jinn solver-nets list` enumerates both shapes.
- [x] `gather-status` test confirms name is derived from config.
- [x] `yarn typecheck` + `yarn test` clean.

Refs: bd `jinn-mono-hjex.2`.

Issue #233, opened 2026-05-14
Operator dogfood — v0.1.5 upgrade run: 18 issues from 2026-05-14 session


oaksprout:
# Operator dogfood — 2026-05-14 v0.1.5 upgrade run

External-operator walkthrough of `@jinn-network/client@0.1.5` (the first
release where the operator app is the way you participate). Author: dogfood
tester upgrading an existing v0.1.3-canary operator on Base Sepolia, with
protocol-team keys alongside.

Pin: `@jinn-network/client@0.1.5`.

## Summary

- Full loop landed?: **no.** Bootstrap halted twice; second halt is the current
  state at the time of writing.
- Time from `jinn update` to running daemon: **never reached running.**
- Two distinct upgrade-path failures (#1, #2) and three UI / copy issues (#3,
  #4, #5) surfaced in a single 30-minute session.

### 1. `jinn update` doesn't stop the running daemon — orphan blocks `jinn run`

**What happened.** Existing v0.1.3-canary daemon (PID 19958, started ~29h
prior) was running and serving on port 7331. Ran `jinn update` cleanly
(`npm-update` + `config-migration` + `task-native-store` + `integrations-install`
all `ok`). Immediately tried `jinn run` on the new binary: exited with
`EADDRINUSE` on 7331.

`jinn stop` reported `state: stopped, pid: null` — it doesn't track daemons
whose pidfile is missing or orphaned, so it cannot clean them up. The
operator has to discover the orphan via `lsof -i :7331`, identify it as
theirs, and `kill` it manually before `jinn run` will succeed.

**Severity.** Friction. Every operator who upgrades while their daemon is
running will hit this.

**Candidate fixes.**
1. `jinn update` should detect a running daemon (via port, pidfile, or `ps`)
   and prompt / auto-restart it on the new binary.
2. `jinn stop` should accept `--force` or `--by-port` that finds a daemon by
   its listening port and offers to kill it.
3. `jinn run`'s `EADDRINUSE` error should name the offending PID and its
   uptime: "Port 7331 is held by PID X (`jinn run`, started Y ago). Run
   `jinn stop --force` or kill it manually."

### 2. v0.1.5 migration can brick a working operator on testnet

**What happened.** After clearing the orphan, `jinn run` started the new
daemon, which auto-ran the `testnet-setup-migration` flow in
`client/src/earning/testnet-setup-migration.ts`. The pre-migration state
was a fully complete v0.1.3 setup:

- `service_id: 47`
[...]

Issue #220, opened 2026-05-14
EPIC: Make operators earn and see tJINN for SWE-rebench v2 work


ritsuKai2000:
## Run-mode

SPIKE → FIX (output of spike defines the next bead)

## Context

Live operator report against v0.1.5: after completing many SWE-rebench v2 tasks on Base Sepolia (solver + evaluator roles, several settled task IDs visible in /launcher/launched/:id and /operator/execution-data), the operator's dashboard reports 0 JINN earned.

Concretely the operator has:
- An active service staked in the SWE-rebench v2 SolverNet
- A produced + donated artifact inventory of 100+ swe-rebench-v2 solutions and 50+ swe-rebench_v2_verdict envelopes on /v1/operator/artifacts
- On-chain claimDelivery transactions for those deliveries (visible in /tmp/jinn-producer.log and the activity surface)
- Reputation-feedback registry updates submitted (post-PR #190 the daemon emits feedback on every verdict; observed `reputation feedback submitted verdict=FAIL` for at least 3 deliveries during v0.1.5 prep)

…and the operator app surfaces 0 JINN earned. Either:

a) JINN reward distribution from the SWE-rebench v2 SolverNet hasn't actually fired yet for this operator (emission epoch / cooldown / threshold issue), or
b) Reward distribution has fired but the operator app is reading the wrong contract / address / event, or
c) The "JINN earned" surface is conceptually disconnected from the path that's actually emitting (e.g. distributor address differs from what the dashboard polls), or
d) The verdicts emitted so far are all FAIL (passed_match: false), and the distribution contract only emits on success — in which case 0 earned is the correct on-chain answer and the UI's bug is that it doesn't explain why.

The investigation has to identify which of these is the root cause before we can shape the fix bead.

## Impact

This is the single most damaging UX failure mode of the v0.1.5 announcement: an operator does the work, completes the loop, and sees 0 reward. Whether the root cause is on-chain or UI-side, the operator's lived experience is "Jinn doesn't pay." Critical for the v1 public-testnet release definition (jinn-mono-uy6v item 3, "JINN reward distribution observed") and for the v0.1.5 "make participating obvious" framing.

## Acceptance criteria (problem-shaped — for the spike)

- Identify which of (a)–(d) above (or another root cause) is responsible, with on-chain evidence: distributor address, emission-event log (or absence), operator agent / safe / EOA mapping to that distributor, balance reads against the actual reward-token contract at the operator's reward-receiving address.
[...]

PR #240, opened 2026-05-15
fix(hjex.3): surface service eviction + in-process auto-restake

ritsuKai2000:
## Summary

- Fleet API now reads on-chain `getStakingState(serviceId)` for each staked service and exposes `staking.evicted` per service in `GET /v1/fleet` (was hard-coded `false`).
- `gather-status` fans out the staking state reads and populates `evictedByServiceIndex` on `GatheredStatusRaw`.
- HeroStats gains an `evicted` prop: when `true`, the "Claim now" CTA is disabled and an inline **"Service evicted — restoring liveness"** notice renders. No OLAS terminology in operator-facing copy.
- New `EvictionLoop` daemon loop polls `getStakingState` every 60 s and calls the extracted `recoverEvictedService` helper when state 2 (Evicted) is detected — no daemon restart required.
- `recoverEvictedService` extracted as a standalone exported function from `FleetBootstrapper` so the loop can call it without a full bootstrapper context. The private class method is unchanged.
- New `POST /v1/setup/restake/:serviceId` endpoint backs the "Re-stake now" dashboard CTA; delegates via a lazy ref populated in running mode.
- New `evictionCheckIntervalMs` config knob (default 60 s, env `JINN_EVICTION_CHECK_INTERVAL_MS`, 0 = disabled).

## Why

Bead `jinn-mono-hjex.3` / GitHub #220 / #233: an operator who completed many SWE-rebench v2 tasks saw `0 JINN claimable` because their service had been silently evicted by the staking proxy (inactivity ratio failure). The dashboard was reading the right contract — the on-chain answer genuinely was 0 — but the operator had no signal that anything was wrong, and "Claim now" silently no-op'd.

Operator-facing copy never mentions OLAS.

## Stacked on

PR #239 (PR-2 of the same stack).

## Test plan

- [x] `fleet-build` regression: `staking.evicted: true` when `evictedByServiceIndex[displayIndex] === true` in `GatheredStatusRaw`.
- [x] `eviction-loop` regression: detects state 2, calls `recoverEvictedService`, skips non-staked services, skips services missing `service_id` or `staking_address`, non-fatal on RPC error, exits immediately when `intervalMs === 0`.
- [x] `HeroStats` regression: "Claim now" disabled + eviction explainer renders when `evicted=true`; neither appears when `evicted=false`.
- [x] `yarn typecheck` + `yarn test` (3281/3281) clean.

Refs: bd `jinn-mono-hjex.3`, GitHub #220, GitHub #233.

### DIFF
--- a/client/src/api/fleet-build.ts
+++ b/client/src/api/fleet-build.ts
@@ -26,7 +26,7 @@ export interface FleetV1Service {
     agent: { address: string; balances: Array<{ asset: string; amountWei: string }> };
     multisig: { address: string; balances: Array<{ asset: string; amountWei: string }> };
   };
-  staking: { staked: boolean; evicted: boolean; sinceBlock: number | null };
+  staking: { staked: boolean; evicted: boolean; sinceBlock: number | null; inactivitySeconds: number | null };
   activity: { lastEventAt: string | null; counts: Record<string, number> };
   rewards: { pending: string; asset: 'reward' };
   attention: null | {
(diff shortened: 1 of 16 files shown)


ritsukai (approved):
Parity-review by independent agent passed (2026-05-19). See review summary on #328. Approving for merge.

PR #262, opened 2026-05-18
fix(u34i): bootstrap reliability — gate+transfer parity, faucet/gate single source, getCode retry, panel auto-continue, harness-readiness holder

ritsuKai2000:
## Summary

End-to-end bootstrap-reliability fix for the Stage 1 → Stage 2 → running-mode flow on Base Sepolia. Four interlocking bugs surfaced on the 2026-05-18 canary; this PR ships all four fixes plus the regression tests that would have caught each one.

bd issue: **jinn-mono-u34i**

## What was broken

1. **Stage 1 funding gate** equaled the Stage 1 transfer amount. The gate required `minEoaGasEth * 2 = 0.01 ETH`; `ensureStage1` then transferred `STAGE1_AGENT_ETH = 0.01 ETH` from master to fleet agent EOA, leaving zero headroom for master's own gas. Master EOAs holding exactly the gate threshold (common with testnet faucet drips) failed the funding tx with `gas required exceeds allowance (0)`. Regression from c56f9f22 (2026-05-14, feat(nghf): ensureStage1 fleet identity walk).
2. **Faucet drip target ≠ daemon gate.** After bumping the daemon gate, the faucet endpoint's target was plumbed via `main.ts:1089` and still computed the old value. Faucet stopped at 0.01 ETH while the daemon waited for 0.015 ETH — deadlock; faucet card said "complete" but panel never advanced.
3. **`gas required exceeds allowance (0)` mis-classified as out-of-gas** by `operator-errors.ts`. Geth-specific signal that allowance (balance ÷ gasPrice) rounded to zero — i.e. insufficient funds, not a gas limit issue. The hint told operators to "re-run; gas estimator drifted past safe-adapter floor", which is wrong.
4. **`getCode` race after Safe deploy.** Public Base Sepolia returns `0x` from a load-balanced node for a block or two after the deploy receipt confirmed — bootstrap threw `deployment succeeded but no code at <addr>`. (wkbw fix, folded in here.)
5. **Hono `Can not add a route since the matcher is already built`** on the running-mode transition. `main.ts` added `/v1/harnesses/readiness` routes post-bootstrap, but Hono had finalized its matcher while serving panel polls. Only reachable now that Bug 1 stopped halting bootstrap.
[...]

PR #275, opened 2026-05-18
test(u34i): tier-2 regression coverage — Playwright sequential-state E2E, late-mount lint, boundary tests

ritsuKai2000:
## Summary

Tier-2 follow-up to #262 (the u34i fix-stack). Each commit closes a specific CI gap that let the 2026-05-18 canary regression ship through unit tests.

Linked: jinn-mono-u34i (the original fix, merged in #262), jinn-mono-j1mc (stale dashboard E2Es — separate cleanup).

### 1. Playwright sequential-state E2E (`test(u34i)`: AwaitingFundingCard auto-continue)

Mocks \`/v1/bootstrap\` to walk three sequential states (Stage 1 gate → Stage 2 gate → running mode) plus the \`/v1/setup/drip\` endpoint. Asserts the operator clicks FUND FROM FAUCET **once**, both drip POSTs fire (the second without operator interaction), and the page lands on \`/overview\`.

Red-green verified: removing the \`useEffect\`-on-\`minimumWei\` block in \`AwaitingFundingCard\` makes the page never reach \`/overview\` within 20s, and the test fails on the URL assertion.

Adds a \`funding-sequence-e2e\` CI job that runs this test on every PR (~4s, chromium-only). Scoped to this one test; the other dashboard E2Es have stale failures (jinn-mono-j1mc).

### 2. Late-mount lint guard (`test(u34i)`: Hono late route mount)

Script \`client/scripts/check-no-late-route-mount.mjs\` scans \`client/src/\` for \`app.METHOD(...)\` calls outside the startApiServer call graph (allowlist: \`client/src/api/**\`, \`x402/handler.ts\` called from \`server.ts:477\`, \`trajectory/llm-proxy.ts\` standalone instance).

Catches the Hono "matcher is already built" class of bug at code review. Wired as \`yarn lint:no-late-mount\` and a new step in the \`check\` CI job (runs before \`yarn test\`).

Red-green verified: injecting \`app.get(...)\` into main.ts makes the check exit 1 with a pointed error + remediation hint; removing it returns to exit 0.

### 3. Boundary tests for Stage 2 master ETH gate + handbook convention

Two new tests in \`staged-bootstrap-stage1and2.test.ts\`:
- Master at \`gate - 1 wei\` (0.0099 ETH): rejected with funding shortfall = 1 wei.
- Master at \`gate\` exactly (0.010 ETH): accepted, proceeds through mocked Stage 2 steps.

Red-green verified: temporarily flipping the gate's \`STANDARD_MASTER_BOOTSTRAP_MULTIPLIER\` to \`1n\` (so the gate becomes 0.005 ETH instead of 0.010 ETH) makes the "rejects at gate - 1 wei" test fail — exactly the regression shape we're guarding against.

Also adds a sub-bullet to AI workflow rule 7 in \`docs/engineering/handbook.md\`: "Boundary tests for numeric gates." Codifies the convention — every gate change should ship with mocks at \`gate\` and \`gate - 1 wei\`, not at "comfortably above" — so the next time someone touches a threshold, the reviewer asks "where's the boundary test?"

## Why all three on one PR

They're a coherent class of regression coverage. Each independently closes a CI gap that let the u34i regression ship:

| CI gap | What was missing | This PR's guard |
|---|---|---|
| Cross-state UX transitions | No E2E mocked sequential daemon states | Playwright sequential-state test (commit 1) |
| Late route mount | No structural rule on where \`app.METHOD\` can live | Lint script + CI step (commit 2) |
| Numeric gate boundaries | Mocks set comfortably-above the threshold | Boundary tests + handbook convention (commit 3) |

## Follow-up filed

jinn-mono-u34i-class bug audit (Stage 2 in-progress gate vs transfer): the gate is \`minEoaGasEth * 1 = 0.005 ETH\` for in-progress fleets while the per-service transfer is also \`minEoaGasEth = 0.005 ETH\` — same 1:1 ratio that produced u34i. Low practical impact (testnet operators have targetServices=1) but identical bug class.

## Test plan

- [ ] CI: \`check\` job green (now includes \`yarn lint:no-late-mount\` step)
- [ ] CI: \`funding-sequence-e2e\` job green
- [ ] Reviewer: skim the three commits in order; each has a red-green proof in the message
- [ ] Reviewer: agree the handbook convention sub-bullet is the right place to land "boundary tests for numeric gates"


ritsukai (approved):
Regression test for u34i Bug F + CI gate. Red-green verified locally. Approving.

PR #279, opened 2026-05-18
fix(k1ng): setAgentWallet retries on returned ok:false (not just thrown errors)

ritsuKai2000:
## Summary

Fixes the second-time-around bootstrap failure operators hit after #262 landed: \`Fleet setAgentWallet failed: Execution reverted for an unknown reason\` mid-Stage-1.

bd issue: **jinn-mono-rie5**

## What was actually broken

The freshly-deployed-Safe revert race is documented in this file's h74p comment block (\`bootstrap.ts:118-137\`): against a fresh 1/1 Safe on Base Sepolia, the first \`setAgentWallet\` attempt reverts with \"Execution reverted for an unknown reason\"; the same Safe + agentId a few seconds later succeeds.

**The h74p retry was wired to catch *thrown* exceptions only.** But \`bindAgentWalletToSafe\` doesn't throw on revert — it returns \`{ ok: false, error }\`. The retry's \`try/catch\` never fires in production. Dead code.

- **Per-service path** "worked" anyway because the catch-all marks \`safe_binding_pending\` and the next bootstrap restart resumes from there. The race effectively retried on the next \`jinn run\`, not in-process.
- **Stage 1 path** had no such resume safety net — \`ok: false\` threw and halted bootstrap. The 2026-05-18 canary's second-time-around failure mode.

## Process failures behind this slipping through

The PR for #262 ran an end-to-end on Sepolia. It passed. I declared the fix-stack reliable. The race didn't fire that run (one die roll, came up six). The user re-ran the flow themselves and hit the bug immediately. I was wrong on three counts:

1. **N=1 validation against a documented-flaky code path.** A known race needs N>1 OR deterministic mocks of the failure mode. I had neither.
2. **Audited the bug instance, not the bug class.** I only stared at the bind path that bit me under u34i. Should have grepped every \`bindAgentWalletToSafe\` call site (there are two — fleet and per-service) and verified each has *real* retry coverage.
3. **Trusted false-green tests.** The h74p tests mock with \`mockRejectedValueOnce(new Error(...))\` — testing the throw path that doesn't fire in production. The retry "worked" against the test mock; it never actually retried the real \`ok: false\` race. The PR for those tests was approved with the false-green still passing.

## What this PR changes

- **\`FleetBootstrapper.bindAgentWalletWithRetry()\`** — unified helper that retries on returned \`ok: false\` AND on thrown exceptions. Same budget (3 × 3 s) the previous mechanism advertised.
- **Stage 1** (\`stepFleetIdentityRegister\`) now uses the helper. On persistent \`ok: false\`, throws with the actual revert reason surfaced (operator sees \"NotAuthorized\" etc., not the generic halt message).
- **Per-service** (\`resumeServiceStandard\`) also uses the helper. In-process retry actually retries the race now; no longer depends on the \`safe_binding_pending\` resume-on-next-bootstrap fallback.

## Tests (red-green verified)

- \`staged-bootstrap-stage1.test.ts\`: Stage 1 with two transient \`ok: false\` then \`ok: true\` succeeds; persistent \`ok: false\` halts with the revert reason surfaced.
- \`bootstrap.test.ts\`: per-service with two transient \`ok: false\` then \`ok: true\` succeeds. **This is the mock-parity test that should have existed alongside the h74p tests** — uses \`mockResolvedValueOnce({ ok: false, ... })\` matching production behavior.

Existing h74p tests (\`mockRejectedValueOnce\`) still pass — the helper retries both shapes. The throw path is defense-in-depth; the \`ok: false\` path is the actual production failure mode.

Red-green: reverting each path's call to use a single-attempt \`bindAgentWalletToSafe()\` directly makes the matching test fail with \"expected 3 calls, got 1.\" Restored — both pass.

## Verification

- \`npx tsc --noEmit\` → exit 0
- \`yarn lint:no-late-mount\` → green
- \`npx vitest run --exclude '**/daemon.test.ts'\` → **3550 passed, 0 failed** (daemon.test.ts excluded for port-7331 collision, pre-existing)
- bd issues filed:
  - This: jinn-mono-rie5
  - Class follow-up: jinn-mono-mw7y (Stage 2 in-progress master gate has the same gate==transfer ratio as u34i)
  - Audit follow-up: should add a unit-test convention for mock shapes matching production behavior

### DIFF
--- a/client/src/earning/bootstrap.ts
+++ b/client/src/earning/bootstrap.ts
@@ -249,6 +249,76 @@ export class FleetBootstrapper {
   }
 
   /**
+   * Run `bindAgentWalletToSafe` with the freshly-deployed-Safe race retry
+   * (jinn-mono-h74p, corrected in jinn-mono-k1ng).
+   *
+   * The race: against a fresh 1/1 Safe on Base Sepolia, the first
+   * setAgentWallet attempt reverts with "Execution reverted for an unknown
+   * reason"; the same Safe + same agentId a few seconds later succeeds.
+   *
+   * The h74p version of the retry only wrapped the call in a try/catch.
+   * In production, `bindAgentWalletToSafe` doesn't throw on revert — it
+   * returns a `{ ok: false, error }` outcome — so the catch never fired
+   * and the retry was dead code. The per-service path "worked" anyway
+   * because it persists `safe_binding_pending` and the next bootstrap
+   * resumes the bind. Stage 1 had no equivalent safety net, so a single
+   * `ok: false` halted bootstrap (the 2026-05-18 canary failure).
+   *
+   * This wrapper:
+   *   - Retries on returned `ok: false` (the way the race actually
+   *     manifests in production).
+   *   - Also retries on thrown exceptions (defense-in-depth in case viem
+   *     error handling changes).
+   *   - Returns the final outcome; callers decide whether to throw, mark
+   *     pending, or proceed.
+   *
+   * Single attempt budget: `safeBindingMaxAttempts` × `safeBindingRetryDelayMs`
+   * (defaults: 3 × 3 s = ~9 s).
+   */
+  private async bindAgentWalletWithRetry(
+    args: Parameters<typeof bindAgentWalletToSafe>[0],
+    label: string,
+  ): Promise<Awaited<ReturnType<typeof bindAgentWalletToSafe>>> {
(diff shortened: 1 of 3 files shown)


ritsukai (approved):
Cross-account approval for v0.1.6 stewardship: k1ng setAgentWallet retry fix. CI green, MERGEABLE. Approved per /goal authority for self-clear via cross-account flow.

PR #284, opened 2026-05-18
fix: use delivery event time for recovery expiry

ritsuKai2000:
## Summary
- skip recovery delivery replay only after the on-chain recovery deadline has actually expired
- key expiry checks off the delivery event block time instead of the later poll time
- cover the recovery window edge cases in Mech adapter tests

## Validation
- `PATH=$HOME/.nvm/versions/node/v22.22.2/bin:$PATH yarn --cwd client vitest run test/adapters/mech/adapter.test.ts`


ritsukai (approved):
Parity-review by independent agent passed (2026-05-20 review batch). Approving for merge.

Issue #297, opened 2026-05-19
investigate: what's the right shape for a real-Hermes E2E test that exercises the integration boundaries the stub-based test can't?


ritsuKai2000:
## Context

\`yarn e2e:hermes\` (\`client/test/e2e/hermes-agent-full-cycle.ts\`) spawns a stub Hermes binary (\`scripts/stub-hermes.js\`) that writes a fixed solution payload and exits 0. The stub never calls a real LLM, never opens an MCP connection, never exercises the per-task plugin mount or the provider resolution path.

The 2026-05-19 stewardship session surfaced two bugs (#293 provider routing, #294 MCP launcher path) that the stub test couldn't catch by design — both bugs live at integration boundaries the stub is explicitly stubbing out. The per-component unit tests (\`adapter.test.ts\`, \`config-builder.test.ts\`) test pass-through and shape but not behaviour under real execution.

A real-binary E2E would have caught both bugs on first run.

## What to investigate

Don't prescribe a test design yet. Read what's there and answer:

1. **What does the stub test actually cover?** Walk \`client/test/e2e/hermes-agent-full-cycle.ts\` end-to-end. List the integration seams it crosses vs. the ones it stubs. The honest list is short.

2. **What's the cost of a real-binary test?** Pricing model on OpenRouter for the cheapest tool-using model (\`deepseek/deepseek-v4-flash\` at the time of writing, near-free). Per-run cost in tokens × current rates. Worst-case if a buggy daemon retries the task in a loop. CI minutes. Local-dev friction (does every contributor need an OpenRouter key, or is it gated?).

3. **What's the gating model?** Options:
   - Always-on in CI (needs a budget + key management)
   - Gated by env var (\`JINN_E2E_HERMES_REAL=1\` like the existing \`JINN_E2E_HARNESS\` pattern), skipped cleanly otherwise
   - Manual / nightly only
   - Pre-cut gate (only runs on the Monday named cut)
   Each has tradeoffs around cost vs. coverage vs. flakiness.

4. **What infrastructure does it need?** A real Hermes binary (\`~/.hermes/hermes\` or PATH), a real OpenRouter key, a real RPC, a place to fund the test operator on testnet (or an Anvil fork with the staking contracts deployed and the master pre-funded). Map what's already available vs. what needs to be built.

[...]

Issue #296, opened 2026-05-19
investigate: why are SolverPlugins copied per-task into $HOME/.jinn-client/solver-plugins/, and what should the layout be?


ritsuKai2000:
## Context

The daemon copies the SolverPlugin tree (e.g. \`network-tools\`, \`swe-rebench-v2-runtime\`) into \`\$HOME/.jinn-client/solver-plugins/<name>/\` on startup or task setup. The per-task Hermes config references the plugin's MCP server launcher by absolute path under that copy.

The launcher at \`client/plugins/network-tools/mcp/jinn-client-server.mjs\` resolves the daemon's actual server entry by walking \`<pluginRoot>/../..\` and looking for \`dist/mcp/server.js\` or \`src/mcp/server.ts\`. In the original-install layout (\`<client-install>/plugins/network-tools/\`), that walk lands at \`<client-install>/\` and works. In the per-task copy layout (\`\$HOME/.jinn-client/solver-plugins/network-tools/\`), the walk lands at \`\$HOME/.jinn-client/\` and fails.

A v0.1.6 patch (#294) injects \`JINN_NETWORK_TOOLS_CLIENT_ROOT\` into the daemon's env block for the MCP server config, papering over the symptom. The patch is correct as a patch but doesn't address the deeper question of *why* the launcher (which is immutable code, not state) is being copied into the per-task / per-operator tree at all.

## What to investigate

Don't prescribe a solution yet. Read the code and answer:

1. **Where is the per-task copy actually performed?** Trace the daemon startup path until you find the code that copies \`client/plugins/<name>/\` into \`\$HOME/.jinn-client/solver-plugins/<name>/\`. Note the commit + bead that introduced it.

2. **Why was the copy added?** Read the commit message, the bead/spec it references, and any related runbooks. Common candidates:
   - Per-task version pinning (operator wants to run a specific plugin version per task; reconcile against an active version on disk)
   - Mutable plugin state (the launcher or a sibling script wants to write into the plugin tree at runtime)
   - Permissions / sandboxing (operator can't read the client-install location, so we copy to a writable place)
   - npm install resolution (the launcher needs \`node_modules\` co-located with the plugin)
   - Snapshot for the trajectory record (audit who-ran-what-version)
   - Some combination of the above

[...]

Issue #293, opened 2026-05-19
release-blocker: Hermes harness routes to Google for OpenRouter-format model ids


ritsuKai2000:
## Symptom

Op B joined SWE-rebench v2 SolverNet with `harness=hermes-agent` and `model=deepseek/deepseek-v4-flash` (a canonical entry from `HERMES_MODELS` shipped in PR #292). The Hermes child process called `cloudcode-pa.googleapis.com` (Google Code Assist) and died with:

\`\`\`
API call failed after 3 retries: HTTP 404: Code Assist 404: Requested entity was not found.
is not available at cloudcode-pa.googleapis.com.
\`\`\`

Hermes ignored the OpenRouter-shaped model id and routed to its operator-pinned default provider (`google-gemini-cli`), which has no matching model for `deepseek/deepseek-v4-flash`, hence the 404.

## Repro

- Operator on Base Sepolia testnet
- Fresh `HOME=/tmp/op-b`, fresh keystore, fresh `.hermes` home (clone of `~/.hermes`)
- Join the SWE-rebench v2 SolverNet via `/operator/join/<cid>`
- Select harness "Hermes Agent 0.1.0"
- Select model "DeepSeek V4 Flash" (id: `deepseek/deepseek-v4-flash`)
- Save, restart daemon, claim a task
- Inspect `~/.jinn-client/engine/impl-state/hermes-agent/swe-rebench-v2_v1/logs/agent.log`

## Impact

Every Hermes operator picking *any* of the 10 entries in `HERMES_MODELS` (Opus 4.7 via OpenRouter, DeepSeek Flash, Hy3 Preview, Kimi K2.6, Gemini Flash Lite, MiniMax M2.7, Owl Alpha, Hermes 405B, …) hits this. The dashboard's canonical Hermes model list ships in v0.1.6 from PR #292; without a fix to the routing, **every dropdown choice misroutes to Google**. Hermes harness is effectively non-functional in v0.1.6 dashboard installs.

## Where the wiring lives

- `client/src/dashboard/spa/src/pages/configuration/claudeModels.ts` — `HERMES_MODELS` list
- `client/src/harnesses/impls/hermes-agent/adapter.ts:135-140` — model + provider arg construction (`--provider` only passed when `config.hermesProvider` is explicitly set)
- `client/src/main.ts` — daemon constructs `HermesHarnessAdapter`; does not derive provider from model id
[...]

PR #299, opened 2026-05-19
fix(hermes): inject JINN_NETWORK_TOOLS_CLIENT_ROOT into MCP server env (#294)

### LINKED ISSUE #294: release-blocker: jinn-client MCP plugin can't resolve client root in per-task plugin copy layout
## Symptom

When the Hermes harness runs against a real task, the `jinn-client` MCP server fails to start 3× in a row, then gives up. Hermes proceeds through its LLM loop and tool calls but cannot reach the daemon's MCP tools (`submit_typed_payload`, `get_task`), so the task never delivers and no envelope is published on-chain.

Daemon log:

\`\`\`
[harness-engine] tick: process(<request>) failed: hermes-agent: child exited code=1 signal=null:
\`\`\`

MCP-stderr log (after enabling logging):

\`\`\`
Unable to find the Jinn client MCP server. Set JINN_NETWORK_TOOLS_MCP_SERVER_PATH or JINN_NETWORK_TOOLS_CLIENT_ROOT.
Searched: $HOME/.jinn-client/solver-plugins, $HOME/.jinn-client/, $PWD
\`\`\`

## Repro

- Operator on Base Sepolia, daemon spawned from `<client-install>/dist/bin/jinn.js`
- Daemon copies the plugin tree to `$HOME/.jinn-client/solver-plugins/network-tools/` on first run
- Hermes harness configured with `jinn-client` MCP server (via PR #292's bundled `network-tools` plugin)
- Task claimed; Hermes child spawned with mcp_servers config pointing at `$HOME/.jinn-client/solver-plugins/network-tools/mcp/jinn-client-server.mjs`
- Launcher's resolver walks `<pluginRoot>/../..` looking for `dist/mcp/server.js` — lands at `$HOME/.jinn-client/`, which has no `dist/`
- Connection retries 3×, gives up

## Impact

Hermes harness can run the LLM and tool calls correctly but cannot submit a solution back to the daemon. **No envelope CID, no delivery tx, no verdict.** The entire claim → solve → deliver loop is broken for Hermes on any non-package-layout install. The launcher's `../..` walk assumes a layout (`<client-install>/plugins/network-tools/`) that doesn't exist after the daemon's per-task copy.

## Where the wiring lives

- `client/plugins/network-tools/mcp/jinn-client-server.mjs` — the launcher script; resolves daemon server entry via `<pluginRoot>/../..` walk
- `client/src/harnesses/impls/hermes-agent/config-builder.ts:55-130` — builds the per-task Hermes config including `mcp_servers.jinn-client.env` block
- Daemon copies plugin tree to `$HOME/.jinn-client/solver-plugins/` somewhere in the bootstrap / SolverNet-load path (precise location is part of what the proper-fix investigation needs to identify)

The launcher's interface already declares two env vars (`JINN_NETWORK_TOOLS_CLIENT_ROOT` and `JINN_NETWORK_TOOLS_MCP_SERVER_PATH`) that, when set, short-circuit the broken walk. Today nothing sets them.

## Verification path after fix

- Same repro as above; after the patch, MCP server connection should succeed on first attempt
- Daemon log shows `[harness-engine] tick: process(<request>) succeeded` instead of `failed`
- Solution envelope CID lands on-chain via delivery tx
- IPFS gateway returns the envelope JSON when fetched by CID

## Discovered in

v0.1.6 stewardship session 2026-05-19. Full session log: `log/decisions/2026-05-19-v0.1.6-stewardship.md`. Patch PR incoming.

## Follow-up

This is a **patch** for v0.1.6 — the patch injects `JINN_NETWORK_TOOLS_CLIENT_ROOT` into the daemon's env block for the Hermes MCP server config. The **proper fix** (audit *why* plugins are copied per-task at all, and whether the launcher belongs in that copy) is filed separately and slated for investigation in the next release cycle.

ritsuKai2000:
**Closes #294** — v0.1.6 release-blocker patch.
**Tracks proper fix in #296** (per-task plugin layout audit).

## Why

The network-tools plugin's MCP launcher (`client/plugins/network-tools/mcp/jinn-client-server.mjs`) walks `<pluginRoot>/../..` looking for `dist/mcp/server.js` or `src/mcp/server.ts`. When the daemon copies the plugin tree per-task into `\$HOME/.jinn-client/solver-plugins/network-tools/`, the walk lands at `\$HOME/.jinn-client/` — which has no `dist/`. The launcher throws "Unable to find the Jinn client MCP server", Hermes's MCP connection fails 3× and gives up. Result: Hermes can run the LLM and tools but cannot call `submit_typed_payload` — **no envelope, no delivery tx, no verdict**.

Reproduced live during the 2026-05-19 stewardship session — see `log/decisions/2026-05-19-v0.1.6-stewardship.md` and Bug #2 narrative in #294.

## What

`client/src/harnesses/impls/hermes-agent/config-builder.ts`:
- Add `resolveClientRoot()` helper that walks 4 dirs up from this module's `import.meta.url` to reach the client install root (works for both `<client>/dist/...` and `<client>/src/...` layouts).
- Inject `JINN_NETWORK_TOOLS_CLIENT_ROOT: resolveClientRoot()` into `buildJinnRuntimeEnv`'s output. That env block flows to every MCP server config's `env:` block via `translateMcpFromFile` → \`env: { ...jinnEnv, ...(server.env ?? {}) }\` (config-builder.ts:120).
- The launcher already honors `JINN_NETWORK_TOOLS_CLIENT_ROOT` and short-circuits its broken walk when set — we just weren't setting it.

## Patch, not proper fix

This patch tells the launcher where to find the daemon's MCP server. It doesn't address *why* the launcher (which is immutable code, not state) is being copied per-task in the first place. The launcher's own `<pluginRoot>/../..` walk would work correctly if the plugin tree were referenced by absolute path from its install location instead of being copied.

The **proper fix** (#296) is an investigation of the per-task plugin copy: why it exists, what parts of the tree need to be per-task vs. install-located, and what the right plugin-mounting model is. Probably involves separating code (immutable, install-located) from state (per-operator, writable). Filed as an investigation issue, not prescribed; the agent picking it up should read the history first to understand *why* the copy was added.

## Tests

`test/harnesses/impls/hermes-agent/config-builder.test.ts` — 2 new cases:
- ✓ env block includes `JINN_NETWORK_TOOLS_CLIENT_ROOT` pointing at a directory containing `dist/mcp/server.js` OR `src/mcp/server.ts` (asserts the resolver's `../` count is correct — **red-probe verified the test fails without the fix**)
- ✓ spawn the actual launcher with the env block, assert stderr/stdout doesn't contain \"Unable to find the Jinn client MCP server\" within 2s — this is the integration smoke that catches the bug class in unit-test form, and the test the existing surface lacked (see #297)

\`\`\`
 ✓ test/harnesses/impls/hermes-agent/config-builder.test.ts (7 tests) 2108ms
 Test Files  1 passed (1)
      Tests  7 passed (7)
\`\`\`

Typecheck clean (`npx tsc --noEmit` exit 0).

## Verification path post-merge

Operator with Hermes harness claims a real task on Base Sepolia. Daemon log shows `[harness-engine] tick: process(<request>) succeeded` instead of \"failed: hermes-agent: child exited code=1\". Solution envelope CID lands on-chain via delivery tx.

Phase 2 of `log/decisions/2026-05-19-v0.1.6-stewardship.md` will run this verification on Op B once both #293 and #294 are on \`next\`.

## Related

- #294 (this PR closes it)
- #296 (proper-fix investigation tracking)
- #297 (real-Hermes E2E test gap — this PR's launcher-spawn test is a unit-style version of what #297 should ship at integration scale)
- `log/decisions/2026-05-19-v0.1.6-stewardship.md` (Bug #2 discovery context)


ritsukai (approved):
Approving on behalf of Captain Oak per the v0.1.6 stewardship cross-account flow. Patch + tests reviewed; launcher-spawn integration smoke verifies the bug class in unit-test form. Closes release-blocker #294.

Issue #328, opened 2026-05-19
Release feedback — v0.1.6 operator app dogfood (2026-05-19)


ritsuKai2000:
## Context

v0.1.6 published to npm \`@latest\` on 2026-05-19 13:48 UTC (#310 closed, gate relaxed in #313). The first day of public dogfood surfaced operator-app issues that are now the highest-priority work for the rest of Sprint 2.

Two operators ran fresh installs in the dogfood thread:

- **rvx** — new user, first contact with Jinn, on macOS. Joined SWE-rebench v2 via the SPA. Hit harness selection bugs, no success affordance, daemon claiming tasks without a working harness (17/17 then 26/26 failures in minutes), confusion around roles + harness+model subscriptions + plug-ins, gas top-up surprise, daemon-restart not actually restarting.
- **Captain Oak** — running v0.1.6 himself for the first time post-cut. Hit the rate-limited default Tenderly RPC (\"No registry catalog could be found\"), the same Claude Code harness dropdown bug, restart-buttons-don't-restart, status banner inconsistency despite a successful join.

## Scope

This Issue is the **umbrella** for the release-feedback work shaped from that session. Sub-issues are problem-shaped (per AI workflow rule 2) and tracked natively via sub-issue links below.

## Decisions made in-session

- **Hide the embedded agent surface in the operator app** until it's more robust (#326 — CHORE, Captain's read coming out of the test).
- **Deprioritize plug-in builder UI promotion** in the operator app (#327 — \`/build\` out of nav, substrate stays live). Strategic call: solving operator-app first-run UX is more load-bearing than promoting builders right now.

## Sprint 2 fit (Option C)

- **P0 release-feedback bugs ship hotfix-track** — out-of-cadence v0.1.7 patch as soon as ready; don't wait for the Monday cut. P0s here are the dropdown bug, the claim-before-ready bug, and the restart-button bug.
- **P1 release-feedback joins Sprint 2** alongside the waxs chain landing (which still has 6 PRs in review) and the hjex.* PR review-drain.
- **P2 release-feedback + the two CHORE hides defer to Sprint 3 if needed** — but pull them forward if there's headroom this week.

## Path through existing hjex.* PRs

The 9 hjex.* PRs (#239-#247) that sat unreviewed since 2026-05-15 likely fix several of the symptoms in this feedback set (per analysis on #218):

[...]

PR #343, opened 2026-05-19
fix(289): in-process respawn on operator-triggered restart

### LINKED ISSUE #289: 'Restart node' button kills the daemon — operator-never-leaves-the-app violation
## Problem

Clicking **Restart node** in the operator panel (the call-to-action that appears after every restart-required config change — RPC URL, password rotation, SolverNet config, etc.) does exactly one thing:

\`\`\`ts
// client/src/main.ts:935-940
admin: {
  onRestartRequested: () => {
    console.log('[main] Restart requested via operator MCP. Exiting...');
    process.exit(0);
  },
},
\`\`\`

`process.exit(0)`. The daemon dies, the terminal returns to a shell prompt, the panel goes 502-on-localhost. The operator is now stranded: their only path back is to switch to a terminal, find the right `jinn run` invocation, and re-launch.

This is the most operator-visible breach of the **operator-never-leaves-the-app principle**. Every config change that the panel marks as `restartRequired: true` funnels through here:

- `POST /v1/setup/network` — RPC URL change
- `POST /v1/setup/change-password` — keystore password rotation
- `POST /v1/setup/solvernets/:name` — SolverNet enable/disable
- `POST /v1/operator/join/:cid` — SolverNet join (`joinedSolverNets[<cid>]`)

In each case the panel says "Restart the node to apply" with a button — and the button hands the operator a dead panel + a terminal prompt.

## Proposal: in-process respawn

The smallest fix that makes "Restart node" actually mean what it says. Before `process.exit(0)`, spawn a detached copy of the current process inheriting the same env, then exit:

\`\`\`ts
// client/src/main.ts
import { spawn } from 'node:child_process';

admin: {
  onRestartRequested: () => {
    console.log('[main] Restart requested. Respawning daemon...');
    const child = spawn(process.execPath, process.argv.slice(1), {
      detached: true,
      stdio: 'inherit',
      env: process.env,
    });
    child.unref();
    // Give the child a moment to bind before vacating the port.
    setTimeout(() => process.exit(0), 250);
  },
},
\`\`\`

The child inherits `HOME`, `JINN_PASSWORD` (if set), `--no-ui`, all flags. The panel reconnects to the same port within ~2 seconds of the restart click. Operator stays in the app the whole time.

### Caveats / tradeoffs

- **Port-7331 race.** The child's `app.listen(7331)` races the parent's `process.exit(0)`. The 250ms gap is empirically enough on macOS; Linux may want a longer delay or SIGTERM + wait-for-port-free before exit. Worth measuring.
- **Stdio handover.** With `stdio: 'inherit'`, the child writes to the same terminal as the parent. Fine when the operator launched `jinn run` interactively; cleaner with `stdio: 'ignore'` + a log file when run as a service. Picking `'inherit'` for now matches current behavior.
- **Test coverage.** Unit-testing process respawn is awkward — easier to validate via E2E: spawn daemon, hit `/api/admin/restart`, poll for port 7331 to come back, confirm new PID. Add to the existing dashboard E2E suite (`client/test/dashboard/`).
- **Packaged-app future.** Once the daemon ships inside a Mac/Windows wrapper binary, the wrapper handles respawn and this in-process logic becomes dead. Easy to gate behind an env var (\`JINN_INPROCESS_RESPAWN=1\`) if needed; for now the in-process path is the only thing that makes Restart node real.

## What this PR does NOT address

This is **restart-via-respawn**, not **live config hot-reload**. The RPC URL specifically is held by reference across many subsystems (FleetBootstrapper.publicClient, daemon loops, harness adapters, cross-chain L1 client). Hot-swapping it without restart requires touching ~30-50 call sites to go through a getter/accessor pattern. Out of scope here.

A handful of config keys already hot-apply today via in-memory mutation (`onSolverNetsUpdated` callback, SolverNet generator config). Extending the hot-apply set is its own work; respawn is the universal escape hatch.

## Severity

P1 operator-onboarding friction. Every operator who finishes bootstrap and changes their RPC, password, or SolverNet config hits this exactly once and then has to learn that "Restart node" really means "go back to terminal." Compounds with #288 (misleading runway display) and the SolverNet manifest 404s on rate-limited public RPC — operators are most likely to change RPC right after bootstrap, which is the worst time to surface a dead panel.

## Related

- Operator-never-leaves-the-app principle (PRINCIPLES.md / CLAUDE.md). Reiterated by Oak verbally several times during 2026-05-18 canary thread.
- #262 / jinn-mono-u34i — the funding fix-stack that surfaced this gap (operators now successfully reach the running dashboard, immediately hit a config change, immediately hit this).
- jinn-mono-xhir — operator-never-terminal architectural cleanup for the bootstrap-halt path. Same principle, different surface (halt path vs restart path).
- #280 — L2 vestigial fleet Safe collapse spec; reduces bootstrap costs but doesn't address this.
- #288 — misleading "0 days runway" display calibration.
- Future: hot-reload RPC without restart (~30-50 call sites). Big enough that it should be its own spec issue.

ritsuKai2000:
Closes #289.

## What changed

Replaced the bare `process.exit(0)` in the restart handler with a spawn-then-exit pattern: spawn a detached copy of the current process inheriting env/argv/stdio, then exit. The new child takes over the port; the operator sees the dashboard reconnect rather than a dead panel + terminal prompt.

Specifically:

- New helper `client/src/restart-daemon.ts` exporting `requestDaemonRestart()` — does the spawn-then-exit, with a `JINN_NO_UI=1` headless gate that skips respawn (lets the supervisor decide).
- `client/src/main.ts` `onRestartRequested` now calls `requestDaemonRestart()` instead of `process.exit(0)`.

## Why

Per #289: clicking "Restart node" in the operator app was supposed to actually restart. With the old pattern, it killed the daemon and stranded the operator. Every restart-required config change funnelled through this — RPC URL, password rotation, SolverNet enable/disable, SolverNet join. The operator-never-leaves-the-app principle (PRINCIPLES.md / CLAUDE.md) was breached on the most-clicked restart surface in the panel.

## Test plan

Regression tests at `client/test/main/restart-daemon-respawn.test.ts` (8 tests, all pass):

- [x] `isHeadless` predicate true on `JINN_NO_UI=1`, false otherwise
- [x] Interactive mode: spawn invoked once with `execPath` + `argv.slice(1)`, options include `detached: true`, `stdio: 'inherit'`, `env` passed through, child `unref()`'d, parent exits 0
- [x] Interactive mode: log line announces respawn
- [x] Headless mode: spawn NOT invoked, parent exits 0, log line mentions `JINN_NO_UI`
- [x] main.ts wiring: `onRestartRequested` no longer contains bare `process.exit(0)`; routes through `requestDaemonRestart`

Full client vitest suite green: **3574 passed / 7 skipped / 0 failed** (24.8s).

Typecheck (`tsc --noEmit`): zero errors.

Manual smoke (recommended pre-merge):
- [ ] Click a restart-required UI action in the dashboard SPA, observe daemon respawn and panel reconnect.
- [ ] Run `JINN_NO_UI=1 jinn run`, hit `POST /api/admin/restart` with the admin token, confirm daemon exits without respawning.

## Headless gate

`JINN_NO_UI=1` is the existing headless flag in `main.ts` (line 721, `keepSetupUiOnBootstrapError`). Reusing it keeps the env surface flat — no new flag invented.

## Out of scope

- Hot-reload of RPC URL (~30-50 call sites; needs its own spec).
- Packaged-app wrapper handling respawn (future work; helper is easy to gate behind `JINN_INPROCESS_RESPAWN=0` if needed).
- Port-7331 race tuning beyond the 250ms default (issue body flagged Linux may want longer; can iterate based on field reports).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/client/src/main.ts
+++ b/client/src/main.ts
@@ -38,6 +38,7 @@ import { readModeState } from './harnesses/mode-state.js';
 import { attachAgentWs, updateAgentClaudePath } from './agent/agent-ws.js';
 import { createSetupModeController } from './setup-mode.js';
 import { formatBootstrapOperatorMessage } from './operator-errors.js';
+import { requestDaemonRestart } from './restart-daemon.js';
 import { buildEnvelope, emitEnvelope, type ErrorCode, type ErrorEnvelope } from './errors/envelope.js';
 import {
   clearBootstrapError,
@@ -944,10 +945,12 @@ export async function main(): Promise<DaemonStartupInfo | SetupHaltedInfo | void
         hermesDoctorTimeoutMs: config.hermesDoctorTimeoutMs,
       },
       admin: {
-        onRestartRequested: () => {
-          console.log('[main] Restart requested via operator MCP. Exiting...');
-          process.exit(0);
-        },
+        // jinn-mono #289: in interactive mode (the dashboard SPA case),
+        // spawn a detached replacement before exiting so the panel reconnects
+        // to a live daemon instead of seeing a 502 + terminal prompt. In
+        // headless mode (`JINN_NO_UI=1`), exit without respawning so the
+        // supervisor / systemd / docker entrypoint decides what to do.
+        onRestartRequested: () => requestDaemonRestart(),
       },
       harnessStatus: {
         getStatus: async () => {
--- a/client/src/restart-daemon.ts
+++ b/client/src/restart-daemon.ts
@@ -1,0 +1,120 @@
+/**
+ * In-process respawn helper for operator-triggered daemon restarts.
+ *
+ * Issue #289: clicking "Restart node" in the operator panel previously called
+ * `process.exit(0)`, which killed the daemon and stranded the operator outside
+ * the app (the panel went 502, the terminal got a shell prompt). Every
+ * restart-required config change funnelled through the same handler:
+ *
+ *   - POST /v1/setup/network              — RPC URL change
+ *   - POST /v1/setup/change-password      — keystore password rotation
+ *   - POST /v1/setup/solvernets/:name     — SolverNet enable/disable
+ *   - POST /v1/operator/join/:cid         — SolverNet join
+ *
+ * The fix is the smallest thing that makes the button mean what it says:
+ * before the parent exits, spawn a detached copy of the current process,
+ * inheriting argv (sans node binary), env, and stdio. The child takes over
+ * the port; the panel reconnects within a couple of seconds.
+ *
+ * Headless gate: `JINN_NO_UI=1` (already the established headless flag in
+ * main.ts) skips the respawn — operators running `jinn run --no-ui` from a
+ * supervisor / systemd unit / docker entrypoint want the supervisor to
+ * decide whether to restart, not the daemon.
+ *
+ * The helper is split out from main.ts so it's unit-testable without
+ * touching the entry-point bootstrap path.
+ */
+
+import { spawn, type SpawnOptions } from 'node:child_process';
+
+/**
+ * Options for `requestDaemonRestart`. All optional in production; the helper
+ * defaults to `process` / `node:child_process`. Tests inject doubles.
+ */
+export interface RequestDaemonRestartOptions {
+  /** Defaults to `process.env`. */
+  env?: NodeJS.ProcessEnv;
+  /** Defaults to `process.argv` (`[node, scriptPath, ...args]`). */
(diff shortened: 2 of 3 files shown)


ritsukai (approved):
Parity-review by independent agent passed (2026-05-19). See review summary on #328. Approving for merge.

PR #353, opened 2026-05-20
fix(338): clarify Leave SolverNet copy — explain immediate vs deferred effects

### LINKED ISSUE #338: fix: Leave SolverNet copy is ambiguous — 'stops claims on next restart' doesn't tell operator what happens now
## Run-mode

BUG-FIX (P2) — copy fix.

## Context

rvx in the dogfood: "Ok so I'm leaving the Solvernet — what does it mean 'stops claims'? and 'on the next restart' — does this mean it'll keep going until I restart?"

The Leave SolverNet affordance's copy doesn't tell the operator what happens *now* vs *on restart*. Worse, rvx then observed: "I've left the Solvernet but it's still solving (Task 26 Failed 26)" — so the *behaviour* matches the copy (it keeps claiming until restart), but the copy doesn't explain that this is intentional or what the workaround is.

## Impact

P2 copy + small behavioural Q. Confusing for first-run operators; minor.

## Acceptance criteria

- Leave SolverNet modal / affordance explains in plain language:
  - What happens immediately on Leave (the SolverNet config marker is removed; in-flight tasks complete; subsequent claims stop *when the claim loop next checks config*).
  - When the operator should restart (or whether restart is required).
  - The "in-flight tasks" caveat — Leave doesn't kill running tasks.
- Optional: if the claim loop can re-read joinedSolverNets without restart, drop the restart-required path entirely.

## Reference

- Parent: #328 (release feedback).
- Adjacent: #289 (restart button respawn — once that lands, restart-required UX everywhere simplifies).

## Discovered in

2026-05-19 v0.1.6 dogfood — rvx.

ritsuKai2000:
## Summary

The Leave SolverNet affordance's copy was ambiguous. It said only:

> Leaving stops claims for this SolverNet on the next daemon restart.

A v0.1.6 dogfood operator (rvx) read this as: "does it stop *now* or keep going until I restart?" — and then observed the daemon was *still solving* after they left. The behaviour is correct (`joinedSolverNets` is loaded into the TaskEngine at startup and not hot-reloaded — confirmed by `joinedSolverNetsViewFromConfig` wiring in `main.ts` and the `restartRequired: true` leave-endpoint response), but the copy never explained it.

This PR rewrites both Leave explainers (`JoinedNetCard.tsx`) to state the three facts plainly:

- The config entry is **removed immediately** on Leave.
- The **running daemon keeps claiming and finishing tasks** for this SolverNet until restarted — restart is what actually stops new claims.
- **In-flight tasks always run to completion** — Leave does not kill running tasks.

Drops the metaphor on a behaviour-with-consequences surface, per `BRAND.md` (plain speech where operator effort/cost is on the line).

The retired-banner copy gets a matching one-line clarification about the post-leave restart step.

### Restart-required path: kept, not dropped

Issue #338 floated dropping the restart-required path *if* the claim loop can re-read `joinedSolverNets` without restart. It cannot — CLAUDE.md states "the daemon does not hot-reload SolverNet config" and the engine reads the config snapshot once at startup. Making the daemon hot-reload would be a daemon-side refactor (out of scope for a copy fix). The fix instead makes the restart requirement explicit and unambiguous. #289 (restart-button respawn) is the right follow-up to simplify this UX once it lands.

## Test plan

- [x] `cd client && tsc --noEmit` — zero errors
- [x] `vitest run` — full suite: 3699 passed, 7 skipped, 0 failed
- [x] Two new component tests on `JoinedNetCard.test.tsx` assert the active-card explainer covers (a) immediate config removal, (b) until-restart behaviour, (c) in-flight caveat, and the retired banner covers the post-leave restart step

Closes #338

### DIFF
--- a/client/src/dashboard/spa/src/pages/configuration/JoinedNetCard.test.tsx
+++ b/client/src/dashboard/spa/src/pages/configuration/JoinedNetCard.test.tsx
@@ -342,6 +342,36 @@ describe('<JoinedNetCard />', () => {
     expect(screen.queryByTestId('joined-net-card-orphaned')).toBeNull();
   });
 
+  // Regression for #338: rvx in the v0.1.6 dogfood asked "what does 'stops
+  // claims' mean? and 'on the next restart' — does this mean it'll keep going
+  // until I restart?" The pre-fix copy said only "Leaving stops claims for
+  // this SolverNet on the next daemon restart." which was ambiguous on three
+  // axes: (a) what happens right now to the config, (b) what the running
+  // daemon does in the meantime, (c) what happens to in-flight tasks. The
+  // copy must now say all three explicitly.
+  it('Leave explainer covers immediate config removal, the until-restart behaviour, and the in-flight caveat (#338)', () => {
+    wrap(<JoinedNetCard joined={ENTRY} catalogEntry={CATALOG} defaultExpanded />);
+    const explainer = screen.getByTestId('joined-net-card-leave-explainer');
+    const text = explainer.textContent ?? '';
+    // (a) immediate effect on config.
+    expect(text).toMatch(/removes this SolverNet from your config right away/i);
+    // (b) running daemon keeps going until restart.
+    expect(text).toMatch(/keeps? .*loaded copy in memory/i);
+    expect(text).toMatch(/until you restart/i);
+    // (c) in-flight task caveat.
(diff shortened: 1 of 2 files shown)


ritsukai (approved):
Canonical code review (superpowers code-reviewer.md template, independent agent) — verdict: Ready to merge: Yes. Clean.

PR #355, opened 2026-05-20
fix(290): wire PluginPublication indexing for plugin:* MetadataSet events

### LINKED ISSUE #290: Ponder indexer is missing pluginPublication indexing — operator app /build panels are always empty
## Problem

The operator app's `/build` page has two registry panels:

- **Published plug-ins for `<solverType>`** — community plug-ins published against a SolverType, browsed by anyone considering joining.
- **Your published plug-ins** — the operator's own publications under their builder agentId.

Both panels are permanently empty even when plug-ins have been published to the on-chain `IdentityRegistry` (as `MetadataSet` events with key `plugin:<cid>`). The daemon's `/v1/discovery/plugin-publications` and `/v1/discovery/builder-artifacts` endpoints return `{publications: []}` / `{artifacts: []}` because the upstream data source — the Ponder indexer — doesn't index those events.

## Reproduction

```bash
# Introspect the deployed indexer's schema
curl -sS -X POST https://jinn-indexer-production.up.railway.app/graphql \
  -H 'Content-Type: application/json' \
  -d '{"query":"{ __schema { queryType { fields { name } } } }"}' \
  | jq -r '.data.__schema.queryType.fields[].name'
```

Output (truncated; full set of fields the schema exposes):

```
task
tasks
attempt
attempts
verdict
verdicts
rewardDistribution
rewardDistributions
solverNetManifest
solverNetManifests
envelope
envelopes
harnessCheckpoint
harnessCheckpoints
attemptEnvelopeMeta
attemptEnvelopeMetas
verdictEnvelopeMeta
verdictEnvelopeMetas
_meta
```

No `pluginPublication` / `pluginPublications` field. The daemon's `HttpDiscoveryAPI.listPluginPublications` (in `client/src/discovery/http.ts:688+`) issues `query ListPluginPublications($where: pluginPublicationFilter, $limit: Int!) { pluginPublications(...) { items { ... } } }` against this schema and gets:

```json
{"errors":[{"message":"Cannot query field \"pluginPublications\" on type \"Query\".","extensions":{"code":"GRAPHQL_VALIDATION_FAILED"}}]}
```

The Http discovery layer surfaces this as a `DiscoveryUnavailableError`, falls back to the onchain floor, which has a stub that returns `[]` (see related issue: onchain-floor needs real `listPluginPublications`). Net result: the panel renders the empty state forever.

## What the indexer needs

Index the `IdentityRegistry.MetadataSet(uint256 agentId, string key, string value)` events where `key` matches `plugin:*`. Each such event records a plug-in publication (with `cid` = `value.split(':')[1]` or whatever encoding is canonical — confirm against the publishing code at `client/src/cli/commands/solver-plugins.ts` or wherever publications write).

Minimum fields the GraphQL `pluginPublication` row should expose (mapping to the `PluginPublicationRow` interface that `client/src/discovery/http.ts` already expects):

- `cid` — IPFS CID of the plug-in tarball
- `name` — npm package name (from the manifest fetched from IPFS)
- `version` — semver (from the manifest)
- `agentId` — builder's ERC-8004 agentId (the indexed event arg)
- `solverType` — from `jinn.supports[]` in the manifest; one row per `(cid, solverType)` so the panel's `solverType` filter works
- `supportsRuntime` — true if `jinn.supports` includes `"jinn.runtime"`
- `publishedAt` — block timestamp
- `revoked` — boolean, true if a subsequent `MetadataSet` revoked this CID
- `revokedAt` — block timestamp of revocation (nullable)

The `pluginPublicationFilter` GraphQL input should support `solverType`, `agentId` (= builderAgentId), and `revoked` filters; the daemon's `listPluginPublications({ solverType, builderAgentId, includeRevoked })` constructs the corresponding `where` clause.

Daemon-side schema reference: see the `pluginPublication` consumption path starting at `client/src/discovery/http.ts:688`. The query and row shape there are the contract the indexer must satisfy.

## Constraint

Plug-in publication payloads (the manifest contents) live on IPFS, not on-chain. The on-chain event only records `(agentId, key="plugin:<cid>", value="<canonical-uri>" or similar)`. The indexer can either:

1. **Index only the on-chain event** — exposes `(cid, agentId, publishedAt, revoked, revokedAt)`. Pluses: lightweight. Minuses: `name`, `version`, `solverType` aren't in the event; the daemon would need a second hop (IPFS fetch) per row to populate those.
2. **Index + fetch** — the indexer fetches the manifest from IPFS at index time, parses it, and stores `name`, `version`, `solverType[]` alongside the event. Pluses: panel renders rich rows in one query. Minuses: indexer takes a dependency on IPFS uptime + gateway choice.

Recommend (2) — the panel is currently structured to display rich rows (see `client/src/dashboard/spa/src/pages/build/PublishedPluginsPanel.tsx`), and IPFS fetches at index time amortize well vs. per-request fetches from each operator's daemon. Indexer's IPFS gateway choice can mirror the daemon's `JINN_IPFS_GATEWAY_URL` default.

## Scope

- New `pluginPublication` table in the Ponder schema (`client/plugins/jinn-mono-attd-ponder/...` or wherever the indexer lives — I see references but haven't traced the repo path).
- Event handler for `IdentityRegistry.MetadataSet` filtering on `key.startsWith('plugin:')`.
- IPFS fetch + manifest parse during indexing (depends on the indexer's runtime; Ponder supports `context.fetch` in handlers).
- GraphQL query / filter wiring to match the daemon's existing query shape.
- Redeploy to `jinn-indexer-production.up.railway.app`.

## Operator impact today

- `/build` page registry panels render permanent empty state.
- Operators considering joining a SolverNet have no way to see the available plug-in catalog from the panel.
- Builders publishing their first plug-in cannot confirm via panel that it landed; they must `cast call` the IdentityRegistry directly.

## Related

- jinn-mono-attd — the original spec that introduced `listPluginPublications` (the daemon-side implementation references this).
- #289 — operator-never-leaves-the-app: dead `/build` panels compound this principle violation.
- Companion issue: onchain floor `listPluginPublications` is also a stub returning `[]`; even if the indexer is down, the daemon should be able to read MetadataSet events directly via `eth_getLogs`.

ritsuKai2000:
## Summary

- The Ponder indexer's `pluginPublication` entity (schema, `handleMetadataSet` plugin path, REST routes) already shipped in **PR #214**. The remaining gap was the **on-chain RPC floor** — the always-live fallback the daemon routes to during an indexer outage — where `listPluginPublications` / `listBuilderArtifacts` were stubs returning `[]`. With the floor blind to plug-in publications, the operator app's `/build` registry panels render a permanent empty state whenever the indexer is unreachable.
- This implements the floor: scan `IdentityRegistry` `MetadataSet` events for `plugin:<cid>` keys, decode the on-chain `PLUGIN_PAYLOAD_TUPLE` / `REVOCATION_PAYLOAD_TUPLE`, and fold most-recent-wins by `(blockNumber, transactionIndex, logIndex)` — mirroring the indexer's `handleMetadataSet` plugin path and the existing `solvernet-manifest:` floor scan. All rich fields (`name`, `version`, `supports`, `pluginSha256`, `publishedAt`) are in the on-chain payload, so the floor serves complete rows **without an IPFS hop**.
- `listBuilderArtifacts` now delegates to `listPluginPublications`, matching the HTTP layer. `getPluginScores` stays a stub — score history needs the indexer's `attemptEnvelopeMeta` + `verdict` enrichment join, which the floor cannot reconstruct.

## Why this, not an indexer change

Investigation found the indexer code is **already complete** — `packages/indexer/ponder.schema.ts` exports `pluginPublication`, `src/handlers.ts` routes `plugin:<cid>` keys, `src/index.ts` registers the handler, and `src/api/index.ts` exposes REST routes. All 186 indexer tests pass. The issue's reproduction (deployed GraphQL missing `pluginPublications`) is an **operational redeploy** of the Railway instance with a `DATABASE_SCHEMA` bump (see `packages/indexer/deploy/README.md` §"Zero-downtime rolling deploys") — not a code change in this repo.

The actionable code gap was the on-chain floor, which the Issue explicitly calls out as a "Companion issue" and §Constraint Option 1. Fixing it means the `/build` panels populate **even when the indexer is down**.

## Test plan

- [x] `client` typecheck (`tsc --noEmit`) — clean
- [x] `test/discovery/onchain.test.ts` — 34 tests pass, including 10 new `listPluginPublications` cases (v1 decode, v2 revocation, un-revoke on republish, `includeRevoked`/`solverType`/`builderAgentId` filters, garbage-payload resilience, `DiscoveryUnavailableError` on RPC failure, `listBuilderArtifacts` delegation)
- [x] `test/discovery/` — all 94 tests pass
- [x] `packages/indexer` — all 186 tests pass (indexer side already complete)
- [ ] Operational follow-up (not in this PR): redeploy `jinn-indexer-production.up.railway.app` with a `DATABASE_SCHEMA` bump so the indexer's `pluginPublications` GraphQL field goes live.

Closes #290

Generated with Claude Code

### DIFF
--- a/client/src/discovery/onchain.ts
+++ b/client/src/discovery/onchain.ts
@@ -16,6 +16,11 @@
  *   from IdentityRegistry (key prefix `solvernet-manifest:`) and fold via the
  *   existing `resolveMostRecentWins` helper.
  *
+ * - `listPluginPublications` / `listBuilderArtifacts` read `MetadataSet` events
+ *   (key prefix `plugin:`), decode the on-chain PLUGIN_PAYLOAD_TUPLE /
+ *   REVOCATION_PAYLOAD_TUPLE, and fold most-recent-wins so the operator app's
+ *   /build registry panels keep rendering during an indexer outage (gh#290).
+ *
  * - `queryEnvelopes` delegates to the existing `runOnchainCorpusQuery`.
  *
  * - `cursorCache` is an optional injection point: when provided, the start
@@ -27,6 +32,7 @@
 
 import {
   createPublicClient,
+  decodeAbiParameters,
   decodeEventLog,
   http,
   type Address,
@@ -42,6 +48,8 @@ import { JINN_ROUTER_ABI } from '../adapters/mech/types.js';
 import { canClaimTask } from '../adapters/mech/contracts.js';
 import { manifestDigestForCid } from '../adapters/mech/digest.js';
 import { resolveMostRecentWins, type SetMetadataEvent, type SetMetadataLifecyclePayload } from '../solvernets/most-recent-wins.js';
+import { PLUGIN_PAYLOAD_TUPLE, REVOCATION_PAYLOAD_TUPLE } from '../erc8004/abis.js';
+import { PLUGIN_METADATA_KEY_PREFIX } from '../erc8004/plugin-registry.js';
 
 // ── Constants ────────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


ritsukai (approved):
Canonical review found 1 Important issue (lossy plugins cursor); fixed in commit f9d154eb and verified by direct diff inspection. Approving.

PR #356, opened 2026-05-20
fix(308): deep-fetch tag ancestry in promote-main workflow

### LINKED ISSUE #308: investigate: Promote main on release fails ancestry check after shallow tag fetch
## Context

`.github/workflows/promote-main.yml` failed on v0.1.6's release event with `main is not an ancestor of release SHA 92ba53615249d7f0447fafb2eab7be7224e90f16`, even though `git merge-base --is-ancestor origin/main 92ba5361` returns success locally (main IS an ancestor — 345 commits behind, with merge-base = main itself at `099d84cf`).

Run: https://github.com/Jinn-Network/mono/actions/runs/26094874295

The failure is in the "Verify fast-forward only" step at line 75 of the workflow:

```bash
if ! git merge-base --is-ancestor HEAD "$RELEASE_SHA"; then
  echo "::error::main is not an ancestor of release SHA $RELEASE_SHA..."
  exit 1
fi
```

Net effect: `main` did not fast-forward to the release SHA after v0.1.6 published, and downstream tooling (and humans) that read `main` see a stale snapshot.

## Suspected root cause — shallow boundary at the tag commit

The workflow does three fetches in sequence:

1. `actions/checkout@v5` with `ref: main` and `fetch-depth: 0` — fetches full history of `main`. After this, `099d84cf` and all its ancestors are present.
2. `git fetch --tags --depth=1 origin "$RELEASE_TAG"` (line 37) — fetches the tag's commit, but `--depth=1` creates a shallow graft at `92ba5361`: git knows the commit object but its parents are unknown locally.
3. `git fetch origin next` (line 59) — fetches `next` (which contains `92ba5361` and its ancestry to main), but **`git fetch` does not auto-unshallow existing shallow grafts**; the boundary at `92ba5361` from step 2 persists.

When `git merge-base --is-ancestor HEAD "$RELEASE_SHA"` walks from `92ba5361` looking for a path to `099d84cf` (HEAD = main), it hits the shallow boundary and concludes ancestry cannot be established.

This matches the symptom exactly: locally with a complete clone the check passes; on the workflow runner with the shallow tag fetch it fails.

## What to investigate (don't prescribe yet)

1. **Confirm the shallow-graft hypothesis.** Re-run the workflow with `set -x` around line 37 and dump `cat .git/shallow` after each fetch. The expected output is `92ba53615249d7f0447fafb2eab7be7224e90f16` after step 2, and that file should still contain it after step 3.

2. **Choose the smallest fix that preserves intent.** Options:
   - **(a)** Remove `--depth=1` from line 37 — fetch the tag deeply. Tag fetches are cheap (no large objects) but pull the full ancestry to a common base.
   - **(b)** Replace the `--depth=1` fetch with `git fetch --tags origin "$RELEASE_TAG":refs/tags/"$RELEASE_TAG"` and rely on the fact that `next` (fetched deeply at line 59) contains the tag commit's full ancestry — but only if line 59 runs BEFORE the shallow boundary matters. (It does not in the current order; the shallow fetch happens first.)
   - **(c)** Reorder: fetch `next` (deeply) first, then resolve the tag's SHA via `git rev-parse origin/next^{commit}` or via `git ls-remote --tags`. Avoid the shallow tag fetch altogether.
   - **(d)** Add `git fetch --unshallow origin` after the tag fetch to repair the boundary.

3. **Is the line-59 deep fetch actually deep?** `git fetch origin next` without `--depth` fetches as deep as the existing local history allows but does not auto-unshallow grafts. Verify that after this fetch, `git rev-parse 92ba5361~10` actually resolves (proving ancestry is walkable).

4. **Does the `Verify release SHA is reachable from next` check on line 60 succeed in the failing run?** Per the failure log it does — only the line-75 check fails. That's consistent with the shallow-graft hypothesis: the line-60 walk goes from `$RELEASE_SHA` toward `origin/next` (very short — usually 0 or 1 hop), so it doesn't hit the boundary. The line-75 walk goes from `$RELEASE_SHA` toward `main` (345 hops for v0.1.6), which crosses the boundary.

## Hotfix preservation

The workflow correctly handles the hotfix case via the early `Skip if main already at release SHA` step (line 41). The fix to the ancestry check must not regress that path. Any solution should be a no-op when `MAIN_SHA == RELEASE_SHA`.

## Manual recovery for this cycle

`main` remains at `099d84cf` pending Captain decision. A manual fast-forward (`git push origin 92ba5361:main`) is the immediate path, gated on Captain authorisation. The fix to the workflow lands separately.

## Discovered in

v0.1.6 release recovery (2026-05-19). First named cut under the qlol two-train flow where this workflow ran end-to-end. The bug likely has been latent on v0.1.4 / v0.1.5 too — check the prior runs for similar failures (`gh run list --workflow="Promote main on release"`).

## Surface area to read first

- `.github/workflows/promote-main.yml` — the failing workflow (full text, ~93 lines)
- `docs/engineering/handbook.md §Cadence` — the two-train cadence design
- `bd memories two-train-release-flow-migration-jinn-mono-qlol` — the migration's captured intent
- #252 — the PR that shipped this workflow
- #307 — the holistic-review-at-main investigation (related: the gate this workflow auto-FFs is the same gate #307 asks whether to restore)

## Don't

- Don't disable the ancestry check. It exists for a reason — to refuse a non-fast-forward that would rewrite `main`'s history.
- Don't switch to a force-push. The check is correct; the fetch shape upstream of it is what's wrong.

ritsuKai2000:
## Summary
- `promote-main.yml`'s release-event ancestry check failed v0.1.6's promotion ("main is not an ancestor of release SHA 92ba5361...") even though main IS an ancestor.
- Root cause: the tag fetch ran with `--depth=1`, which writes the tag commit's SHA into `.git/shallow`, creating a shallow graft. The subsequent `git fetch origin next` does NOT auto-unshallow existing grafts, so the `git merge-base --is-ancestor main <tag>` check failed walking across the shallow boundary.
- Fix: drop `--depth=1` from the tag fetch so no shallow graft is created. Tag fetches are cheap (no large objects); the deep fetch pulls only the ancestry needed between the tag commit and the existing local history rooted at main (already fetched with `fetch-depth: 0` by `actions/checkout@v5`).

## Test plan
- [ ] Confirm the next Monday named cut triggers `promote-main.yml` and the ancestry check passes
- [ ] Verify the hotfix early-exit path (`Skip if main already at release SHA`) still short-circuits when `MAIN_SHA == RELEASE_SHA`

Closes #308

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/promote-main.yml
+++ b/.github/workflows/promote-main.yml
@@ -34,7 +34,14 @@ jobs:
         env:
           RELEASE_TAG: ${{ github.event.release.tag_name }}
         run: |
-          git fetch --tags --depth=1 origin "$RELEASE_TAG"
+          # Fetch the tag deeply (no --depth=1). A shallow tag fetch creates a
+          # shallow graft at the tag commit; subsequent `git fetch origin next`
+          # does NOT auto-unshallow that graft, so the line-75 ancestry check
+          # (`git merge-base --is-ancestor main <tag>`) fails when walking
+          # across the boundary. Tag fetches are cheap (no large objects); the
+          # deep fetch pulls only the ancestry needed to resolve ancestry from
+          # main to the tag commit. See #308.
+          git fetch --tags origin "$RELEASE_TAG"
           SHA="$(git rev-parse "$RELEASE_TAG^{commit}")"
           echo "release_sha=$SHA" >> "$GITHUB_OUTPUT"
 


ritsukai (approved):
Canonical code review — verdict: Yes. Minor polish only.

PR #377, opened 2026-05-20
fix(engine): clean up engine/work task dirs after terminal state (#320)

### LINKED ISSUE #320: engine/work/ task working dirs accumulating — 412 dirs / 49GB on operator at ~/.jinn-client (investigate cleanup gap)
## What

Local operator at \`~/.jinn-client/\` accumulated **412 per-task working directories** under \`engine/work/\`, totaling **49GB**, with no apparent cleanup. Disk on the host is now at 96% capacity (21GB free on 460GB), and the bulk is attributable to this single subtree.

Each \`engine/work/<task-hash>/\` directory contains:

\`\`\`
0x<task-hash>/
├── env/                       # env state for the task
├── plugins/                   # per-task plugin copy (relates to #294)
├── repo/                      # cloned git repo the task operates on
├── sessions/                  # session transcripts
├── system_snapshot.tar.gz     # snapshot for restoration verdict
└── task.json                  # task metadata
\`\`\`

These are per-task scratch the daemon creates when claiming/solving a task. \`engine/work-archives/\` exists (86MB, presumably the compressed-rotation destination), suggesting cleanup is *intended* but not consistently running.

## Why investigate first

Don't have a proposed fix yet. Possible directions to rule out before proposing:

- Is cleanup logic actually running and getting blocked (e.g. open file handles, partial deliveries)?
- Are some of these intentionally retained (recovery in-flight, audit window for completed deliveries)?
- Is \`work-archives\` rotation working — should completed dirs be tarred there and removed from \`work/\`?
- Is there a max-age / max-size policy that should exist and doesn't?
- Per-impl differences — does swe-rebench-v2 leave scratch behind that prediction.v1 cleans up?
- Are these specific to dev-layout vs production-install? Acceptance operator at \`~/.jinn-testnet-acceptance/\` is 172KB total, suggesting the issue is not universal.

## Acceptance criteria

- [ ] Root cause documented for *why* 412 work-dirs persist (which path leaks, under which conditions)
- [ ] Proposed fix (could be: cleanup-after-delivery hook, periodic reaper, max-age TTL, size cap, per-impl policy)
- [ ] One-shot cleanup script for existing dirs that's safe to run on a live operator
- [ ] Regression test that exercises the cleanup path

## Impact

- Disk pressure on operator hosts (real today; operator-app-principle violated by a daemon that leaks 49GB)
- Slows substrate-snapshot operations (need to exclude this subtree to be feasible)
- Likely correlated to general operator-host reliability degradation over time

## Where this surfaced

Caught during release-readiness brainstorm (in flight 2026-05-19) while sizing the test-operator substrate. The fact that the substrate adoption had to exclude 49GB of working-dir cache was the visible signal of the underlying cleanup gap.

ritsuKai2000:
## Summary

The TaskEngine provisions a heavy per-task scratch directory under `engine.workingDirRoot` (default `~/.jinn-client/engine/work/<requestId>/`) — a cloned repo, a per-task plugins copy, a `system_snapshot.tar.gz`, session transcripts and an env dir. **Nothing ever removed them.** A long-lived operator hit 412 dirs / 49 GB (#320).

### Root cause

`TaskEngine.process()`'s `COMPLETE` / `FAILED` cases are a literal `// Terminal — nothing to do.` no-op. There is no cleanup-after-delivery hook, no periodic reaper, no max-age/size policy anywhere in the engine, harness or runner code. The `work-archives/` directory observed on the operator is a leftover from an older daemon version — `work-archive` appears nowhere in the current codebase, so archive rotation isn't "broken", it simply doesn't exist. The `system_snapshot.tar.gz` is itself written *inside* the workingDir at `pack()` time and is uploaded to IPFS at delivery — so once a task is terminal the entire on-disk directory is dead weight.

### The fix

- **`work-dir-reaper.ts`** — single decision point for what may be deleted. Conservative by design: an in-flight task's dir is **never** removed regardless of age; a terminal task's dir (evidence already on-chain / on IPFS) is removed immediately; a directory with no DB row at all (crash-orphan, older daemon, wiped DB) is removed only once older than a 24h TTL so we never race a task the DB hasn't observed yet. Never throws — filesystem errors are collected into the report.
- **`TaskRunPersistence.getTerminal()`** — enumerates COMPLETE/FAILED runs.
- **`TaskEngine.reapWorkDirsNow()`** — runs every `tick()`. No separate loop to keep alive or crash; idempotent and cheap (one `readdir` + a few `rmSync`). Tunable via the new optional `workDirReaper` engine option (`orphanMaxAgeMs`, `disabled`).
- **`scripts/cleanup-engine-work.ts`** (`yarn cleanup:engine-work`) — one-shot backlog drainer to clear the existing 412-dir / 49 GB backlog. **Safe on a live operator**: opens the daemon's SQLite store *read-only*, dry-run by default (prints what would go + reclaimable size), `--apply` to delete, `--orphan-age` tunable. Same in-flight protection as the reaper.

### Retention default

Terminal dirs are reaped immediately (snapshot already on IPFS); orphans get a 24h TTL. A time-based `work-archives/` rotation was considered and rejected — the snapshot tarball is already pinned, so archiving it again to local disk just re-leaks disk more slowly.

## Test plan

- [x] `work-dir-reaper.test.ts` — terminal / in-flight / orphan-age matrix, missing-root no-op, 12-dir backlog drain (6 tests)
- [x] `engine-work-dir-cleanup.test.ts` — `reapWorkDirsNow()` + `tick()` integration against a real `Store` DB and real temp dirs, including the in-flight-protection invariant (5 tests)
- [x] Full engine suite green (337 tests); persistence + recovery green
- [x] `yarn typecheck` clean
- [x] Cleanup script manually exercised: dry-run + `--apply` against a synthetic work root — removed the COMPLETE dir, protected the RUNNING dir, kept the young orphan
- [ ] Operator runs `yarn cleanup:engine-work -- --apply` once to drain the 49 GB backlog

Closes #320

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -77,6 +77,7 @@
     "release:testnet-acceptance:docker": "node scripts/testnet-acceptance-docker.mjs",
     "release:testnet-acceptance:host": "node scripts/testnet-acceptance.mjs",
     "status": "tsx scripts/status.ts",
+    "cleanup:engine-work": "tsx scripts/cleanup-engine-work.ts",
     "e2e": "tsx test/e2e/validate.ts",
     "e2e:task-first-local": "tsx test/e2e/task-first-local.ts",
     "e2e:claude-code-learner": "tsx test/e2e/claude-code-learner-portfolio-v0.ts",
--- a/client/scripts/cleanup-engine-work.ts
+++ b/client/scripts/cleanup-engine-work.ts
@@ -191,10 +191,12 @@ function main(): void {
     }
   }
 
-  const reclaimable = candidates.reduce(
-    (sum, name) => sum + dirSizeBytes(join(workingDirRoot, name)),
-    0,
+  // Size each candidate once, up front. The dry run reports this as the
+  // reclaimable estimate; the --apply path sums the subset actually removed.
+  const candidateSizes = new Map(
+    candidates.map((name) => [name, dirSizeBytes(join(workingDirRoot, name))]),
   );
+  const reclaimable = [...candidateSizes.values()].reduce((a, b) => a + b, 0);
 
   console.log(
     `  scanned ${entries.length} dir(s); ${candidates.length} removable ` +
@@ -221,7 +223,11 @@ function main(): void {
     orphanMaxAgeMs,
   });
 
-  console.log(`  removed ${report.removed.length} dir(s); ~${humanBytes(reclaimable)} freed.`);
+  const freed = report.removed.reduce(
+    (sum, name) => sum + (candidateSizes.get(name) ?? 0),
+    0,
+  );
+  console.log(`  removed ${report.removed.length} dir(s); ${humanBytes(freed)} freed.`);
   if (report.errors.length > 0) {
     console.log(`  ${report.errors.length} dir(s) could not be removed:`);
     for (const e of report.errors) console.log(`    - ${e.requestId}: ${e.error}`);
--- a/client/scripts/cleanup-engine-work.ts
+++ b/client/scripts/cleanup-engine-work.ts
@@ -1,0 +1,232 @@
+#!/usr/bin/env node
+/**
+ * One-shot engine working-directory cleanup (issue #320).
+ *
+ * The daemon provisions a heavy scratch directory per task under
+ * `engine.workingDirRoot` (default `~/.jinn-client/engine/work/<requestId>/`).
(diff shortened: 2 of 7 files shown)


ritsukai (approved):
Reviewed and approved.

PR #379, opened 2026-05-20
feat(launcher): surface operator-join count on launched dashboard (#351)

### LINKED ISSUE #351: T2.3 multi-op SPA flow: missing operator-count surface on launched dashboard
Discovered while implementing T2.3 (Tier 2 scenarios plan, Task 8).

The scenario expects `data-testid="operator-count"` on the launched-SolverNet dashboard (the page op-a sees after launching). The element should show how many operators have joined the SolverNet (e.g., "1" when op-b has joined).

Current state: `LaunchedSolverNetRecord` (`client/src/dashboard/spa/src/api/types.ts:466`) does NOT include a joined-operator count. The data is not fetched/computed by the SPA today.

To make T2.3 pass, we need either:
(a) Add a daemon endpoint that returns joined-operator count for a manifest CID, surface it in the SPA's launched dashboard, and tag the element with `data-testid="operator-count"`, OR
(b) Have the SPA derive the count from a registry/discovery query, surface it, tag it.

Companion to #349 (T2.1), #350 (T2.2).

Spec: `docs/superpowers/specs/2026-05-19-release-readiness-and-substrate-design.md` §3.
Plan: `docs/superpowers/plans/2026-05-19-tier-2-scenarios-plan.md` Task 9.

ritsuKai2000:
## Summary

Wires an indexer-backed operator count end-to-end onto the launched-SolverNet dashboard, so a launcher can see how many operators participate in their SolverNet.

- **DiscoveryAPI** gains `getSolverNetOperatorCount(manifestCid)` — implemented in the HTTP (GraphQL) and on-chain (RPC floor) backings, threaded through the `withFallback` wrapper. It counts distinct `attempt.operator` Safe addresses across every task whose `manifestDigest === keccak256(toBytes(cid))`. No indexer schema or handler change — it's a pure read-side aggregation over the existing `task` + `attempt` entities, so no re-sync.
- New daemon route `GET /v1/discovery/solvernet-operator-count?cid=`.
- `LauncherLaunchedPage` polls it; `StatusHeader` renders an **Operators** field with `data-testid="operator-count"` (aliased as a hidden span, mirroring the existing `manifest-cid` pattern). The field is omitted — not shown as a misleading `0` — while the discovery query is loading or after an outage.

## Investigation finding (why T2.3's gate stays — partly)

An operator "join" in the app is **purely a local config write** to `joinedSolverNets[<cid>]` on the joining operator's own daemon (`spec/2026-05-05-solvernet-creation-and-launch.md` §12; `POST /v1/operator/join/:cid` in `setup-endpoints.ts`). It leaves **no on-chain footprint** — no event, no tx, no indexer entity. The launcher's daemon has no way to observe it.

The only protocol-observable signal tying an operator to a SolverNet is `TaskAttemptCreated` (the `attempt` indexer entity) — an operator becomes visible network-wide once they **claim a task**. So `getSolverNetOperatorCount` honestly counts *participating* operators, not config-level joins. This is the same shape as the explorer's existing `countDistinct(attempt.operator)`.

A true "operators *joined*" count would need a contract change (an on-chain operator-registration event + new indexer entity/handler) — too large for this PR, and arguably a design question.

## T2.3 status

- The `operator-count` **element-presence** assertion is now **live and unconditional** — T2.3 asserts `getByTestId('operator-count')` renders a real number, catching any regression that drops the surface.
- The **"exactly 1 operator joined"** magnitude assertion **stays gated** behind `JINN_T23_OPERATOR_COUNT_READY`. T2.3's op-b joins but never claims a task, so the honest on-chain count is `0`, not `1`. Removing the gate entirely would make T2.3 fail. Closing the gap needs either op-b claiming a task in-scenario, or the on-chain operator-registration event above — tracked as follow-up.

## Test plan

- [x] `yarn typecheck` clean (client src)
- [x] SPA `tsc -b && vite build` clean
- [x] `test/discovery/http.test.ts` — new `getSolverNetOperatorCount` suite (distinct-operator count, 0-task / 0-attempt cases, GraphQL-error + network-error → `DiscoveryUnavailableError`)
- [x] `test/discovery/with-fallback.test.ts` — new routing test
- [x] `test/api/discovery-endpoint.test.ts` — new route tests (200 / 400 missing-cid / 503 outage)
- [x] `StatusHeader.test.tsx` + `LauncherLaunched.test.tsx` — operator-count rendered / 0 / omitted-on-undefined / omitted-on-query-error
- [ ] T2.3 Playwright run (`run-tier-2`) — requires `BASE_SEPOLIA_RPC_URL`; the `operator-count` element-presence assertion now runs unconditionally

Closes #351

### DIFF
--- a/client/src/api/discovery-endpoint.ts
+++ b/client/src/api/discovery-endpoint.ts
@@ -6,6 +6,7 @@
  *   GET /v1/discovery/plugin-publications?solverType&builderAgentId&includeRevoked
  *   GET /v1/discovery/builder-artifacts?builderAgentId&limit
  *   GET /v1/discovery/plugin-scores?cid&limit
+ *   GET /v1/discovery/solvernet-operator-count?cid
  *
  * Used by the /build SPA route (hfmf) to render published plug-ins, the
  * local operator's published plug-ins, and per-plug-in score history.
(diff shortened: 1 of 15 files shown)


ritsukai (approved):
Reviewed and approved.

Issue #332, opened 2026-05-19
fix: SolverNet join form lets operators select harnesses that are not installed/authed — wire vh74.2 Stage A readiness primitive


ritsuKai2000:
## Run-mode

BUG-FIX (P1)

## Context

In v0.1.6 the SolverNet join form lets operators select Codex even when Codex isn't installed locally; lets them pick Hermes without an OpenRouter / Nous Portal key; lets them pick paid-API-key models with no surfaced auth dependency. Both Oak and rvx hit this in the dogfood (rvx joined with Codex he had no auth for, immediately produced 17/17 failures).

PR #248 (vh74.2 Stage A, merged 2026-05-15) shipped a **per-harness readiness primitive** (`Harness.isReady()` + `HarnessReadinessRegistry` + `/v1/harnesses/<name>/readiness`). The infrastructure is in place; the join form just doesn't use it.

vh74.3 (Stage B, currently blocked on Hermes chain) was the planned SPA-side generalisation — but Stage B's scope is large and the operator-app needs the minimum-viable version of this **now**, not whenever Hermes chain merges.

## Impact

P1. Combined with #322.2 (claim-before-ready cascade) and #329 (Claude Code dropdown bug), this is the cluster that makes the join flow feel "broken" for first-time operators.

## Acceptance criteria

- Join-form harness selector calls `/v1/harnesses/<name>/readiness` and renders not-ready harnesses with a clear non-selectable / install-affordance state.
- The "Save & Join" button gates on all selected-role harnesses reporting ready (solver harness + evaluator harness if applicable).
- The not-ready render includes the readiness primitive's `nextStep` text (e.g., "install Codex CLI", "add OPENROUTER_API_KEY to ~/.jinn-client/config.json").
- Composes cleanly with the broader vh74.3 Stage B work — this fix is the minimum-viable subset that should land before Stage B completes.

## Reference

- Parent: #328 (release feedback).
- PR #248 (vh74.2 Stage A — the primitive this fix consumes).
- #236 (vh74.2 umbrella) and vh74.3 (Stage B, blocked on Hermes chain).

## Discovered in

2026-05-19 v0.1.6 dogfood — both Oak and rvx.

Issue #330, opened 2026-05-19
fix: daemon claims tasks before any harness is ready — N/N failure cascade on fresh join


ritsuKai2000:
## Run-mode

BUG-FIX (P0, hotfix-track per #328 Option C)

## Context

A new operator (rvx) joined SWE-rebench v2 in v0.1.6 with no harness configured/authed. Within minutes the dashboard showed Tasks 17 / Failed 17, then Tasks 26 / Failed 26 — the daemon was claiming tasks despite having no working harness to execute them with, and burning testnet gas top-ups on the failed deliveries.

On mainnet this would be catastrophic: real gas burned for guaranteed-failed claims, reputation damage, and operator-side surprise. Even on testnet today, every failed claim costs faucet-budget runway and confuses first-run operators into thinking the daemon is working.

PR #248 (vh74.2 Stage A, merged 2026-05-15) shipped a per-harness readiness primitive (`Harness.isReady()` + `HarnessReadinessRegistry`) and a claim-loop gate primitive. The gate is in place at the daemon level but does NOT appear to be wired against the SolverNet's selected harness at claim time.

## Impact

P0. First-run operators on v0.1.6 will encounter this immediately if they join a SolverNet before authing their harness. Pretty much the worst possible "first impression" outcome — Jinn appears to silently fail and burn gas.

## Acceptance criteria

- Claim loop refuses to claim a task whose SolverNet's selected solver harness is not ready (per `Harness.isReady()`).
- Same for the evaluator harness on evaluator-role claims.
- Operator sees a clear surface ("harness not ready — sign in / install ..." per harness's `nextStep`).
- Regression test: spin up a daemon with a joined SolverNet whose harness is intentionally not-ready; assert zero claims attempted; assert operator-facing reason surfaced.

## Reference

- Parent: #328 (release feedback).
- Related: #236 (vh74.2 per-harness auth), #322.4 (#332 — install-precheck in join form).
- PR #248 (vh74.2 Stage A — readiness primitive that this fix consumes).

## Discovered in

2026-05-19 v0.1.6 dogfood — rvx, first 30 minutes of fresh install.

PR #393, opened 2026-05-20
fix: v0.1.6 operator-app dogfood follow-ups (#328)

ritsuKai2000:
## Summary

Follow-up fixes from the **v0.1.6 operator-app dogfood** (umbrella #328). Two test rounds against a live daemon (Base Sepolia) surfaced operator-facing issues and two daemon performance bugs; this PR lands all of them. 13 fix branches integrated; merged with zero conflicts.

## Fixes

**Operator dashboard / join form**
- STATUS hero tile now shows *why* it needs attention (was a bare "ATTENTION")
- **Hermes harness readiness now gates on OpenRouter auth** — was a false-positive `ready: true` with no model provider configured, which defeated the join-form readiness gate (#332) and the claim-loop gate (#330/#348), re-opening the "burned failed claims" dogfood bug. Hermes is now OpenRouter-only.
- Join-success card gained a **"Restart node now"** action (it instructed operators to restart but offered no way to)
- Restart-requested notice now auto-clears
- Offline banner no longer renders an inert fake `jinn run` button
- Spurious deprecated-`prediction.v1` "ATTENTION" no longer fires for operators who never joined a prediction SolverNet
- Internal `/api/...` paths no longer leak into operator-facing copy
- Inline-help "Read more" links fixed (`cargo/client/docs` → `client/docs`, `blob/main` → `blob/next`)
- Inline-help copy tightened; verbose cost-estimate breakdown shortened to a one-line caveat

**Daemon**
- Mech evaluation loop prunes terminal opportunities instead of re-scanning on-chain history every cycle
- Engine-watcher skip-caches a stale task backlog (TTL) and yields to the event loop — stops the "Daemon offline" banner flicker

**Tooling**
- testing-jinn-app skill: documented that `--no-ui` disables the in-app restart-respawn; fixed a stale `cargo/client` path

## Verification

- `typecheck` clean; full suite **3914 tests pass** (480 files); daemon suite **82 pass**
- Every operator-facing fix live-walked against a daemon on Base Sepolia
- Offline-banner flicker confirmed resolved (simulated the SPA detector: 1 isolated probe miss / 40, never 2 consecutive)

## Known follow-ups (not in this PR)

- **#14 residual** — occasional sub-threshold daemon event-loop micro-spikes remain (no longer trip the offline banner); needs a CPU profile, not another blind round. Heavily amplified by the test fleet's on-chain backlog.
- **Pre-existing** — 3 flaky `EADDRINUSE:7331` daemon tests under the parallel pool (verified pre-existing); `tsc` emitting `.js` into `src/` (build-hygiene).

Umbrella: #328

### DIFF
--- a/.claude/skills/testing-jinn-app/SKILL.md
+++ b/.claude/skills/testing-jinn-app/SKILL.md
@@ -19,10 +19,13 @@ The jinn app is the operator dashboard SPA at `client/src/dashboard/spa/`, serve
 
 ## Daemon spawn (shared by both recipes)
 
-All commands assume cwd = `jinn-mono/cargo/client`.
+All commands assume cwd = `jinn-mono/client`.
 
 1. Build: `yarn build` — produces `dist/bin/jinn.js` and the SPA bundle in `dist/dashboard/`. **Re-run after every SPA source edit** — the daemon serves the bundled SPA from disk.
 2. Spawn: `node dist/bin/jinn.js run --no-ui`. **Against the operator's real `~/.jinn`, that's the whole command** — the daemon auto-reads `~/.jinn-client/keystore-password` (written at first bootstrap) when `JINN_PASSWORD` is unset. Do NOT ask the operator for a password.
+
+   > **CAVEAT — Restart-button / respawn tests must NOT use `--no-ui`.**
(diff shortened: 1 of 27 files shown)


ritsukai (approved):
Reviewed — v0.1.6 operator-app dogfood follow-ups. All 12 findings fixed and verified; full suite green.

PR #411, opened 2026-05-20
fix(engine): canAcceptTask uses cached harness readiness, not a per-task blocking probe

### LINKED ISSUE #398: engine.canAcceptTask() runs harness isReady() (blocking spawnSync) per task, uncached
## Context

Surfaced during the v0.1.6 operator-app dogfood (#328), efficiency review of #393.

`engine.canAcceptTask()` (`client/src/harnesses/engine/engine.ts`) calls `impl.isReady()` **directly, uncached, for every task announcement** in the engine-watcher loop (`client/src/daemon/daemon.ts` `_runEngineWatcherLoop`).

The Hermes harness's `isReady()` runs **blocking `spawnSync` child processes**. #393's Hermes-readiness fix (gating on OpenRouter auth) raised that from one `spawnSync` to **two** per call. So a backlog of Hermes-harness tasks pays 2× blocking child-process spawns per task on the first scan, and again after each skip-cache TTL expiry.

## Impact

Per-task blocking child-process spawns on the engine-watcher hot path — directly works against the event-loop-responsiveness the #393 daemon fixes aimed for. Partially masked by #393's own skip-cache (30s TTL), but the first scan and every post-TTL re-scan still pay it.

## What's needed

`canAcceptTask`'s readiness check should consult the already-wired, already-cached `HarnessReadinessRegistry` snapshot (`isReadyForClaim`, O(1)) instead of calling `impl.isReady()` directly.

## Acceptance

`canAcceptTask`'s harness-readiness check is O(1) — no `spawnSync` (or other blocking work) per task.

Refs #328 · follow-up to #393.

ritsuKai2000:
Closes #398.

## Problem

`TaskEngine.canAcceptTask()` ran `impl.isReady()` uncached for **every task announcement** on the engine-watcher hot path. For the Hermes harness, `isReady()` spawns two blocking `spawnSync` child processes — so a task backlog paid per-task blocking spawns and starved the daemon event loop.

## Fix

Dropped the redundant per-task `impl.isReady()` probe from `canAcceptTask`'s path — **scoped, not blanket-removed**. The shared `runnableFailureReason` helper is also used by `claim()`, which genuinely needs the probe, so a `skipReadinessProbe` option is threaded through:

- `canAcceptTask` sets `skipReadinessProbe: true`. Readiness for the claim gate is served O(1) from the daemon's cached `HarnessReadinessRegistry` — the engine-watcher loop calls `gateClaimByReadiness(...)` one line after `canAcceptTask` returns. A ~tick-stale snapshot is acceptable; harness readiness changes on a minutes scale.
- `claim()` (the DISCOVERED → CLAIMED transition) leaves it unset — the `impl.isReady()` probe stays as the authoritative pre-execution gate. It runs once per *claimed* task, not per announcement.

Routing `canAcceptTask` through the cached registry directly was rejected: `TaskEngine` has no registry reference (it lives on the daemon, threaded into `gateClaimByReadiness`), and the watcher already calls that gate — adding a registry dependency to the engine would be the messier change.

## Tests

3 new regression tests in `claim-gate.test.ts`:
- `canAcceptTask` does NOT invoke `impl.isReady()` (spy asserts zero calls)
- `canAcceptTask` accepts even when the harness reports not-ready (gate is downstream)
- `claim()` still probes `impl.isReady()` exactly once

Verification: `tsc --noEmit` clean; per-file vitest green — `claim-gate` 10/10, `engine` 51/51, `joined-eligibility` 18/18, `prediction-v1-runtime-gate` 8/8, `claim-readiness-gate` 5/5, `skip-log-dedup` 19/19, `daemon` 3/3.

### DIFF
--- a/client/src/harnesses/engine/engine.ts
+++ b/client/src/harnesses/engine/engine.ts
@@ -424,12 +424,39 @@ export class TaskEngine {
     }
   }
 
+  /**
+   * Pre-claim acceptance check used by the daemon's engine-watcher loop.
+   *
+   * Performance contract (issue #398): this runs once per task announcement,
+   * on the engine-watcher hot path, for every observed task. It MUST NOT
+   * perform per-task blocking I/O. In particular it does not probe
+   * `impl.isReady()` — for the Hermes harness that runs two blocking
+   * `spawnSync` child processes, so a backlog would pay per-task blocking
+   * spawns and starve the daemon event loop.
+   *
+   * Harness readiness for the claim gate is instead served O(1) from the
+   * daemon's cached `HarnessReadinessRegistry` snapshot: the engine-watcher
+   * loop calls `gateClaimByReadiness(...)` immediately after `canAcceptTask`
+   * returns. A ~tickIntervalMs-stale snapshot is acceptable — harness
+   * readiness changes on a minutes scale (auth/config) and the daemon
+   * already trusts that cached registry for its post-`canAcceptTask` gate.
+   *
+   * `claim()` (the DISCOVERED → CLAIMED transition) still probes
+   * `impl.isReady()` directly — it runs once per claimed task, not per
+   * announcement, and is the authoritative pre-execution gate.
+   */
   async canAcceptTask(input: {
     solverType?: string;
     taskRole?: 'restoration' | 'evaluation';
     task?: Task;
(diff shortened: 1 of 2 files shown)


ritsukai (approved):
Reviewed and approved.

PR #412, opened 2026-05-20
fix(mech): classify terminal evaluation reasons from the structured revert name

### LINKED ISSUE #399: mech adapter: isTerminalEvaluationReason regex-reparses a formatted revert string
## Context

Surfaced during the v0.1.6 operator-app dogfood (#328), code-reuse review of #393.

`isTerminalEvaluationReason` (`client/src/adapters/mech/adapter.ts`, added in #393) decides whether an evaluation opportunity is permanently unclaimable. It does so by **regex-stripping** `Name(args)` → `Name` from `canClaimEvaluation`'s formatted `reason` string, then looking the bare name up in `isNonRecoverableInnerRevert`.

But that `reason` string was itself **formatted from structured `{ name, args }` data** one layer down (`safe-revert.ts` / `client/src/adapters/mech/contracts.ts`). So the pipeline is: structured revert → format to `Name(arg, arg)` string → regex-strip back to `Name` → set lookup.

## Impact

The format → regex-unformat round-trip is fragile: an arg value containing a `(` (e.g. a stringified tuple) corrupts the strip, and the `flattenErrorMessage` fallback path produces arbitrary text the regex also mangles. A mis-stripped name silently fails the terminal check — either re-scanning a terminal opportunity forever, or (worse) wrongly pruning a still-claimable one. Currently low-probability (revert args are mostly numeric), but brittle.

## What's needed

Have `canClaimEvaluation` return the structured revert `name` alongside (or instead of) the formatted `reason`, so the terminal check calls `isNonRecoverableInnerRevert(name)` on structured data with no re-parsing.

## Acceptance

Terminal-reason classification uses the structured revert name, not a regex over formatted text.

Refs #328 · follow-up to #393.

ritsuKai2000:
Closes #399.

## Problem

`isTerminalEvaluationReason` decided whether a `canClaimEvaluation` failure was permanently unclaimable (prune) or transient (retry) by **regex-unformatting the operator-facing `reason` string** — stripping a `(...)` arg suffix to recover the bare revert name. That round-trip was fragile:

- an arg value containing a `(` corrupted the strip
- the `flattenErrorMessage` fallback produced arbitrary text the regex mangled

Either case silently mis-classified opportunities. A false-prune drops still-claimable work.

## Fix

Classify on the **structured** revert name decoded straight from the inner revert data, never on the formatted string.

- `safe-revert.ts` — new `formatKnownRevertDetail(error)` returns `{ name, reason }` (the bare revert identifier + the operator-facing formatted string). `formatKnownRevert` becomes a thin wrapper over it — no behavior change for its existing callers.
- `contracts.ts` — `canClaimEvaluation`'s failure shape gains `revertName: string | null` (new `CanClaimEvaluationFailure` interface). `reason` is unchanged, still used for log lines.
- `adapter.ts` — `isTerminalEvaluationReason` now takes the structured `revertName` and calls `isNonRecoverableInnerRevert` directly. The regex strip is gone.

When in doubt we still keep (transient): a false-keep costs one RPC, a false-prune loses real work.

## Tests

Regression coverage added for the arg-with-`(` corruption case and the `flattenErrorMessage` fallback path. 48 tests green across `adapter.test.ts` + `safe-revert.test.ts`; `tsc --noEmit` clean.

### DIFF
--- a/client/src/adapters/mech/adapter.ts
+++ b/client/src/adapters/mech/adapter.ts
@@ -77,20 +77,23 @@ const DEFAULT_MECH_DELIVER_BACKFILL_LOOKBACK_BLOCKS = 100_000n;
 const EVALUATION_RETRY_YIELD_EVERY = 10;
 
 /**
- * Decide whether a `canClaimEvaluation` failure reason means the opportunity
- * can NEVER become claimable (terminal, prune it) versus one that could still
- * clear later (transient, keep retrying).
+ * Decide whether a `canClaimEvaluation` failure means the opportunity can NEVER
+ * become claimable (terminal, prune it) versus one that could still clear later
+ * (transient, keep retrying).
+ *
+ * Classification is done on the *structured* `revertName` decoded straight from
+ * the inner revert data — not by regex-unformatting the operator-facing `reason`
+ * string. The format→regex round-trip was fragile: an arg value containing a
+ * `(` corrupted the strip, and the `flattenErrorMessage` fallback produced
+ * arbitrary text the regex mangled, silently mis-classifying opportunities.
  *
  * A false-keep (re-checking a dead opportunity) only costs one more RPC; a
  * false-prune (dropping a still-claimable opportunity) loses real work — so
- * when in doubt we keep. Anything that does not resolve to a known
- * non-recoverable revert name is treated as transient.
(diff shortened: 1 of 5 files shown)


ritsukai (approved):
Reviewed and approved.

Issue #406, opened 2026-05-20
[#220] PR 4: real tJINN earned dashboard


ritsuKai2000:
## Run-mode

FIX / operator dashboard. Target branch: `next`.

## Problem

The operator dashboard's JINN tile reads `rewards.pendingStakingRewardsWei`, which is the OLAS staking proxy pending reward surface, not the Sepolia tJINN ERC-20 minted by `JinnDistributor`. Operators need to see their real tJINN Safe balance.

## Acceptance

- `/v1/status` exposes top-level `tJinn` with `state`, `chainId`, `tokenAddress`, `safeBalanceWei`, `safeCount`, `services[]`, and `error`.
- Status reads Sepolia ERC-20 `balanceOf` for token `0x0bc0B2f733bF4229FD58Baaac5ebFEf2AEc83C4A`.
- Shared Safe addresses are deduplicated before summing top-level balance.
- Overview/HeroStats read `status.tJinn.safeBalanceWei`, not `rewards.pendingStakingRewardsWei`.
- Tile label is `tJINN earned` and shows pending/error copy instead of a bare zero while the Sepolia balance read is unavailable.
- Remove the old `Claim now` action from this tile because it submits stOLAS reward claims, not tJINN earning claims.
- Focused API/dashboard tests, typecheck, and SPA build pass.

Parent: #220

ritsuKai2000:
**Re-scoping note (2026-05-21).** The implementation target for this issue has moved.

PR #410 implemented it against the HeroStats **JINN tile** — but the operator-app Funds/Rewards split (Task 3.3) has since landed on `next`: the JINN tile is gone, and rewards now live in `client/src/dashboard/spa/src/pages/overview/RewardsCard.tsx`.

`RewardsCard` today shows `claimableJinn` + `claimedJinnLifetime` + last-claim — but **not** the real on-chain tJINN balance. So this issue should now be implemented as:

> Surface the real `JinnDistributor`-minted tJINN balance in `RewardsCard`, sourcing it correctly rather than from `rewards.pendingStakingRewardsWei` (the OLAS staking-proxy pending-reward surface).

#410 has been closed as superseded; its branch `codex/220-pr4-tjinn-dashboard` still has reusable daemon-side logic (the real-balance `/v1/status` field) worth cherry-picking.

ritsuKai2000:
**Re-scoping note 2 (2026-05-21).** The display target has moved again.

[...]

PR #447, opened 2026-05-21
[codex] expose real tJINN status balance

ritsuKai2000:
## Summary

Part of #406.

- Adds top-level `/v1/status.tJinn` with Sepolia chain id, token address, Safe balance, Safe count, per-service rows, and public state/error fields.
- Reads real Sepolia tJINN ERC-20 `balanceOf` for token `0x0bc0B2f733bF4229FD58Baaac5ebFEf2AEc83C4A`.
- Deduplicates shared Safe addresses before summing the top-level balance.
- Keeps `rewards.pendingStakingRewardsWei` separate from real tJINN earned and redacts raw RPC failures before they reach `/v1/status`.

## Verification

- `yarn vitest run test/api/gather-status.test.ts test/api/status-build.test.ts`

Part of #406

### DIFF
--- a/client/src/api/gather-status.ts
+++ b/client/src/api/gather-status.ts
@@ -930,10 +930,6 @@ export async function gatherGatheredStatusRaw(
     transport: http(status.rpcUrl),
   });
 
-  // The tJINN read started above (overlapping the Base-RPC fan-out). It never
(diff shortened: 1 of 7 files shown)


ritsukai (approved):
Reviewed and approved.

PR #482, opened 2026-05-21
feat(client): expose live generator state on launched records

ritsuKai2000:
## Issue

Implements #471.

## Root cause

The generator runtime state already existed in the daemon and SPA types/rendering, but the launched-record API returned persisted records directly. `GET /v1/solvernets/launched/:id` and `GET /v1/solvernets/launched` never overlaid the live generator snapshot, so the SPA Generator panel could render a blank Last poll even while the generator had real runtime state.

## Changes

- Added regression coverage for live `generatorState` overlay and persisted fallback on both the single launched-record endpoint and the owned launched-record list.
- Added an optional `getGeneratorState(solverNetId)` launch dependency and a narrow response helper that overlays live state when present.
- Wired `main.ts` to resolve a launched record solver type from `solverNetId` and read the live generator-state map populated by launched-record generator wiring.

## Command evidence

Initial regression, after adding tests and building the SDK prerequisite:

```text
yarn vitest run test/api/solvernets-endpoints.test.ts
FAIL test/api/solvernets-endpoints.test.ts > GET /v1/solvernets/launched/:id (Task 14) > overlays live generatorState on single-record and list responses
AssertionError: expected persisted generatorState to deeply equal liveState
Test Files  1 failed (1)
Tests  1 failed | 77 passed (78)
```

Final focused test:

```text
yarn vitest run test/api/solvernets-endpoints.test.ts
✓ test/api/solvernets-endpoints.test.ts (78 tests) 477ms
Test Files  1 passed (1)
Tests  78 passed (78)
Duration  893ms
```

Typecheck:

```text
yarn typecheck
➤ YN0000: · Yarn 4.13.0
➤ YN0000: · Done in 0s 118ms
exit 0
```

Build:

```text
yarn build
vite v5.4.21 building for production...
✓ 1833 modules transformed.
✓ built in 1.80s
exit 0
```

## Notes

PR target: `next`. This is the bottom implementation branch stacked on the planning commit from `feat/471-generator-health-visibility`.

PR #484, opened 2026-05-22
fix(client): recover underpriced tx replacements

### LINKED ISSUE #469: fix(client): on-chain txs jam with 'replacement transaction underpriced'
## Context

The daemon's Safe `execTransaction` submissions intermittently fail at the RPC with `replacement transaction underpriced` (viem `eth_sendRawTransaction` against the Base Sepolia RPC). A transaction with the same nonce is pending in the mempool and the resubmit does not raise the gas price enough to replace it.

## Impact

Claims, deliveries, and (likely) generator task-posts stall behind a stuck nonce. Contributes to claims expiring before settlement.

## Evidence

- Daemon log: repeated `Details: replacement transaction underpriced` / `Version: viem@2.47.18 | replacement transaction underpriced` on `eth_sendRawTransaction`.
- `/v1/status` taskRuns: `Safe execTransaction inner revert (estimate): TCAttemptClaimExpired(151, 0)` — claim expired before settlement.

## Acceptance criteria

- Transaction resubmission bumps the gas price by at least the replacement minimum (geth requires ≈+10%).
- Stuck-nonce detection and recovery, so the daemon does not wedge behind a single pending transaction.

## Run-mode

`fix`

ritsuKai2000:
## Summary

Closes #469.

- Persist submitted EOA nonce fee snapshots keyed by chain/from/nonce so replacement retries bump against the previously submitted fee, not only fresh RPC estimates.
- Pin explicit EOA nonce during generic and Safe transaction retries so resubmissions replace the pending tx.
- Add stuck nonce detection for `pending > latest` with an unresolved stale ledger entry, clearing via a 0-value self-transfer using replacement-bumped fees.
- Wire the daemon Store as the persisted tx submission ledger.

## Validation

- `cd client && yarn vitest run test/tx-retry.test.ts test/store/store.test.ts test/daemon/balance-topup-loop.test.ts test/earning/agent-wallet-binding.test.ts test/earning/migrate-agent-id.test.ts test/adapters/mech/eviction-recovery.test.ts test/earning/orphan-sweep.test.ts test/earning/stolas-claim.test.ts test/earning/bootstrap-mech-safe-direct.test.ts` -> 9 files, 92 tests passed
- `cd client && yarn typecheck` -> passed
- `cd client && yarn build` -> passed; Vite emitted the existing chunk-size warning

## Notes

A broader `yarn test` was attempted before the lightweight-mock compatibility guard landed. It exposed mocked-client failures that are now covered by the affected targeted rerun above; it also hit an existing helper startup timeout in `test/helpers/multi-op-daemon.test.ts`.

### DIFF
--- a/client/src/adapters/mech/safe.ts
+++ b/client/src/adapters/mech/safe.ts
@@ -12,10 +12,9 @@ import { privateKeyToAccount } from 'viem/accounts';
 import { base } from 'viem/chains';
 import { SAFE_ABI } from './types.js';
 import {
-  getDefaultTxSubmissionLedger,
-  recoverStuckNonceIfNeeded,
+  type TxSubmissionLedger,
+  withNonceLedger,
   withRecoverableRetry,
-  viemFeeOverridesForAttempt,
 } from '../../tx-retry.js';
 import {
   SafeInnerRevertError,
@@ -37,6 +36,10 @@ export interface SafeTransactionParams {
   data: Hex;
 }
 
+export interface SafeExecutionOptions {
+  ledger?: TxSubmissionLedger;
+}
+
 // Per-Safe transaction lock to prevent nonce races when concurrent
 // loops (creator + harness) share the same Safe
 const safeLocks = new Map<string, Promise<void>>();
@@ -45,6 +48,7 @@ export async function executeSafeTransaction(
   publicClient: PublicClient,
   walletClient: WalletClient,
   params: SafeTransactionParams,
+  options: SafeExecutionOptions = {},
 ): Promise<Hex> {
   const lockKey = params.safeAddress.toLowerCase();
 
@@ -58,7 +62,7 @@ export async function executeSafeTransaction(
   await pending;
 
   try {
-    const hash = await executeSafeTransactionInner(publicClient, walletClient, params);
+    const hash = await executeSafeTransactionInner(publicClient, walletClient, params, options);
     return hash;
   } finally {
     releaseLock();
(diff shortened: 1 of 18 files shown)


ritsukai:
## Code review findings (`/code-review`)

Reviewed as part of the 5-PR stack (code-reuse / code-quality / efficiency passes). Four findings on the tx-retry + Safe-ledger code — none block merge, all worth a follow-up. The `tx-retry.ts` retry/nonce core and the `Store` `tx_submissions` persistence are otherwise clean (they correctly extend the canonical retry module rather than forking it).

### Medium

**1. Global mutable ledger singleton** — `tx-retry.ts:279`
`defaultTxSubmissionLedger` is a module-level `let` mutated by `setDefaultTxSubmissionLedger`. Verified `main.ts:865` sets it before bootstrap runs, so this is **not a live bug** — but it is fragile init-order coupling and makes tests order-dependent across files sharing the module. Prefer threading the ledger explicitly through the Safe-execution `opts` object.

**2. Duplicated Safe-path ledger orchestration** — `safe.ts` / `safe-adapter.ts`
The same ~40-line preamble (resolve chainId -> `getDefaultTxSubmissionLedger()` -> `recoverStuckNonceIfNeeded` -> pin pending nonce) plus the post-send `recordTxSubmission` / `markTxSubmissionResolved` pair is copy-pasted across `executeSafeTransactionInner`, `executeSafeTxDirect`, and a third near-copy inline in `viemSendTransactionWithRetry`. Extract one `withNonceLedger(...)` helper in `tx-retry.ts`.

**3. Redundant `getChainId` RPC + missed `Promise.all`** — `tx-retry.ts:534`
`recoverStuckNonceIfNeeded` re-resolves chainId the caller already resolved (two round-trips on every Safe tx), and `await resolveTxChainId(...)` runs serially *before* the nonce `Promise.all` although all three reads are independent. Fold chainId into the `Promise.all` and accept an optional pre-resolved `chainId`.

**4. Type-safety holes** — `tx-retry.ts:413,521`
`viemFeeOverridesForAttempt` returns a union that forces `as` casts at every call site; `recoverStuckNonceIfNeeded` uses `tx: any` + `as unknown as {...}` instead of viem's `PublicClient` / `WalletClient` types (which already type `getChainId` / `getTransactionCount`). Share one discriminated result type and drop the casts.

_Posted for the reviewer — no code pushed._

ritsukai (approved):
Approving as part of the 5-PR stack promotion (#482-#486). CI green; /code-review findings on the stack were addressed (branches advanced past the reviewed tip).

Issue #492, opened 2026-05-22
refactor: implement SWE-rebench-v2 fill-the-pool generation and claim caps


ritsuKai2000:
## Context

Follow-up implementation for #487 and DR-2026-05-22-a, after docs PR #488 ratifies the SWE-rebench-v2 task-generation and claiming semantics.

The current implementation has two live failure modes:

- one recent SWE posting globally stalls the generator pool;
- one SWE posting can expose the old flat `maxClaims: 50` claim budget instead of the remaining `N_target_successes - successful_count` budget.

The implementation plan also found a shared batching gap: emitting `Task[]` from a generator is insufficient unless the creator loop posts all candidates and generated tasks can opt into distinct bucket keys.

## Run-mode

REFACTOR

Handbook discipline: TDD + integration tests on the shared creator/posting path and generator contract surfaces. Stack as two implementation PRs after #488:

1. Shared batching infrastructure: additive creator/source/posting support, independently green.
2. SWE-rebench-v2 policy/config/UI implementation on top.

## Acceptance Criteria

- `CreatorLoop.tick()` can post all candidates collected for a tick, fault-isolates per candidate, and preserves intended idempotency.
- Generated task sources support an opt-in bucket-key override; duplicate bucket keys still dedupe.
- SWE-rebench-v2 generator fills the pool up to `post_batch_size`, has no cooldown gate, derives live/repostable state from `last_posted_at + posting_window_ms`, and never selects saturated or abandoned instances.
- Each SWE posting sets `maxClaims = N_target_successes - successful_count` and `maxClaimsPerOperator = min(remaining, override ?? remaining)`.
- SWE generator config uses `N_target_successes`, `N_max_postings_per_task`, `posting_window_ms`, `post_batch_size`, optional top-level `maxClaimsPerOperator`, and `claimLeaseTtlSeconds`; legacy `cooldown_ms` / nested `claimPolicy` is tolerated on read and normalized forward.
- Launcher state surfaces pool progress: `poolSize`, `posted`, `unposted`, `live`, `repostable`, `saturated`, `abandoned`.
- Existing task 157 with old `maxClaims=50` is documented as accepted transient legacy exposure.

## Required Tests

Regression-first:

- one prior post must not stall the SWE pool;
- one posting must not exceed remaining N claims;
- a generator batch lands through source -> creator -> posting service;
- a mid-batch post failure does not prevent later candidates from posting.

Shared-loop guards:

- prediction-style single-candidate source remains unchanged;
[...]

PR #488, opened 2026-05-22
docs: SWE-rebench-v2 generation & claiming semantics (DR for #487)

ritsukai:
## Summary

Design / scoping deliverable for #487 — defines the intended protocol semantics
for SWE-rebench-v2 task generation & claiming **before any code change**, as the
issue requested. No code is touched in this PR.

- Adds a **proposed DR** — `log/decisions/2026-05-22-swe-rebench-v2-generation-claiming-semantics.md`
- Adds a proposed-amendment banner to spec §3.6 of `docs/superpowers/specs/2026-05-06-agent-harness-solvernet-design.md`

## The design

`N_target_successes` is the single memorisation cap; every other mechanism
exists only to respect it.

- **Generator-sized claim caps** — each posting's `maxClaims` and
  `maxClaimsPerOperator` both derive from `N − successful_count`. `TaskCoordinator`
  enforces `maxClaims` per-Task on-chain, so a posting can't exceed the cap. Kills
  the over-solving mode (one posting → up to 50 trajectories).
- **Fill-the-pool, no cadence** — post every scorable unposted instance,
  batch-throttled. The global cooldown gate (the under-posting bug) is deleted.
- **Retry-on-expiry** — one live posting per instance; the existing 7-day window
  is the repost trigger. This is load-bearing: `TaskCoordinator` claim slots are
  a one-way budget (`expireAttempt` never decrements `claimCount`), so expiry →
  repost is what hands a stranded instance a fresh budget. Churn accepted.
- **Generic generator knobs** — expiry / cooldown / pool source / target-success
  / attempt-budget are per-SolverNet knobs; swe-rebench-v2 is one config.

Resolves both failure modes from #487 and answers all five of its open questions.

## Status — proposed, not ratified

The DR is `status: proposed`. It awaits Captain ratification.

**On ratification (a separate session):**
- Rewrite spec §3.6 to swe-rebench-v2's config + add a generic Task-generator
  framework sub-section to the same spec.
- File the `refactor`-shape implementation issue (scope is enumerated in the DR's
  Consequences).

**Deferred:** the operator-app `OPERATOR-APP-SPEC.md` §2.13 Launcher component
spec (a canonical-doc change — CODEOWNERS + Discussion). The DR's Consequences
flag the concrete app-surface reshape (`cadenceMs` removed; poll summary →
fill-the-pool/saturation progress).

## Review checklist

- [ ] Semantics in the DR match intended protocol behavior
- [ ] Ratify the DR (`proposed` → `ratified`)
[...]

Issue #487, opened 2026-05-22
Clarify SWE-rebench-v2 task generation, claim capacity, and success-target semantics


ritsuKai2000:
## Problem

SWE-rebench-v2 task generation currently mixes three related but separate concepts without a clear contract:

- posting new on-chain tasks for a benchmark instance
- allowing operators to claim/solve an already-posted task
- counting successful evaluator verdicts toward the instance's target

This makes the intended cadence and stopping behavior hard to reason about, and the current live behavior appears inconsistent with the expected SolverNet workflow.

## Observed Behavior

During the Phase 2 testnet run on 2026-05-22, the launched SWE-rebench-v2 SolverNet had this generator config:

```json
{
  "N_target_successes": 3,
  "N_max_postings_per_task": 10,
  "cooldown_ms": 86400000
}
```

The generator posted one task:

- instance: `carsdotcom__skelebot-280`
- request/task id: `157`
- posted at: `2026-05-22T05:19:21.797Z`

After that, generator state reported polls with the vetted pool available but no more posts:

```json
{
  "cadenceMs": 86400000,
  "lastPollSummary": {
    "evaluated": 20,
    "posted": 0,
    "skipped": 20
  }
}
```

The code appears to apply `cooldown_ms` globally across the eligible pool before selecting a candidate, so one recent post blocks all other vetted instances. Separately, the selector itself treats cooldown as per-instance. Relevant files:

- `client/src/solver-types/swe-rebench-v2.ts`
- `client/src/solver-types/swe-rebench-v2-auto.ts`

## Semantic Ambiguity

The current config names suggest that `N_target_successes: 3` means an instance should stop once three successful verdicts have landed. But a single posted task can have many solver claims, based on claim policy defaults. That means one task posting could theoretically yield far more than three successful solver attempts, even if the generator never reposts that instance.

So there are two separate saturation surfaces:

- generator-level saturation: do not post more tasks for an instance after enough successful verdicts
- claim-level saturation: how many operators may still claim/solve an already-posted task

The current generator only tracks per-instance `posted`, `successful`, and `last_posted_at` in `~/.jinn-client/swe-rebench-v2/generator-state.json`. It does not appear to account for already-posted tasks that still have claim capacity, active claims, pending evaluations, or already-submitted-but-not-yet-counted verdicts when deciding whether another posting is needed.

## Why This Matters

This can produce two opposite failure modes:

[...]

PR #494, opened 2026-05-22
feat: add creator batch posting infrastructure

ritsukai:
## Summary
- post all collected creator candidates in a tick instead of returning after the first successful post
- fault-isolate candidate failures so a batch tail can still post
- add opt-in generated bucket key overrides while preserving default once-per-bucket dedupe

## Stack
- Builds on #488 docs branch.
- PR2 will stack SWE-rebench-v2 policy/config/UI on top of this branch.

## Tests
- cd client && yarn vitest run test/daemon/creator.test.ts test/tasks/sources.test.ts
- cd client && yarn typecheck

Refs #492
Refs #487


ritsuKai2000 (approved):
Approved on the Captain's explicit direction for this batch integration. Review basis: a dedicated review pass was run on this base PR and returned verdict APPROVED. This PR is the middle of the #488 -> #494 -> #495 stack. Recorded from the secondary maintainer account because GitHub blocks PR-author self-approval; this is not a fresh human diff-read.

Issue #302, opened 2026-05-19
investigate: codex-code-learner adapter session-start hook fails on Op A (missing path)


ritsuKai2000:
## Symptom

On Op A (running from sibling worktree `~/harbor/jinn-mono/cargo/client/`, harness `codex-code-learner`), every task claim's solver path dies at the session-start hook:

\`\`\`
[harness-engine] tick: process(<request>) failed: codex-code adapter: session-start hook failed:
  bash: /Users/adrianobradley/harbor/jinn-mono/cargo/client/plugins/learner/hooks/session-start: No such file or directory
\`\`\`

The hook script is resolved at the worktree-relative path `plugins/learner/hooks/session-start` but does not exist there. The daemon's solver attempts complete CLAIMED → WAITING → PRE_SNAPSHOT → RUNNING, then fail at the hook invocation. No solution gets submitted. The on-chain claim slot is consumed; nothing else useful happens.

## Context

- Op A is a long-running Base Sepolia operator (~22h uptime as of 2026-05-19). Generator + evaluator paths work correctly; only the **solver** path on this harness is broken.
- Same pattern was observed across multiple un-related task request IDs in Op A's recent log window — it's not task-specific.
- Op B (running from `~/life's-work/jinn-mono/client/`, harness `hermes-agent`) does NOT have this issue. The session-start hook may be a `codex-code-learner`-specific path.
- The hook path is hardcoded relative to the worktree where the daemon was launched (`process.argv[1]`-relative or similar?). If Op A was launched from `~/harbor/jinn-mono/cargo/client/dist/bin/jinn.js` and the hook lookup walks `<install>/plugins/learner/hooks/session-start`, the hook script is missing from that install.

## What to investigate

Don't prescribe a solution yet. Read the code and answer:

1. **Where is the session-start hook actually loaded from?** Trace `codex-code-learner` adapter's hook resolution path. Is it `<plugin-root>/hooks/session-start`? `<solver-plugin>/hooks/session-start`? `<client-install>/plugins/learner/hooks/session-start`?
2. **Should the hook script ship with the dist?** If yes, why isn't it being copied to `dist/`? Check the `client/package.json` build steps and the `client/plugins/learner/` directory layout. If no, where should it come from, and is there a missing setup step?
[...]

PR #501, opened 2026-05-22
chore(learner): ship plugin tree inside dist/ so it version-locks with compiled code

### LINKED ISSUE #417: chore: Ship the learner plugin tree inside dist/ so it version-locks with compiled code
## Context

Out of the #302 investigation (codex-code-learner session-start hook fails — missing path).

`resolvePluginRoot()` (`client/src/harnesses/impls/learner/plugin-path.ts`) resolves the learner plugin by walking up 4 directories from the compiled file (`dist/harnesses/impls/learner/plugin-path.js`) to the **package root**, then appending `plugins/learner`. So `dist/` is not self-contained — at runtime it depends on a `plugins/` tree that lives as a *sibling* of `dist/`.

The build copies `templates/` → `dist/templates/` and `scripts/*` → `dist/scripts/`, but does **not** copy `plugins/`. The published npm package only works because `package.json` `files` ships both `/dist/` and `/plugins/`. In a source checkout the compiled `dist/` and the `plugins/` tree can drift out of sync — which is exactly the #302 failure: a stale checkout whose `dist/` expected `plugins/learner/hooks/session-start` that its older `plugins/learner/` tree did not yet contain.

## Impact

The running compiled code and a separate runtime asset tree must stay version-locked across a directory boundary, with nothing enforcing it. A stale or partial build silently breaks the codex-code / claude-code learner harnesses at session-start (observed: ~19 consecutive task failures on a real operator).

Suggested priority: P2.

## Acceptance criteria

- The learner plugin runtime assets ship inside `dist/` (e.g. `cp -R plugins dist/plugins` in the build step, the same pattern as `templates/`).
- `resolvePluginRoot()` resolves the plugin tree relative to `dist/`, so it is always version-locked with the compiled code — for published installs and source checkouts alike.

## Run-mode

`chore` — build / packaging.

ritsuKai2000:
## Summary

- The `build` script now copies `plugins/learner` → `dist/plugins/learner` (same pattern as `templates/` and `scripts/`) so the compiled daemon and its runtime plugin assets are always co-located in `dist/` — for published npm installs and source checkouts alike.
- `resolvePluginRoot()` now detects whether it is running from a compiled `dist/` tree (3-level walk to `dist/`, then `plugins/learner`) or a source-checkout via tsx (4-level walk to the package root, then `plugins/learner`). The detection is a `/dist` path-suffix check on the resolved 3-level-up directory.
- The existing `plugin-path.test.ts` still passes (source-checkout branch). A new test verifies the `dist/plugins/learner` layout is present and correct after `yarn build`.

Fixes the root cause of the #302 drift failure: a stale source checkout whose `dist/` expected `plugins/learner/hooks/session-start` that its older `plugins/learner/` tree did not yet contain — because previously the build did not copy plugins into dist/ at all.

## Test plan

- [x] `yarn typecheck` — zero errors
- [x] `npx vitest run test/harnesses/impls/learner/plugin-path.test.ts` — 3 tests pass (source path + dist layout check after build)
- [x] `yarn build` — `dist/plugins/learner/` created alongside compiled `dist/plugins/index.js` (TypeScript output preserved)
- [x] `npx vitest run` (full suite) — 482 passed, 5 skipped; one pre-existing intermittent `transcript-watcher.test.ts` timeout unrelated to this change (passes when run in isolation)
- [ ] Stage 7 skipped — change is build/packaging only, no operator-visible surface touched

Closes #417


ritsukai (approved):
Approved on the Captain's explicit direction for this batch integration. Review basis: the implement-issue pipeline ran end-to-end on this issue — Stage 5 independent code review returned APPROVED and Stage 6 /security-review returned no findings. Closes #417. Recorded from the secondary maintainer account because GitHub blocks PR-author self-approval; this is not a fresh human diff-read.

PR #503, opened 2026-05-22
fix(creator): drop un-bindable legacy tasks[] config entries at startup

### LINKED ISSUE #415: fix: Creator retries un-bindable legacy tasks[] config entries forever (recurring TICK ERROR)
## Context

A legacy top-level `tasks: [...]` array in `~/.jinn-client/config.json` is wired into the creator loop via `StaticConfiguredTaskSource(config.tasks)` (`client/src/main.ts:2278`). Under the spec-§14 BINDING rule, every on-chain Task must carry a `solverNetManifestCid`; a `tasks[]` entry has no manifest binding, so `signTaskDocument` (`client/src/adapters/mech/adapter.ts:474`) throws a `PermanentError` before the task is ever posted.

The creator records a 30-minute backoff and re-throws; `run()` catches it and emits a `tick_error`, which renders as a red **TICK ERROR** row on the Dashboard activity feed. After 30 minutes the entry is retried, fails identically, and emits another — indefinitely.

## Impact

Any operator whose config still carries a top-level `tasks[]` array (legacy / hand-authored — current onboarding does not generate one) gets a permanent red error on their Dashboard for a condition that can never self-resolve. The activity row shows `—` for its detail columns, so a permanent config error is visually indistinguishable from a transient RPC blip without reading daemon logs.

Observed on a real fleet config during the v0.1.6 wave-2 smoke pass (a stale `health-check` entry). The §14 guard itself is correct — the problem is that the daemon keeps a permanently-un-postable entry as a live candidate.

Suggested priority: P3.

## Acceptance criteria

- A `tasks[]` entry that can never be bound to a SolverNet manifest is detected at config load and skipped with a single one-time warning — not retried every 30 minutes.
- Consider surfacing the `tick_error` `detail` / source on the Dashboard activity row so a permanent config error is distinguishable from a transient one.

## Run-mode

`fix` — regression test first.

ritsuKai2000:
## Summary

- Adds `filterBindableTasks` to `client/src/tasks/sources.ts`: a pure filter function that drops any `tasks[]` config entry missing `solverNetManifestCid`, emitting a one-time `console.warn` per dropped entry with a clear remediation message.
- Calls `filterBindableTasks(config.tasks)` in `main.ts` before passing the tasks array to `StaticConfiguredTaskSource`, so un-bindable entries are removed at daemon startup rather than entering the creator loop.
- Previously, such entries would throw `PermanentError` in `signTaskDocument` on every attempt, triggering a 30-minute backoff that resets forever — causing a permanent red TICK ERROR row on the Dashboard for a condition that can never self-resolve.

## Test plan

- [ ] `cd client && yarn test` — all 4066 tests pass (including 4 new regression tests in `test/tasks/sources.test.ts > filterBindableTasks — un-bindable entry guard (issue #415)`)
- [ ] `yarn typecheck` — zero errors
- [ ] Regression tests cover: single un-bindable entry dropped with warning; bindable entry passes through untouched; mixed list drops only un-bindable; warning fires once per entry (not per tick)

Stage 7 (jinn-app test) skipped — change does not touch any operator-visible surface (no dashboard SPA, no API routes, no bootstrap flows).

Closes #415


ritsukai (approved):
Approved on the Captain's explicit direction for this batch integration. Review basis: the implement-issue pipeline ran end-to-end on this issue — Stage 5 independent code review returned APPROVED and Stage 6 /security-review returned no findings. Closes #415. Recorded from the secondary maintainer account because GitHub blocks PR-author self-approval; this is not a fresh human diff-read.

PR #509, opened 2026-05-22
[#220] Default testnet JINN claim loop on

ritsuKai2000:
## Summary

- Defaults testnet operators into the JINN claim loop now that the L2 emitter and standing relayer are live.
- Keeps the default submission mode as `emit-only`, so operators only emit L2 `ClaimTicket`s; the relayer handles Sepolia minting.
- Preserves explicit opt-out via config/env and keeps mainnet default off.

## Verification

- `yarn vitest run test/config.test.ts test/daemon/jinn-claim-loop.test.ts`
- `yarn typecheck`

Refs #220


ritsukai (approved):
Approved on the Captain's explicit direction. Review basis: inline code review of the full 24-line diff — a config-default flip that enables the cross-chain JINN claim loop for testnet only (`loadConfig` sets it when unset; the Zod schema default stays `false` so mainnet/unknown networks are unaffected). The loop stays in `emit-only` mode — no cross-chain submission — so this is a safe progressive-rollout default. The `=== undefined` guard correctly preserves an explicit operator opt-out; three test cases cover testnet-on / explicit-opt-out / mainnet-off; CI is green. Recorded from the secondary maintainer account because GitHub blocks PR-author self-approval; this is not a fresh human diff-read.

PR #522, opened 2026-05-22
fix(earning): make computeRequiredMasterEth the single source of truth for the cold-start gate

ritsuKai2000:
### The gap — wiring-seam drift

`computeRequiredMasterEth()` in `client/src/earning/bootstrap.ts` is documented in its JSDoc as the single source of truth for the master-ETH cold-start funding gate. But it was **dead code** — no caller invoked it. Two live call sites each re-computed the gate inline with logic that diverged from the helper and from each other:

- **`bootstrap.ts` — `FleetBootstrapper.ensureStage1And2`**: computed `standardFleetAlreadyComplete` inline with a `!pendingSetupMigration` guard, a `services.length >= targetServices` floor, and a `services.every(isOperationalServiceStep)` check.
- **`funding-plan.ts` — `planFleetFunding`**: computed `standardFleetAlreadyComplete` inline but **omitted the `pendingSetupMigration` guard entirely**, and imported only `STANDARD_MASTER_BOOTSTRAP_MULTIPLIER`, never the helper.

Result: a migration-wiped fleet computed a different required-ETH gate in the funding-plan view than in the bootstrapper gate. This is the **u34i cross-module invariant-drift hazard** — cited verbatim in this same code's own comments — being violated.

### The fix

- **Rewrote `computeRequiredMasterEth` as the actual single source of truth.** One reconciled `standardFleetAlreadyComplete` definition (the strictest — the FleetBootstrapper variant: `services.length >= targetServices` AND `every(isOperationalServiceStep)` AND `!pendingSetupMigration`). Added a `preStage1` parameter so the helper covers both the Stage 1 gate (`stage1MinMasterEth`) and the Stage 2 gate, plus the self-bond flat budget. Hoisted `SELF_BOND_ETH_PER_SERVICE` to a module-level export.
- **Replaced the inline computation in `ensureStage1And2`** with a call to the helper (`preStage1: false` — this gate is only reached after `ensureStage1` succeeds).
- **Replaced the inline computation in `funding-plan.ts`** with a call to the helper, importing it from `./bootstrap.js`, and passing `pendingSetupMigration` (computed via `detectDeprecatedTestnetSetup`, a pure function — safe in the read-only path) so the funding-plan view agrees with the bootstrapper for migration-wiped fleets.
- **Corrected the helper JSDoc** so the "single source of truth / imported by funding-plan.ts" claim is now true.

The **only behavioural change** is the bug fix: the funding-plan view now also applies the `pendingSetupMigration` guard. All non-migration behaviour is preserved (verified — the full earning suite, including the pre-existing funding-plan and staged-bootstrap tests, is green).

### Regression test (`fix` shape)

New file `client/test/earning/master-eth-gate-seam.test.ts` — a cross-module wiring-seam test that asserts the bootstrapper gate and the funding-plan gate return identical `requiredMasterEth` for **(a)** a migration-wiped fleet fixture and **(b)** a normal cold fleet fixture.

Verified by stashing the `src/` changes: the new test **fails 3/3 against pre-fix code** (e.g. `(a)` — pre-fix funding-plan returned `0n`/satisfied for a migration-wiped fleet while the bootstrapper required funding) and **passes 3/3 after the fix**. The pre-existing `computeRequiredMasterEth` unit-test block in `bootstrap.test.ts` was rewritten to assert the new (correct) semantics, since the old block encoded the dead-code helper's divergent behaviour.

### Verification

```
yarn typecheck   → 0 errors
yarn test test/earning/   → Test Files 23 passed (23), Tests 200 passed (200)
```

Did not run the full e2e suite (out of scope).


ritsuKai2000:
Superseded by #529 — folded into the single `release/v2026.05.25` integration branch (release-readiness driver run). The fix is preserved as a stacked commit there; closing this standalone PR so the release candidate is one testable branch, not a scatter.

PR #540, opened 2026-05-23
fix(api): show owned launched SolverNet manifests without requiring a subgraph

### LINKED ISSUE #114: Show owned launched SolverNet manifests without requiring a subgraph
## Problem

The launched SolverNet dashboard fetches manifest details through `GET /v1/solvernets/registry/:cid`. In local or first-launch setups where no registry client / subgraph is configured, that route cannot resolve the manifest and the dashboard can show placeholder content such as an unnamed SolverNet or missing prices.

This is especially confusing for owned SolverNets: the daemon just launched the SolverNet and should already have enough local information to display the manifest that the launcher authored.

## Files to inspect

- `client/src/api/solvernets-endpoints.ts`
- `client/src/solvernets/store.ts`
- `client/src/solvernets/launch-state-machine.ts`
- `client/src/dashboard/spa/src/pages/LauncherLaunched.tsx`

Current route of interest:

```ts
GET /v1/solvernets/registry/:cid
```

The route currently relies on the registry client before returning a manifest. Owned launched records are available through the local launched-record store.

## Expected behavior

For SolverNets launched by the current daemon, the launched dashboard should show the manifest name, description, pricing, and contract metadata without requiring a subgraph or external registry read.

External discovery can still rely on the registry path. This issue is about the owned-record fallback.

## Suggested approach

Either of these shapes is acceptable:

1. Teach `GET /v1/solvernets/registry/:cid` to first check local launched records for an owned record with the requested CID and return the cached manifest body when available.
2. Or teach `GET /v1/solvernets/launched/:id` to return the launched record plus the manifest body inline, and update the SPA to use that for owned records.

If the launched record does not currently persist a `manifestPath` or manifest body, populate a local cache during launch recording.

## Acceptance criteria

- A just-launched owned SolverNet dashboard displays manifest details when no subgraph / registry client is configured.
- The fix does not weaken registry verification for non-owned SolverNets.
- Add an API test covering the no-subgraph owned-record path.
- Add or update a SPA test so the launched dashboard does not render unnamed / missing-price placeholders for an owned launched record.

ritsuKai2000:
## Summary

`GET /v1/solvernets/registry/:cid` now serves owned, locally-cached SolverNet manifests with a synthesised `launched` lifecycle when the on-chain `metadataBlockNumber` isn't anchored yet (records still in `launching` or `failed` status). This lets the post-launch dashboard render the manifest name, description, and prices immediately — instead of placeholders — on a freshly-launched owned SolverNet, even when no registry / subgraph client is configured or when the configured client is wired but unreachable (Anvil / local dev).

**Approach.** Broaden the existing owned-record fallback at `client/src/api/solvernets-endpoints.ts`:

- `localLifecycleForRecord` now returns a synthetic `launched` lifecycle with `sourceBlock: 0` for owned records lacking `metadataBlockNumber` (previously returned `null`, which let the route fall through to the registry path and produce 503 / 404). The record's real `status` (`launching` / `failed`) is still surfaced separately via `/v1/solvernets/launched/:id` and rendered as a status pill — the synthetic lifecycle is purely for the manifest-render path.
- CID shape validation now runs **before** the owned-cache lookup and the no-registry 503 guard, so malformed CIDs consistently return 400 regardless of registry/cache state.
- The existing hash-mismatch + `solverNetId`-mismatch guards in `tryGetOwnedCachedManifest` are unchanged — non-owned manifests can never be served from the owned cache, and a tampered on-disk file cannot be served as a legitimate manifest.

No changes to `launch-state-machine.ts`, `solvernets/store.ts`, or the SPA source — the on-disk manifest cache is already populated at the `recording` phase via `writeManifestCache`, and the SPA's existing `manifestQuery` automatically picks up the new API contract.

## Test plan

- [x] `client/test/api/solvernets-endpoints.test.ts` — new regression tests:
  - serves an owned cached manifest for `launching` / `failed` records before lifecycle is anchored (parametrised)
  - falls back to owned cached manifest when the registry client throws (subgraph unreachable)
  - rejects malformed CIDs with 400 in no-registry mode
  - existing hash-mismatch test continues to pass (verification semantic preserved)
- [x] `client/src/dashboard/spa/src/pages/LauncherLaunched.test.tsx` — new SPA test:
  - mounts an owned `launching` record with a synthesised `launched` lifecycle
  - asserts manifest name renders ("Polymarket") instead of falling back to `record.solverNetId`
  - asserts `launcher-launched-spend-solution-price` and `launcher-launched-spend-verdict-price` render real values (not the `—` placeholder)
- [x] `yarn run typecheck` — clean
- [x] `yarn test` (full vitest suite) — 484 files, 4090 tests passing, 8 skipped (no regressions)

Closes #114


ritsukai (approved):
LGTM — rebased onto current next; local verification passed (typecheck/test). Merging despite stale GitHub PR-sync state.

PR #556, opened 2026-05-24
fix(test): stub RPC for funding-sequence e2e instead of dialling Tenderly

oaksprout:
## Summary

- `funding-sequence.e2e.test.ts` set `BASE_RPC_URL=http://127.0.0.1:65000` to dodge a real RPC, but [`client/src/config.ts`](client/src/config.ts) only honours `JINN_RPC_URL`/`BASE_SEPOLIA_RPC_URL` on testnet — `BASE_RPC_URL` silently fell through to the default Tenderly gateway. Once that gateway started rate-limiting CI on 2026-05-23, the daemon's pre-bind `checkRpcNetwork` failed and the process exited before [`startApiServer`](client/src/main.ts:984) could bind, leaving the test polling a dead port for 15s. Every PR's funding-sequence run has been red since.
- Replaced with a loopback `http.createServer` in `beforeAll` that answers `eth_chainId` with Base Sepolia (`0x14a34`), falls back to `0x0` for unknown methods so viem's `BigInt` parser doesn't choke during the early bootstrap RPC reads, and passes the URL through `JINN_RPC_URL` — the env var `config.ts` actually honours.
- Relay daemon `stdout`/`stderr` to the test log; the previous version only fed the chunks into a regex looking for the handshake URL, so a failed startup left no diagnostic.

## Test plan

- [x] `yarn e2e:funding-sequence` passes locally (was failing with `daemon never came up on test port` against `next` HEAD).
- [ ] CI on this PR (and any other open PR rebased on it) goes green for the funding-sequence check.


ritsuKai2000:
Closing as superseded — `next` already has an inline `createServer` JSON-RPC stub in `client/test/dashboard/funding-sequence.e2e.test.ts` serving the same intent. Branch is ~17k lines stale; no net change to integrate.

PR #614, opened 2026-05-25
feat(rpc): multi-RPC fallback chain across daemon / relayer / indexer / CI

### LINKED ISSUE #592: Multi-RPC fallback chain across daemon / relayer / indexer / CI via viem fallback() transport
**Context.** This session (2026-05-25) accumulated multiple distinct RPC-reliability failures across components that all share one shape: a single RPC provider hit a quota / range cap / rate limit and the component went silent. Accumulated evidence:

| Surface | Failure | Resolution this session |
|---|---|---|
| Operator daemon + relayer | Shared Tenderly free-tier key (`15b0C3dTipsUOekY0Fmuia`) exhausted its monthly quota → both went silent | Operator-local config swap to `base-sepolia.publicnode.com` |
| `sepolia.base.org` (public) | 2,000-block `eth_getLogs` cap broke the mech canonical scan (which needs 9999-block chunks); aggressive IP rate-limiting | Switched to publicnode (50,000-block cap) |
| Ponder indexer (Railway) | 2026-05-20 cascade: Tenderly key quota → indexer crashed → silent fallthrough; 2026-05-25 [#567](https://github.com/Jinn-Network/mono/issues/567): `TaskCreated` handler silently stopped | Operator-side `fallbackToOnchain=false` default (`638228f1`); indexer rotation; pending [#567](https://github.com/Jinn-Network/mono/issues/567) fix |
| CI Tier 1 T1.1 (`bootstrap-fresh-anvil`) | `getTransactionReceipt` timeout against `BASE_SEPOLIA_RPC_URL` secret in the npm-publish workflow | Re-ran; flaked-passed |

Every one of these is a **single point of failure** plus **silent degradation**. The standard mitigation in viem (which we already use throughout the daemon, relayer, and Ponder via dependency) is the `fallback()` transport: an ordered list of RPCs that try primary first, transparently fall through to the next on error / timeout / quota / rate-limit.

The team has explicitly chosen NOT to pay materially for operator-side RPC (operator-owned concern) but does want stability for project-managed infra (the indexer). This design balances both:

- **Operator default**: zero-config, no key required — two free public RPCs in a fallback chain (`publicnode` → `sepolia.base.org`)
- **Operator opt-in**: paste their own paid Alchemy/Infura/Tenderly URL as primary; existing free chain remains as backup
- **Project-managed indexer**: paid Alchemy free-tier (300M CU/mo, sufficient for testnet) as primary + 3 free backups → effectively-zero-cost while having SLA + failover

**Impact.** Today every operator hits the RPC SPOF lottery on their own. Indexer outages cascade across the entire network (one Tenderly key behind all of it). Even with the `fallbackToOnchain` default-off fix, operators see explicit `DiscoveryUnavailableError` but have no in-place recovery — they wait for the project to fix the indexer. With this design, both halves of the network get failover automatically, no operator action required.

**Acceptance criteria.**

- [ ] **Daemon `config.ts`**: `rpcUrl` accepts either a single string (current shape, backward-compatible) OR an array of strings. When an array, the daemon constructs a viem `fallback()` transport with the entries in order. Same for `ethereumRpcUrl`.
- [ ] **Default operator config**: `rpcUrl` defaults to `["https://base-sepolia.publicnode.com", "https://sepolia.base.org"]` on testnet (not the current single-entry value). Operators with custom keys can prepend or replace.
- [ ] **Relayer wrapper** (`packages/claim-relayer/`): `JINN_CLAIM_RELAYER_L2_RPC_URL` and `JINN_CLAIM_RELAYER_L1_RPC_URL` accept comma-separated lists; parsed into a viem `fallback()` chain at startup.
- [ ] **Ponder indexer config** (`packages/indexer/` or wherever Ponder lives): network's `transport` is a viem `fallback()` of up to 4 RPCs sourced from env (`INDEXER_RPC_PRIMARY_BASE_SEPOLIA`, `_SECONDARY`, etc.). Document the recommended ordering (paid → free → free → free).
- [ ] **CI `BASE_SEPOLIA_RPC_URL` secret**: workflow scripts (`run-tier-1.ts`, `run-tier-2.ts`, `run-tier-3.ts`, `daemon-harness-cycle.ts`) accept comma-separated and pass through to viem fallback. Solves the T1.1 receipt-timeout class of CI flake.
- [ ] **Operator docs (CLAUDE.md / client/README.md / OPERATOR-APP-SPEC.md)** explain: (a) the default fallback chain is zero-config + free-tier, (b) how to add a custom paid primary, (c) what failure modes are now automatic (vs the manual rpcUrl swap we did this session).
- [ ] **Backward compatibility**: existing operators with single-string `rpcUrl` in their config don't have to change anything; the loader normalizes to a single-element array internally.
- [ ] **Boot-time logging**: daemon emits `[rpc] L2 transport: fallback chain (3 providers) — primary=base-sepolia.publicnode.com` so the operator can see what's loaded without grep-ing config.
- [ ] **Unit + integration tests**: cover the fallback path (primary errors → secondary serves; primary 429s → secondary serves; all error → throws clear `AllRpcsFailedError`).

**Design notes (not blocking the acceptance criteria above).**

- **Why viem `fallback()` not a custom helper**: viem's implementation handles all the edge cases (retries, exponential backoff between attempts, error normalization, racing on slow primaries via `rank: true` if we want it). Already a transitive dependency. No new code surface.
- **Why fallback over round-robin**: round-robin spreads load but multiplies the failure surface area (one slow RPC slows every Nth request). Fallback keeps latency at primary-RPC's level until primary errors. Better p99.
- **Why not a single shared paid project RPC for operators**: defeats the GROWTH model + headless brand posture + introduces a real centralization risk. Project pays for the indexer only.
- **Provider ordering recommendation**: paid > free-public > free-public-backup. Don't put `sepolia.base.org` first — its 2k-block range cap means the daemon's canonical scan eats CPU re-chunking even when other providers would handle it in one call.
- **Per-call quota awareness**: viem `fallback()` doesn't know about monthly quotas, only per-call errors. If primary is silently quota-exhausted (returns 429), fallback works. If primary returns wrong/empty data without error (the Ponder-handler bug pattern of [#567](https://github.com/Jinn-Network/mono/issues/567)), fallback doesn't help — that's a separate completeness-monitoring concern.

**Files/components.** `client/src/config.ts` (rpcUrl + ethereumRpcUrl typing), `client/src/adapters/mech/adapter.ts` (transport construction), `client/src/main.ts` (boot-time logging), `packages/claim-relayer/src/index.ts` (wrapper + URL parsing), `packages/indexer/ponder.config.ts` (estimated path — Ponder transport config), `.github/workflows/npm-publish.yml` + `client/scripts/release/run-tier-*.ts` (CI RPC handling), `CLAUDE.md` / `client/README.md` / `client/OPERATOR-APP-SPEC.md` (docs). Related issues: [#567](https://github.com/Jinn-Network/mono/issues/567) (indexer handler completeness — orthogonal but mentioned here), [#576](https://github.com/Jinn-Network/mono/issues/576) (per-operator claim quota — RPC reliability affects the rate at which operators exhaust slots).

---

## Update 2026-05-25 — corrected provider ordering after fetching actual provider docs

Original draft suggested Alchemy as primary for the indexer. **That's wrong** for our chains. From [Alchemy's eth_getLogs reference](https://www.alchemy.com/docs/reference/eth-getlogs):

| Alchemy tier | `eth_getLogs` block range — Ethereum | `eth_getLogs` block range — "All Other Chains" (incl. Base Sepolia) |
|---|---|---|
| Free | **10 blocks** | 10 blocks |
| Pay-As-You-Go | unlimited | **10,000 blocks** |
| Enterprise | unlimited | 10,000 blocks |

`publicnode` free-tier already gives **50,000 blocks** — wider than Alchemy's paid testnet tier. Alchemy is genuinely the wrong primary for this Ponder indexer at our scale.

**Corrected per-component fallback chains (all $0/month):**

```ts
// Ponder indexer (project-managed)
//   Primary: Envio HyperSync — purpose-built for event indexing,
//   1000-2000× faster than RPC, free tier requires an account
//   (https://docs.envio.dev/docs/HyperSync/hypersync-supported-networks).
//   Free-tier ceiling not fully published but very likely fits a single
//   testnet indexer; if exceeded, falls through to publicnode.
fallback([
  http(process.env.PONDER_RPC_URL_84532_PRIMARY),   // https://base-sepolia.rpc.hypersync.xyz
  http('https://base-sepolia.publicnode.com'),       // 50k-block cap, no quota
  http('https://sepolia.base.org'),                   // 2k-block cap, last resort
])

// Operator daemon + relayer (operator-managed; zero-config $0 default)
fallback([
  http('https://base-sepolia.publicnode.com'),
  http('https://sepolia.base.org'),
])

// Ethereum Sepolia L1 (relayer + L1↔L2 loop)
fallback([
  http('https://ethereum-sepolia-rpc.publicnode.com'),
  // no widely-used free public ETH Sepolia secondary worth listing as default;
  // operator opts into Infura free as backup if they want
])
```

**What changes vs original draft:**
- Indexer primary moves from `paid Alchemy free-tier` to **Envio HyperSync** (purpose-built for events).
- Indexer chain depth reduced from 4 → 3 providers (Alchemy removed; Infura optional).
- All three indexer providers are free.
- Operator default unchanged: `publicnode → sepolia.base.org` (two free public RPCs, no account needed).

**Ponder's `ethGetLogsBlockRange` knob** is the per-chain cap to set so the slowest fallback (sepolia.base.org at 2k) doesn't blow up under chunking. Ponder's [config docs](https://ponder.sh/docs/api-reference/ponder/config) document this. Set `ethGetLogsBlockRange: 2000` for Base Sepolia so the same code path works regardless of which provider in the chain actually serves the request.

**Sources** (all fetched 2026-05-25):
- [Alchemy eth_getLogs](https://www.alchemy.com/docs/reference/eth-getlogs)
- [Envio HyperSync supported networks](https://docs.envio.dev/docs/HyperSync/hypersync-supported-networks) (Base Sepolia + Sepolia confirmed)
- [Envio HyperSync overview](https://docs.envio.dev/docs/HyperSync/overview) (1000-2000× speed claim)
- [Ponder config](https://ponder.sh/docs/api-reference/ponder/config) (viem `fallback` first-class; `ethGetLogsBlockRange` knob)

---

## Update 2026-05-25 (later) — incorporate existing Tenderly RPCs

The team's `client/.env` already has two Tenderly gateway URLs (operator-local secrets, not in repo):

```
BASE_SEPOLIA_RPC_URL=https://base-sepolia.gateway.tenderly.co/<KEY>/<KEY>   # Base Sepolia (chain 84532) — has current consumers (indexer + daemon)
BASE_RPC_URL=https://base.gateway.tenderly.co/<KEY>/<KEY>                   # Base mainnet  (chain 8453) — no current indexer/daemon consumer (per ponder.config.ts "intentionally NOT indexed yet"); only used today as the Anvil fork source for CI Tier 1 T1.1
```

Only the Base Sepolia URL feeds into the indexer/daemon/relayer fallback chains in this design. The Base mainnet URL stays where it already is — as the `BASE_RPC_URL` secret in the npm-publish + tier-1 workflows for Anvil forking. If/when Phase 2 begins indexing Base mainnet (`jinn-mono-280n.4`), THEN we add the mainnet chain to this design.

These add headroom to the fallback chain without changing the $0 baseline. Key positioning rule: viem `fallback()` is **strict ordered failover, not round-robin** — it only tries provider N+1 if provider N errors. So putting Tenderly LATER in the chain reserves its monthly quota for crisis mode (when publicnode itself misbehaves), instead of burning it on every steady-state call.

**Revised per-component chains (all $0/month for the default path; Tenderly only burns quota under failover):**

```ts
// Ponder indexer (project-managed)
//   Envio HyperSync primary keeps the indexer fast.
//   publicnode handles steady-state for free.
//   Tenderly (Base Sepolia) is the THIRD line of defense — engaged only
//   when both Envio and publicnode are erroring at the same time.
//   sepolia.base.org is last resort.
fallback([
  http(process.env.PONDER_RPC_URL_84532_PRIMARY),   // https://base-sepolia.rpc.hypersync.xyz (Envio)
  http('https://base-sepolia.publicnode.com'),       // 50k-block cap, no quota
  http(process.env.PONDER_RPC_URL_84532_TENDERLY),   // BASE_SEPOLIA_RPC_URL from indexer's secrets
  http('https://sepolia.base.org'),                   // 2k-block cap, last resort
])

// Operator daemon + relayer (operator-managed; zero-config $0 default)
//   Operators don't have Tenderly URLs by default — they're operator-local
//   secrets the team has set up but shouldn't be expected of new operators.
//   If an operator HAS a Tenderly key, they can prepend it via config.
//   Default chain stays free + public:
fallback([
  http('https://base-sepolia.publicnode.com'),
  http('https://sepolia.base.org'),
])

// Operator with a Tenderly URL (opt-in): they configure
//   rpcUrl: ["https://base-sepolia.publicnode.com",
//           "https://base-sepolia.gateway.tenderly.co/<their-key>/<their-key>",
//           "https://sepolia.base.org"]
// publicnode still primary so their Tenderly quota survives.

// Ethereum Sepolia L1 (relayer + L1↔L2 loop) — Tenderly not currently
// configured for L1 in client/.env; keep public-only:
fallback([
  http('https://ethereum-sepolia-rpc.publicnode.com'),
  // operator adds Infura free as backup if they want
])
```

**Why Tenderly goes in position 3 (after publicnode), not position 1:**
- The 2026-05-20 cascade was caused by Tenderly's monthly quota being silently exhausted while it was *primary*. Burning publicnode's IP-rate-limit first means Tenderly stays available longer for genuine emergencies.
- publicnode has no monthly quota cliff, so it's the safer baseline.
- If publicnode itself goes down (or sustainedly throttles), Tenderly is there.

**Acceptance criteria addendum** (in addition to those above):
- [ ] Indexer deployment accepts `PONDER_RPC_URL_84532_TENDERLY` (and `_8453_TENDERLY` for the future mainnet path) as additional fallback entries, slotted at position 3 in their respective chains.
- [ ] Operator config schema doesn't require Tenderly env vars (operator default stays zero-config-zero-account); operators who paste a Tenderly URL into their `rpcUrl` array get it slotted in position 2 (their own paid backup, ahead of public free).
- [ ] Docs section "Adding a paid RPC backup" walks operators through the prepend pattern with Tenderly as the worked example.

---

## Update 2026-05-25 (live RPC probe) — implementation notes

Ran `eth_blockNumber` + `eth_getLogs` (50k-block range) against every provider in the chain. Two findings the implementer should bake in:

### Envio HyperRPC tail lag (~1-2 blocks)

HyperSync is **typically 1-2 blocks behind chain head** as observed in the probe. A naive `eth_getLogs` with `toBlock: "latest"` against Envio when the caller already knows chain head from a *different* provider will error:

```
{"error":{"message":"Invalid params: requested block 41976868 is greater than latest block 41976866"}}
```

Concrete probe result (2026-05-25): publicnode reported block 41,976,868; Envio reported 41,976,866; the 50k-block range with `toBlock = 41,976,868` (chain head per publicnode) failed against Envio with the message above.

**Implementer mitigation** — the indexer's tail-polling path must clamp `toBlock` to whichever provider is actually serving the request, not to a globally-cached "latest" derived from a different provider. Two patterns work:

1. **Per-provider block-head cache** — each provider in the fallback chain caches its own latest block number, and `eth_getLogs` for that provider clamps to its own cached head.
2. **Conservative offset** — when calling Envio for tail polling specifically, use `toBlock = head - 3` (or whatever margin) to stay safely behind Envio's tip. Backfill is unaffected (the `toBlock` for old ranges is always well past Envio's tip).

Ponder's built-in tail-following may already handle this — verify before adding the workaround. If it does, no change needed beyond noting it for whoever debugs a future "block ahead of latest" error.

### Tenderly key is alive today; still keep in slot 3

The probe shows `BASE_SEPOLIA_RPC_URL` (Tenderly key `15b0...iaa`) is responding to both `eth_blockNumber` and a 50k-block `eth_getLogs` — i.e., **it's not currently quota-exhausted**, contradicting our 2026-05-25-morning observation. Possible explanations: month-end quota reset, OR the morning's exhaustion was caused by sustained multi-loop burn (daemon + relayer + indexer all hitting it concurrently) and single calls slip under whatever rolling-window meter Tenderly uses.

**Implementer guidance:**
- Don't promote Tenderly to a higher chain position based on "it works today". It's a quota-bounded reserve, and we have evidence the quota CAN exhaust under realistic load. The slot-3 position (behind Envio + publicnode) is right.
- Add a small **boot-time probe** that calls `eth_blockNumber` against every provider in the chain at startup and logs the latency. If any provider returns 429/5xx at boot, log a warning so operators see it without grep-ing later.
- The Tenderly key's resilience pattern matters: heavy concurrent burn exhausts faster than the same total call count spread over time. Worth a load-profile aware throttle later (out of scope for #592).

### What the probe DID confirm cleanly

- ✅ Envio token works (HyperRPC returns 401 without it, 200 with it)
- ✅ publicnode handles 50k-block ranges as advertised (256 logs returned in test)
- ✅ Tenderly handles 50k-block ranges (286 logs returned)
- ✅ sepolia.base.org confirmed 2k-block cap (explicit `query exceeds max block range 2000` error) — last-resort positioning is correct
- ✅ Envio also works on Ethereum Sepolia L1 with the same token — useful if we add L1 indexing later

ritsukai:
## Summary

Adds a viem `fallback()`-based multi-RPC transport across the daemon, claim-relayer, Ponder indexer, and CI release scripts. Every component that previously hit a single RPC URL now accepts a 1-4 provider chain (string-or-array config; comma-separated env vars) and transparently fails over on network error, 429, or 5xx. Zero-config operator default on testnet is `["https://base-sepolia.publicnode.com", "https://sepolia.base.org"]` — operators with their own paid primary prepend it.

Closes #592.

### What changed

- **Daemon (`client/`):**
  - New `client/src/rpc/transport.ts` — `parseRpcUrls` / `buildFallbackTransport` / `AllRpcsFailedError` / `describeFallbackChain`. Wraps viem `fallback([...], { rank: false, retryCount: 0 })`; dedupes before applying the 4-slot cap; mirrors viem's own `shouldThrow` predicate (by numeric JSON-RPC code) so contract reverts propagate unchanged rather than being mis-wrapped as "all RPCs failed".
  - `client/src/config.ts` — RPC fields (`rpcUrl`, `archiveRpcUrl`, `l2ProofRpcUrl`, `ethereumRpcUrl`, `ethereumArchiveRpcUrl`) accept `string | string[]`. Loader normalises to a `rpcUrls: readonly string[]` sibling while keeping `rpcUrl: string` (the head URL) for display-site backward compat across ~30 call sites.
  - `client/src/earning/viem-clients.ts`, `client/src/adapters/mech/{safe,adapter,types}.ts` — factories accept the wider type and route through `buildFallbackTransport`.
  - `client/src/preflight/rpc-network.ts` — new `probeFallbackChain` issues a serial `eth_blockNumber` per slot, logs per-slot latency + warns on 429/5xx but never gates startup. `checkRpcNetwork` (chain-id check) is unchanged and still fail-loud on the head URL.
  - `client/src/main.ts` — wires the probe + emits the AC7 summary line for L1 and L2.
- **Relayer (`packages/claim-relayer/`):**
  - `src/transport.ts` (new) — intentional duplication of `parseRpcUrls` / `buildFallbackTransport` / `AllRpcsFailedError` / `isViemShouldThrowError` to keep the package self-contained.
  - `src/config.ts`, `src/index.ts` — accept comma-separated `JINN_CLAIM_RELAYER_L1_RPC_URL` / `_L2_RPC_URL`. `redactConfig` shows "fallback chain (N providers): primary host=…" instead of bare `[redacted]`.
- **Indexer (`packages/indexer/ponder.config.ts`):** each chain's `rpc` is now `fallback(urls.map(http), { rank: false })` sourced from `PONDER_RPC_URL_*` env vars (up to 4 slots). Added `ethGetLogsBlockRange: 2000` for `baseSepolia` so the slowest fallback (`sepolia.base.org`, 2k cap) doesn't blow up under chunking.
- **CI (`.github/workflows/npm-publish.yml`):** comment-only — documents that `BASE_SEPOLIA_RPC_URL` may now be a comma-separated list. The loader handles it transparently; no script changes needed.
- **Docs:** `CLAUDE.md`, `client/README.md`, `client/OPERATOR-APP-SPEC.md` — default chain, BYO-paid-primary recipe, automatic failure modes, distinguished explicitly from the unrelated `discovery.fallbackToOnchain` flag.

### Acceptance criteria (all from #592)

- [x] AC1 — `config.ts` accepts string or array; backward-compatible normalisation.
- [x] AC2 — testnet default chain `[publicnode, sepolia.base.org]`.
- [x] AC3 — relayer envs accept comma-separated.
- [x] AC4 — Ponder uses viem `fallback()` + `ethGetLogsBlockRange: 2000`.
- [x] AC5 — CI `BASE_SEPOLIA_RPC_URL` secret comma-separated flows through `loadConfig()`.
- [x] AC6 — docs cover default, BYO primary, automatic failure modes.
- [x] AC7 — boot log emits `[rpc] L2 transport: fallback chain (N providers) — primary=<host>`.
- [x] AC8 — unit tests cover primary error / 429 / all-fail → `AllRpcsFailedError`.
- [x] AC9 — boot probe `eth_blockNumber` per slot; latency + 429/5xx warnings.

### Out of scope (follow-up `chore`)

Ten low-volume `http(rpcUrl)` call sites not migrated in this PR (CLI doctor/tasks, gather-status, x402 facilitator, discovery onchain, corpus onchain-query, prediction-v0-auto, harnesses, manifest-validation). They are sequential one-shots that don't quota-storm and can ride the helper after this lands. Will file as a follow-up `chore` Issue referencing this PR.

## Test plan

- [x] `cd client && yarn typecheck` — clean.
- [x] `cd client && yarn test --run` — 4271 passed / 8 skipped.
- [x] `cd packages/claim-relayer && yarn test --run` — 39 passed.
- [x] `cd packages/indexer && yarn test --run` — 190 passed.
- [x] `cd client && yarn build` — green.
- [ ] Smoke run (deferred to canary): `JINN_RPC_URL='https://invalid-host.example, https://base-sepolia.publicnode.com' node dist/bin/jinn.js run` — daemon should log `[rpc] L2 invalid-host.example warn unreachable` then `[rpc] L2 base-sepolia.publicnode.com ok latency=...`, emit the summary line, and proceed to bootstrap via the secondary.

### Notes for reviewers

- The Stage 5 round-1 review caught a semantic gap where `AllRpcsFailedError` wrapped viem's `shouldThrow`-class errors (contract reverts misclassified as "all RPCs failed"). Round 2 caught a further refinement — viem's HTTP transport delivers JSON-RPC errors as `RpcRequestError` with a numeric code, so the predicate is now code-based (matching viem's own `shouldThrow`) plus the `ExecutionRevertedError.nodeMessage` regex plus a cause-chain name walk for wallet-provider stacks.
- Tenderly is intentionally NOT promoted to slot 1 even though it works today — the 2026-05-20 cascade showed its monthly quota can exhaust under load. publicnode in slot 1 burns the IP-rate-limit reserve first; Tenderly stays available for genuine crises.
- The security review noted that `AllRpcsFailedError.cause` retains the raw viem error (which embeds the URL with api-key). Currently contained — the daemon never serialises `.cause` to the operator app — but worth a follow-up hardening before any code path `JSON.stringify`s the cause.


ritsuKai2000 (approved):
Approved on Oak's authorization (/merge-batch 2026-05-26).

PR #477, opened 2026-05-21
feat(346): daily spend budget per credential (daemon-side)

ritsuKai2000:
## Summary

Daemon-side implementation of #346 — a per-credential daily spend budget for the Jinn client. When a credential's estimated USD spend for the UTC day reaches the operator's cap, the daemon stops claiming new tasks for that credential — a graceful pause *before* the operator's provider-side limit (the real hard ceiling) is hit.

- **Credential model** — `{provider}:{authMethod}` (e.g. `anthropic:api-key`), resolved from env-var presence (`client/src/spend/credential.ts`).
- **Accounting** — actual cost recorded at harness-run completion: Claude Code self-reports `total_cost_usd`; Codex tokens priced via `tokenlens`; hermes/unknown via heuristic. Stored as `cost_usd_micros` on `activity_events` rows — that table *is* the ledger; "spent today" is a `SUM` since UTC midnight. No reserve/settle (the cap is deliberately approximate).
- **Gate** — a pre-claim gate in `_runEngineWatcherLoop`, sibling to `gateClaimByReadiness`; skips claims for an over-budget credential and emits one `spend_cap_reached` event on the pause edge.
- **Config** — `spendCaps` map + `JINN_SPEND_CAP_USD` blanket env. Default unset = feature fully inert (zero overhead, no behaviour change).
- **Status** — `/v1/status` carries a per-credential `spend` block (cap / spent-today / paused / resetsAt).

Built TDD across 10 reviewed tasks. Design spec `docs/superpowers/specs/2026-05-21-per-credential-spend-budget-design.md` and plan `docs/superpowers/plans/2026-05-21-per-credential-spend-budget.md` (on branch `claude/focused-elion-5ee19d`).

## Scope notes

- **Operator-app UI deferred** — the Spend component is a canonical-doc (`OPERATOR-APP-SPEC.md`) amendment tracked as #453.
- **Cost-surface rewire reverted** — a planned 11th task (rewire #345's `harnessUsesPaidApiKey` to credential resolution) was reverted after the final whole-feature review: `harnessUsesPaidApiKey`/`decideCostSurface` are browser-SPA-only, and `resolveCredentialId` reads `process.env` which Vite stubs to `{}` in the browser — so a browser-side rewire delivers nothing. The real #331-UI fix needs daemon-exposed paid-ness + SPA consumption — re-homed to #474.
- Provider-API reconciliation descoped (design spec §12 / #346 D3).

## Test Plan

- [x] `yarn typecheck` — clean
[...]

PR #616, opened 2026-05-25
design(#576): record + reject DR-2026-05-25 — claim-as-commitment is the intended property

### LINKED ISSUE #576: Per-operator claim quota does not refund on attempt expiry; transient infra failures permanently consume slots
**Context.** The TaskCoordinator contract gates per-operator claims with:

```solidity
// contracts/src/tasks/TaskCoordinator.sol:331
if (claimsByTaskByOperator[taskId][operator] >= policy.maxClaimsPerOperator) {
    revert TCOperatorClaimLimitReached(taskId, operator);
}
// ...:346
claimsByTaskByOperator[taskId][operator]++;
```

The counter increments on EVERY claim and **has no decrement path** (verified by grep — the mapping is only ever incremented, never reduced). Consequence: when an operator's attempt fails for any reason — transient API errors, RPC outage, daemon restart mid-attempt, lease expiry without delivery — the slot is permanently consumed from that operator's quota for that task. The contract treats "claimed and successfully delivered" and "claimed and runtime exploded" identically.

Observed concretely 2026-05-25: this operator's daemon claimed tasks `0x6d4cc88a...` and `0x2b11ad47...`, then both attempts hit an OpenRouter monthly-cap 403 inside the hermes subprocess. Both task_runs went to FAILED. The two on-chain claim slots are now consumed and the operator cannot re-attempt those tasks even after the OpenRouter cap was resolved.

**Impact.** Operators are penalized for transient infrastructure failures (provider quota, RPC outage, daemon crash) by permanent slot loss. In a SolverNet where `maxClaimsPerOperator = N_target_successes - successful_so_far` (default), a single ops blip can cost the operator a whole task. Affects operator economics meaningfully and disincentivizes attempting tasks early in a SolverNet's lifetime when quotas are largest.

**Acceptance criteria.**
- [ ] A DR under `log/decisions/` proposes the contract semantics decision: should `claimsByTaskByOperator[taskId][operator]` decrement when an attempt's lease expires without a delivered solution (`expireAttempt`)? Trade-offs documented: refund prevents legitimate operators from being permanently locked out by transient failures; non-refund prevents grief patterns where an operator deliberately holds + drops slots to deny others.
- [ ] If the DR recommends refund-on-expiry: contract change spec'd with the modified `expireAttempt` flow, plus a migration plan for the deployed `0xdC9BCcEB...` instance (in-place upgrade, or replacement contract with read-through to old state).
- [ ] If the DR recommends keeping current semantics: operator-side mitigations are spec'd — e.g., readiness gates that block claims when the harness's provider credentials are in a known-failing state, so operators don't burn slots on a provider that just hit its cap.
- [ ] The DR addresses how this interacts with `requiredVerdicts` (a finalized task is closed before `maxClaims` is reached, so refunding empty slots on a finalized task may or may not be desirable depending on whether other operators could still benefit from filling them).

**Files/components.** `contracts/src/tasks/TaskCoordinator.sol:331,346` (the gate + increment), `contracts/src/tasks/TaskCoordinator.sol` (`expireAttempt` function), `client/src/harnesses/impls/hermes-agent/` (where credit-exhaustion failures originate today), related issues: [#545](https://github.com/Jinn-Network/mono/issues/545), [#562](https://github.com/Jinn-Network/mono/issues/562), [#569](https://github.com/Jinn-Network/mono/issues/569).

ritsukai:
Closes #576 as `wontfix`.

## What this PR lands
- `docs/superpowers/specs/2026-05-25-issue-576-claim-quota-refund-design-note.md` — Stage 1 brainstorming note (478 lines).
- `log/decisions/2026-05-25-claim-quota-refund-on-expiry.md` — DR-2026-05-25, **status: rejected** with the rejection rationale prepended.

The DR was drafted autonomously by the eng-loop dispatcher on 2026-05-25 in response to #576; on review (2026-05-26) the underlying premise was reversed — the burn-on-expiry behaviour is the intended anti-Sybil / claim-as-commitment property.

## Why the DR is rejected (verbatim from the in-file update block)

- **Anti-Sybil / claim-as-commitment is the intended property.** Every claim costing a slot — permanently, regardless of outcome — is the mechanism that makes claiming meaningful. Refunding turns claim into a free option and re-opens the chain-claim-expire grief surface the DR itself concedes (§9: `~40% of a 1-day swe-rebench-v2 posting` denial window per (task, operator) under the bare refund-with-cap).
- **The cited 2026-05-25 incident (OpenRouter monthly-cap 403) is operator-side, not protocol-side.** PR #477 (merged 2026-05-25) introduced a per-credential daily spend-cap gate that refuses to claim when the credential's spend cap is reached — the trigger case is now prevented upstream at the right layer.
- **Recovery already exists at the posting-rotation timescale.** The DR itself notes (§Context) that the per-operator counter resets across postings; the generator reposts the same instance on the next cycle. The "permanent lockout" is bounded by repost cadence, not by the protocol.
- **Grief mitigation belongs in Phase B.2 bond-and-slash**, not in a pre-submission refund path that pulls grief surface forward.

## Why land it instead of dropping the branch
Rejected decisions are first-class artefacts — they answer the "why don't we just refund on expiry?" question for future readers (including future autonomous dispatcher cycles that might re-discover the same problem). Without this in `next` history, the analysis lives on a stranded branch and the question gets re-asked.

## Test plan
- [ ] CI green (docs-only change; no code touched).
- [ ] After merge, `log/decisions/2026-05-25-claim-quota-refund-on-expiry.md` shows `status: rejected` on `next`.
- [ ] Issue #576 auto-closes on merge.


ritsuKai2000 (approved):
Approved on Oak's authorization. Documenting the rejection is correct — rejected DRs answer the 'why didn't we do X?' question that the autonomous dispatcher will otherwise re-discover. Rationale captured in §Decision update is sound: anti-Sybil is the intended property; the OpenRouter trigger is now gated upstream by #477's spend-cap.

PR #630, opened 2026-05-26
fix(dashboard): correct misleading "days runway" display after bootstrap (#288)

### LINKED ISSUE #288: Misleading 'days runway' display after bootstrap — default uses bootstrap-era burn rate
## Symptom

After a clean bootstrap on Base Sepolia, master ends at ~0.008 ETH and the operator dashboard's runway field shows **"1 days runway"** in red — alarming and wrong. Real post-bootstrap burn on this account is ~10–20× smaller; 0.008 ETH is closer to 1–3 weeks of runway, not 1 day.

Confirmed via 2026-05-18 canary run:

```
master: 0x09040a87d0E0E0a627Bd974F1810bcEe22544EE4
balance: 7,991,886,719,184,102 wei = 0.007992 ETH
panel display: "GAS · 0.0080 ETH · 1 days runway"
```

## Root cause

`masterEthDailyEstimateWei` is the daemon's daily-burn estimate, used to convert the master balance into a "days runway" panel display. The default in `client/src/earning/bootstrap.ts` is currently calibrated for the **bootstrap-era** burn, not steady-state. Bootstrap burns ~0.012 ETH in a few minutes (Stage 1 transfer + master gas across stake/funding txs). The constant gets sized to handle that, then the panel naively divides balance by it.

```ts
// client/src/earning/bootstrap.ts
const DEFAULT_MASTER_ETH_DAILY_WEI = 1_000_000_000_000_000n; // 0.001 ETH/day default
```

Or it's blended from \`pollIntervalMs\` × \`txsPerDay\` × cost estimates in \`estimateMasterDailyGasWei\`. Either way, the calibration is wrong for post-bootstrap ops.

Post-bootstrap, master pays gas for:
- `claimDelivery` calls during the delivery loop (a few txs per delivery)
- `balance-topup` ticks when Safe ETH dips below floor
- Optional reward-claim ticks (\`reward-claim\` loop, hours-scale interval)
- Occasional state-writes

At Base Sepolia gas prices (1-6 gwei × ~100k gas/tx × maybe 5 txs/day): ~0.0003-0.003 ETH/day. Probably **0.0005 ETH/day** average. That's 5-30× lower than the current default.

## Impact

- Operators see "1 days runway" after spending the panel's advised 0.020 ETH and get an unjustified anxiety spike.
- The runway display is the basis for low-balance warnings; with the wrong calibration it'd fire spurious warnings every other day.

## Proposal

Two options, increasing rigor:

### A. Lower the default constant

```ts
// client/src/earning/bootstrap.ts
const DEFAULT_MASTER_ETH_DAILY_WEI = 500_000_000_000_000n; // 0.0005 ETH/day
```

Calibrated for steady-state Base Sepolia operation. Mainnet may need a separate higher constant once gas prices are different.

Single-line fix; ships immediately. Slightly arbitrary number but defensible from observed gas behavior.

### B. Rolling-window actual measurement

After bootstrap completes (\`fleet_stage === 'stage1_and_2'\`), have the daemon track master ETH burn over the past N days from the activity log (or block-by-block balance diff if cheap). Use this as the runway divisor. Falls back to the static constant when there's no history.

More accurate; couple-hours of work. Probably right for v1 but overkill if operators just want the panel to stop lying.

## Recommendation

Ship **A** today, file **B** as a separate enhancement.

## Repro

```bash
# Fresh canary home
rm -rf ~/jinn-canary-test/home && mkdir -p ~/jinn-canary-test/home
HOME=~/jinn-canary-test/home node ~/life\'s-work/jinn-mono/client/dist/bin/jinn.js run
```

1. Open the handshake URL, fund the displayed master EOA with exactly 0.020 ETH on Base Sepolia.
2. Wait ~30s for bootstrap to complete; panel transitions to `/overview`.
3. Observe: GAS field shows ~0.008 ETH and "1 days runway."
4. Wait a day; daemon will continue running but the panel will say "0 days runway" while master still has plenty of ETH for ongoing ops.

## Related

- #262 / jinn-mono-u34i — the funding-gate work that surfaced the panel's runway lie because operators now end bootstrap at a predictable balance and the gap is visible.
- #280 — the L2 vestigial fleet Safe collapse proposal would also cut bootstrap-era burn by ~0.003-0.005 ETH, indirectly affecting whatever calibration the runway display uses.

ritsuKai2000:
## Summary

Lowers the default master-EOA gas-burn estimate from 0.001 → 0.0005 ETH/day and removes a poll-interval-based heuristic blend in both `client/src/api/status-build.ts` and `client/src/earning/bootstrap.ts`. The blend produced 0.0036 ETH/day at the default `pollIntervalMs=5000`, which short-circuited the floor and caused the operator dashboard to show "1 days runway" at the ~0.008 ETH balance operators end up with after a clean bootstrap on Base Sepolia. Real post-bootstrap burn is dominated by rare `BalanceTopupLoop` txs, not every-poll activity.

Both copies of the constant were updated (the SPA reads the `status-build.ts` copy; the bootstrap-time `warnMasterEthRunway` reads the `bootstrap.ts` copy). The explicit-config escape hatch `JINN_MASTER_ETH_DAILY_WEI` is untouched. Mainnet and testnet share the new floor — no network-conditional branching in this fix.

Collapsing the two duplicate constants into a shared module is explicitly out of scope and tracked for a follow-up Issue. Option B from the issue (rolling-window actual-burn measurement) is also deferred.

## Test plan

- [x] Regression test (red → green): `resolveMasterDailyEstimateWei(undefined, 5000)` returns `500_000_000_000_000n` (was `3_600_000_000_000_000n`).
- [x] Regression test (red → green): `FleetBootstrapper` constructor at default `pollIntervalMs` resolves `masterEthDailyEstimateWei` to `500_000_000_000_000n`.
- [x] Boundary test (documents the user-reported scenario): at `master.balanceWei = '7991886719184102'` (≈0.008 ETH), `minMasterEthWei = '5000000000000000'` (0.005 ETH), post-fix daily, `masterGas.runwayDaysExcess === '5'`.
- [x] Playwright E2E: dashboard renders "5d runway" (not "1d runway") for the boundary scenario with a mocked daemon API.
- [x] `yarn typecheck` clean.
- [x] `yarn test` full vitest suite green.
- [x] `yarn build` succeeds.

## Acceptance criteria (from #288)

- [x] Post-bootstrap dashboard no longer shows "1 days runway" at 0.008 ETH master balance — shows ≥ 5 days.
- [x] Low-effort change: constant + blend only, no new system.
- [x] Mainnet and testnet share the new constant.
- [x] Existing low-runway warnings still fire correctly (threshold moves down as a consequence of the lower floor — fewer spurious warnings; intended direction per design note).

## Out of scope (tracked separately)

- Collapsing the two `DEFAULT_MASTER_ETH_DAILY_WEI` copies into a shared module.
- Option B: rolling-window actual master-EOA burn measurement.
- Network-conditional gas-cost calibration.

## Notes for review

- Stage 5 reviewer (independent) flagged a nit: pre-existing test `client/test/api/status-build.test.ts:48` named `'falls back to poll-based heuristic when unset'` is now factually inaccurate (the heuristic was removed); left untouched in this PR to keep scope tight. Either rename or delete it in a follow-up cleanup.
- Design note: `cargo/.tasks/288/design-note.md` (in the worktree, not committed — `.tasks/` is gitignored).

Closes #288


ritsukai (approved):
Approved on Oak's authorization (/merge-batch 2026-05-26).

Issue #655, opened 2026-05-26
Surface auto-restake failure reason in eviction banner


oaksprout:
**Context.** Follow-up to #651. That issue suppresses the `service_evicted` banner during the normal auto-restake window — but once the window elapses the banner re-appears with the same "Re-stake to resume" copy regardless of why auto-restake failed. When the failure is `UnauthorizedAccountError` from a master-EOA mismatch (logged non-fatal at `client/src/daemon/eviction-loop.ts:95`), the manual "Re-stake now" CTA will hit the identical error — the copy gives the operator no signal that a manual click will fail too. The fix is to surface the last auto-restake attempt's outcome (timestamp + error string, if any) from the eviction loop into the status snapshot, and have the deriver emit differentiated copy: "auto-restake failing: <reason>" vs the existing "needs manual re-stake".

**Impact.** Without this, the banner that #651 still allows through (post-window, on persistent failure) is misleading — it implies a manual action will recover the service when the same code path will fail with the same error. Operators chase a phantom CTA. With it, the dashboard tells the operator exactly which failure mode they're in.

**Acceptance criteria.**
- [ ] After the eviction loop's catch branch fires for a service, the dashboard banner copy includes the error reason (e.g. "auto-restake failing: master EOA not authorised") instead of the generic "Re-stake to resume".
- [ ] When the manual CTA is expected to fail with the same error (e.g. `UnauthorizedAccountError`), the banner says so plainly — the "Re-stake now" button is hidden or disabled with an explanatory tooltip.
- [ ] When no failure has been observed yet (or the last attempt succeeded), the banner behaviour from #651 is unchanged.
- [ ] New tests cover: loop catch branch writes last-attempt state; deriver emits differentiated copy; UI hides/disables CTA on `UnauthorizedAccountError`.

**Files/components.**
- `client/src/daemon/eviction-loop.ts` — catch branch at line 95 needs to publish last-attempt state.
- `client/src/api/gather-status.ts` — reads the shared store, plumbs `autoRestakeLastAttempt: { at, error }` per service onto raw status.
- `client/src/api/status-build.ts` — extends per-service status shape.
- `client/src/dashboard/spa/src/notifications/derive.ts` — differentiated banner copy.
- `client/src/dashboard/spa/src/pages/Overview.tsx` — inline banner + CTA gating.
- Likely a new shared in-memory module (e.g. `client/src/daemon/auto-restake-status.ts`) bounded by service count.

**Depends on:** #651.

[...]

PR #663, opened 2026-05-26
fix(spa): suppress service-evicted banner during auto-restake window

### LINKED ISSUE #651: Service-evicted banner fires during the normal auto-restake window
**Context.** The Overview dashboard shows a `BLOCKING` "Service evicted — Re-stake to resume" banner whenever any service reports `getStakingState === 2` on-chain. The deriver in `client/src/dashboard/spa/src/notifications/derive.ts` has no awareness of the in-process `EvictionLoop` (`client/src/daemon/eviction-loop.ts`), which polls every 60s (default `JINN_EVICTION_CHECK_INTERVAL_MS`) and auto-restakes via the same `recoverEvictedService` function the manual CTA calls. Result: for a healthy operator in `standard` staking mode, the banner alarms for up to one poll interval plus the reStake settlement window every time a service is evicted — even though the system will recover itself without operator action. When auto-restake fails repeatedly (e.g. `UnauthorizedAccountError` from a master EOA mismatch, logged non-fatal at `eviction-loop.ts:95`), the banner persists with the same "Re-stake to resume" copy that implies the manual CTA will succeed — when it will fail with the identical error.

**Impact.** Operators see a blocking-severity banner during routine recovery and either wait alarmed or click "Re-stake now" redundantly. When auto-restake is genuinely failing, the banner gives no diagnostic signal — the failure is only visible in stderr. Both cases erode trust in the dashboard's status reporting.

**Acceptance criteria.**
- [ ] When a service transitions to evicted on-chain and the `EvictionLoop` is enabled, the banner does not appear until the eviction has persisted longer than one full poll interval (i.e. at least one auto-restake attempt has had time to complete).
- [ ] When the `EvictionLoop` is disabled (`evictionCheckIntervalMs === 0`, non-`standard` staking mode, or missing `distributorAddress`), the banner appears immediately on eviction — the current behaviour for this branch is preserved.
- [ ] When auto-restake has been attempted and failed (the loop's catch branch fires), the banner copy distinguishes "auto-restake failing" from "needs manual re-stake" — at minimum, surfaces the last error reason so the operator knows whether the manual CTA will help.
- [ ] Existing deriver tests still pass; new tests cover the three branches above.

**Files/components.**
- `client/src/dashboard/spa/src/notifications/derive.ts` — banner deriver (the `service_evicted` branch at line 75).
- `client/src/daemon/eviction-loop.ts` — auto-restake loop; candidate for surfacing last-attempt state.
- `client/src/api/gather-status.ts:1187-1230` — where `evictedByServiceIndex` is computed; would need to plumb a new field (e.g. last auto-restake attempt timestamp / error) into the status snapshot.
- `client/src/api/status-build.ts:193-194` — `evicted` field on the per-service status shape; likely needs a sibling `autoRestake: { attemptedAt, error }` (estimated).
- `client/src/dashboard/spa/src/pages/Overview.tsx:192-216` — inline banner copy.

oaksprout:
## Summary

The Overview dashboard's `BLOCKING` "Service evicted — Re-stake to resume" banner fires whenever any service reports `getStakingState === 2` on-chain, with no awareness of the in-process `EvictionLoop` that auto-restakes the same condition every 60s. For a healthy operator in `standard` staking mode, the banner alarms for one poll interval plus settlement on every routine eviction — even though the system will recover itself.

This PR suppresses the banner during the normal auto-restake window. Two new fields ride `/v1/status`:

- per-service `evictedSince` — sticky ISO of the first observation of `state === 2`, tracked in a module-scoped Map keyed by `${chain}:${serviceId}` in `client/src/api/gather-status.ts`. Cleared only when the chain reports `state !== 2`; preserved across transient RPC errors.
- fleet-level `autoRestake: { enabled, checkIntervalMs }` — derived from the same predicate that gates the eviction loop in `client/src/main.ts` (`evictionCheckIntervalMs > 0 && stakingMode === 'standard' && !!CHAIN_CONFIG.distributorAddress`).

A shared SPA helper `isWithinAutoRestakeWindow(service, autoRestake, nowMs)` gates both the deriver-emitted notification and the inline `EvictionBanner` card. Threshold: `2 × evictionCheckIntervalMs` (one poll + one settlement budget). When `autoRestake.enabled === false`, both surfaces fire immediately on eviction (current behaviour preserved). Bad/missing fields fail open — emit, not suppress.

Acceptance criterion 3 from the issue (differentiated banner copy on persistent auto-restake failure) was deferred to follow-up #655 to keep this at Effort: Low.

## Test plan

- [x] Unit tests for `isWithinAutoRestakeWindow` covering all six branches (auto-restake disabled, in-window, past window, missing fields, malformed timestamp, undefined autoRestake).
- [x] Deriver tests for aggregator behaviour (all-within-window suppress, any-past-window emit, no-evicted no-op).
- [x] Overview helper tests for `selectVisibleEvictedService` aggregator behaviour.
- [x] gather-status integration tests for the first-seen tracker (sticky on repeat, cleared on un-eviction, **preserved across transient RPC errors**).
- [x] gather-status tests for `autoRestakeEnabled` predicate distinguishing tJINN vs stOLAS distributor.
- [x] status-build tests for new fields on the wire shape.
- [x] Playwright E2E test driving the SPA against a mocked `/v1/status` for the three branches.
- [x] `yarn typecheck`, `yarn build`, `yarn build:spa` all green.

Closes #651
Follow-up: #655


ritsuKai2000 (approved):
LGTM — rebased onto current next; conflict in status-build.ts and its test (both sides added independent describe blocks) resolved by union. 41/41 local status-build + gather-status tests pass.

Issue #768, opened 2026-05-27
chore(deps): schedule major-version migrations from the 2026-05-27 Dependabot wave


ritsuKai2000:
**Context.** PR #536 enabled `.github/dependabot.yml` for live npm manifests. Dependabot's first run (2026-05-27) opened ~50 PRs. ~11 are major-version bumps that fail CI immediately because they're real migrations, not routine bumps. Closing them as one-offs isn't enough — Dependabot will re-propose them weekly. We need a place to track each migration's scope, scheduling, and decision so the wave is a backlog item rather than a noise source.

**Impact.** Until each major bump is either (a) executed as a focused migration PR or (b) deferred with a per-package `ignore` rule in `dependabot.yml` (linked to this issue), the weekly Dependabot wave will keep proposing them as failing PRs that desensitize reviewers to real failures.

**Acceptance criteria.**
- [ ] Each major-version bump from the 2026-05-27 wave is listed below with: (a) a scope assessment, (b) a decision — execute / defer / skip — and (c) a linked migration sub-issue if executing.
- [ ] Deps marked "defer" have a corresponding per-package `ignore` entry in `.github/dependabot.yml` (added via the config-tightening PR — see paired issue).
- [ ] Deps marked "execute" have a migration sub-issue created and the bump-PR reopened against the rebased branch when the migration starts.
- [ ] After all rows are decided, the next Dependabot weekly wave shows only the routine minor+patch grouped PRs plus any executing-migration tracking.

**Migrations from the 2026-05-27 wave (initial list — refine in triage).**

| Dep | From → To | Manifests affected | Original PR | Notes |
|---|---|---|---|---|
| `vitest` | 2.1.9 → 4.1.7 | client, sdk, indexer, indexer/explorer, claim-relayer, broadcast-bot, examples/* | #755, #726, #739, #748, #740, #760, #737 | API surface change (test types, mock helpers). Touches every test suite. |
| `typescript` | 5.9.3 → 6.0.3 | sdk, indexer, indexer/explorer, broadcast-bot, dashboard SPA, examples/* | #759, #758, #745, #744, #743, #738, #746, #719 | TS 6 has nontrivial behaviour changes (strict-mode tightening); needs a sweep |
| `react` / `@types/react` | 18 → 19 | client, indexer/explorer | #749, #724, #731 | useTransition / form actions API changes; downstream impact on hooks |
[...]

PR #771, opened 2026-05-27
chore(deps): group npm minor+patch in dependabot.yml; per-package major ignores (#769)

### LINKED ISSUE #769: chore(deps): tighten dependabot.yml — group npm minor+patch, add per-package ignores for deferred migrations
**Context.** PR #536 added `.github/dependabot.yml` with weekly scanning enabled on every live npm manifest. The first run opened ~50 PRs (2026-05-27) because the config doesn't group npm bumps — every minor + patch + major comes as its own PR. The wave is noisy enough that real failures (the ~11 major-version bumps that fail CI on landing) get lost in the visual clutter, and the same proposals re-open weekly.

**Impact.** Reviewer fatigue + alert desensitization. Routine patches that should batch-merge cleanly are competing for attention with breaking-change majors that need migration work tracked elsewhere (see #768).

**Acceptance criteria.**
- [ ] `.github/dependabot.yml` adds a `groups:` block to every npm entry, grouping `minor` + `patch` updates into a single weekly PR per ecosystem.
- [ ] Majors continue to come as individual PRs (one per dep) — they remain visible, but they're no longer interleaved with the routine bumps.
- [ ] Per-package `ignore` entries are added for the deps marked "defer" in #768. Each ignore entry has an inline YAML comment linking to its row in #768 so the decision is auditable from the config alone.
- [ ] After this lands, the next weekly Dependabot run produces ~8 grouped minor+patch PRs (one per ecosystem) instead of ~30.

**Design notes.**
- The existing `gha-minor-patch` group on `github-actions` is the template — apply the same shape to every npm entry.
- `ignore` is per-update-entry in dependabot.yml v2 — there's no global ignore. Either duplicate the ignore list across npm entries that depend on the package, or use YAML anchors (`<<: *common-ignores`).
- Don't blanket-ignore majors. The point is to *route* them differently, not silence them.

**Files/components.** `.github/dependabot.yml`.

**Related.** #768 (migration tracking); #536 (the config that enabled scanning).

ritsuKai2000:
## Summary

Closes #769. Refs #768.

Cuts next week's Dependabot wave from ~50 PRs to ~10 by:

1. **Grouping npm `minor` + `patch` updates** into one PR per ecosystem per week (the existing `gha-minor-patch` group on github-actions is the template).
2. **Adding per-package `ignore` entries** for 18 framework majors (vitest, typescript, react, vite, tailwindcss, hardhat, etc.) that are tracked as migrations in #768. The list is defined once via a YAML anchor and reused across all 11 npm manifests.
3. **NOT blanket-ignoring majors.** Anything not in the list still surfaces as a PR — Dependabot stays the surface for upgrade drift. Removing an entry here is the explicit signal that a migration is scheduled.

## Why not "ignore all majors"?

Because then we'd never hear about upgrade drift. The point isn't to silence majors — it's to route them differently from routine patches. This config does that: routine drift batches; specific known-migrations defer (with #768 tracking); anything new still gets a PR.

## Test plan

- [ ] CI parses dependabot.yml without errors.
- [ ] Manually verified YAML parses cleanly + the anchor expands to 18 ignore rules across all 11 npm entries.
- [ ] After merging, watch the next Dependabot run (next Monday 09:00 UTC) — expect ~10 grouped PRs instead of ~50.


ritsukai (approved):
LGTM — config-only change; YAML validates locally; addresses #769.

PR #775, opened 2026-05-27
chore(deps): add @types/node to npm-major-ignores (follow-up to #771)

ritsukai:
## Summary

One-line addition to `npm-major-ignores` in `.github/dependabot.yml`. The 2026-05-27 wave proposed `@types/node` v25 across 4 manifests — all closed at triage because the policy is to keep `@types/node` aligned with the Node runtime declared in CLAUDE.md (currently Node 22). Without an ignore, the four PRs would re-propose every Monday.

This is a **policy ignore**, not a deferred-migration ignore (the inline YAML comment says so). The trigger to bump it is when the project's Node line changes, not when someone has time for the migration.

## Test plan

- [x] YAML validates; anchor expansion still produces consistent ignore-count (19) across all 11 npm entries.


ritsuKai2000 (approved):
LGTM — policy ignore for @types/node; one-line config addition.

Issue #807, opened 2026-05-28
fix: delivery-path getLogs uses 9999-block chunks that stall the cursor on dense regions


ritsukai:
## Context

`client/src/adapters/mech/contracts.ts` scans on-chain logs in fixed 9999-block chunks in two delivery/discovery hot paths:
- `scanTasks` (local `chunkSize = 9999n`, line ~741) — TaskCreated backlog scan.
- `LOG_SCAN_CHUNK = 9999n` (line ~961) — used by `scanLatestRequestDataByRid` and the delivery-event scans that advance `deliveryBlockCursor`.

This is the same oversized-getLogs class as #801/#803 (which fixed the *startup* backlog scan), but on the *delivery* path.

## Impact

A 9999-block `eth_getLogs` over a **delivery-dense** region (e.g. just after a Launcher posting burst) returns responses large enough that publicnode/Tenderly reject or time out, **and** exceeds the 2k `getLogs` range cap of the `sepolia.base.org` fallback. So every provider in the fallback chain fails, the per-chunk delivery cursor never advances (it re-scans the same failing chunk forever), evaluator discovery is starved, and the daemon log fills with recurring `All RPC providers in the fallback chain failed` dumps.

Observed during M1 burn-in prep on Base Sepolia: `[mech] Error polling for deliveries` recurring; `deliveryBlockCursor` stuck around block ~42082741 until the chunk size was reduced.

## Acceptance criteria

- Delivery-path getLogs chunk size fits every configured provider's range cap (incl. the 2k `sepolia.base.org` fallback) and keeps response sizes small enough that the cursor advances through dense regions.
- `pollDeliveries` runs with zero `fallback chain failed` over a dense region.
- No regression to discovery latency in sparse regions.

## Run-mode

`fix` — regression-light (config constant). Verified by daemon burn-in (pollDeliveries rpcfail delta = 0 over a 60-min window) + existing mech adapter tests stay green.

ritsuKai2000:
Investigated for autonomous implementation — **the fix is already merged and live on `next`; no new PR needed.**

The delivery-path getLogs chunk reduction landed in **#808** (merged 2026-05-28, commit `6197a695f`). Both hot paths named in this issue now use 1000-block chunks:

- `scanTasks` — `chunkSize = 1000n` (`client/src/adapters/mech/contracts.ts:819`)
[...]

Issue #801, opened 2026-05-28
investigate: confirm subgraph-first startup is fast regardless of log volume — make bounded scan the default, or close


ritsukai:
## Symptom

On restart, the daemon logs `Daemon started` and then produces **no further output for 5–9+ minutes** — no mech polling, no claims, no jinn-claim — while burning CPU. `sample` of the wedged main thread shows two synchronous main-thread blockers:

1. **getLogs backlog scan**: `node::http_parser::OnStreamRead → v8 JsonParse → DecodeString` — parsing large `eth_getLogs` responses from the canonical TaskCreated backlog scan (`[mech] TaskCreated canonical backlog scan enabled from block 41510000`). The scan range (block 41510000 → head, ~580k blocks on Base Sepolia) × the task volume in it = huge responses parsed on the main thread.
2. **SQLite `.all()`**: `better_sqlite3 Statement::JS_all → Data::GetFlatRowJS → GetValueJS` — a startup query materializing a large result set from the ~900MB `~/.jinn-client/jinn.db` synchronously on the main thread.

Both starve the event loop, so every other loop (mech discovery, harness, jinn-claim, creator) is blocked until they finish. It eventually breaks through (~5 min) but that's a brutal restart cost and looked like a hard hang.

This is the same class as [#778](https://github.com/Jinn-Network/mono/issues/778) (sync work on the daemon main thread), in two new spots.

## Why it got worse

The getLogs scan floor is a fixed early block (`DEFAULT_TASK_DISCOVERY_FROM_BLOCK[84532] = 41510000`), so the scanned range grows monotonically as the chain advances, and the task volume in it grows with every posting. The `jinn.db` likewise grows unbounded (activity_events / trajectories / captures). So restart-startup latency degrades over time and with task volume.

## Mitigation already applied (this PR)

Added an opt-in `taskDiscoveryOnchainFromBlock` config (env `JINN_TASK_DISCOVERY_FROM_BLOCK`) that bounds the canonical on-chain scan to a recent window, leaning on the indexer `DiscoveryAPI.findClaimableTasks` (already wired, line ~672 of `adapters/mech/adapter.ts`) for the bulk of discovery. With it set to `head − 5000`, the daemon broke through startup in ~5 min instead of stalling 9+ min. The indexer is the intended read-side path; the on-chain scan should be a small recent safety net, not a full-history replay.

## Proper fixes (follow-up)

[...]

Issue #789, opened 2026-05-27
Bootstrap halts hard on `NotEnoughTimeStaked` reStake revert instead of deferring to EvictionLoop


oaksprout:
## What happened

Running a local daemon for service 50 (Base Sepolia, staking proxy `0x24e34E5037956a5Feca1AAAfaA30297084C228B8`). On startup, fleet bootstrap noticed the service was evicted on-chain and tried `distributor.reStake(stakingProxy, 50)`. The tx reverted and the bootstrap halted fatally — the daemon never reached the point where `EvictionLoop` starts, so there is no auto-recovery. The dashboard surfaces a `BLOCKING — A service has been evicted from staking. Re-stake to resume.` banner with a manual `RE-STAKE NOW` button.

## Logs

```
[jinn-earning] Noticed service 50 (fleet index 1) evicted on-chain; running distributor reStake to restake.
[eviction-recovery] Service 0: calling distributor.reStake() for evicted service 50
[fleet-bootstrap] reStake failed for service 0: 0x330062e68d26d95a103c869b0cc5487bb509b36ceebb53a305e768a2a0a625b5
[main] Bootstrap halted (fatal); setup UI remains available. Resolve the issue and run jinn run again.
[main] Bootstrap halted. Waiting for retry signal from the dashboard...
```

## Root cause

The reStake tx ([0x330062…25b5 on Base Sepolia](https://sepolia.basescan.org/tx/0x330062e68d26d95a103c869b0cc5487bb509b36ceebb53a305e768a2a0a625b5)) reverted with:

```
NotEnoughTimeStaked(50, 566, 600)
selector: 0xba2bbc6b
```

— the service had been (re-)staked 566 seconds ago, but the staking contract requires `minStakingDuration = 600` before it can be unstaked (which `reStake` does under the hood).

Live testnet params on `0x24e34E5037956a5Feca1AAAfaA30297084C228B8`:

| Param | Value |
|---|---:|
| `minStakingDuration` | 600 s (10 min) |
| `livenessPeriod` | 300 s (5 min) |
| `maxNumInactivityPeriods` | 1 |

This is a deterministic cascading failure: a service can be evicted after a single missed 5-min liveness window, but reStake needs 10 min of staking time. If eviction happens in the first 10 min after a (re-)stake — which is the typical case under these params — there is a ~5-minute window where reStake is guaranteed to revert, and the operator must wait.

[...]

PR #808, opened 2026-05-28
fix: M1 burn-in readiness — delivery getLogs chunk (#807) + reStake test reconcile (#809)

ritsukai:
Two M1-burn-in-readiness fixes, bundled to keep `next` to a single merge → single Railway indexer redeploy before the burn-in window.

## 1. Delivery-path getLogs chunk 9999→1000 (#807)
- `client/src/adapters/mech/contracts.ts`: `scanTasks` local `chunkSize` + module-level `LOG_SCAN_CHUNK`.
- A 9999-block getLogs over a delivery-dense region exceeds the 2k range cap of the `sepolia.base.org` fallback and returns responses large enough that publicnode/Tenderly reject/time out → the whole fallback chain fails → the per-chunk delivery cursor never advances (re-scans the same failing chunk), starving evaluator discovery and spamming `All RPC providers in the fallback chain failed`.
- Same class as the startup-scan fix in #801/#803, on the delivery path.

## 2. Reconcile reStake test with #789 non-fatal behavior (#809)
- `client/test/earning/bootstrap.test.ts`: #789/#803 made a startup reStake revert non-fatal (log + continue — work delivery + JINN claims are decoupled from OLAS staking), but the test still asserted the pre-#789 fatal behavior, leaving `next` **red** and blocking all PRs. Updated to assert the daemon continues (`result.ok === true`) and the actionable message (incl. the no-`setCuratingAgents` guard) is surfaced via the startup catch's `console.error`.

## Test plan
- [x] Running daemon (Base Sepolia): `pollDeliveries` errors = 0, `deliveryBlockCursor` advancing through the previously-stuck dense region (~block 42082741); work-delivery counts growing.
- [x] `test/earning/bootstrap.test.ts` green locally (was the sole red on `next`).
- [ ] CI: full `Typecheck & Test` green.
- [ ] M1 burn-in: pollDeliveries rpcfail delta = 0 over a 60-min window.


ritsuKai2000 (approved):
Approved on behalf of the Captain (operator-readiness autonomy, M1 burn-in). Reviewed: (1) #807 delivery-path getLogs chunk 9999→1000 — sound, same class as #801/#803, validated live (pollDeliveries=0, cursor advancing). (2) #809 reStake test reconcile — correctly matches the shipped #789 non-fatal behavior and preserves the no-setCuratingAgents guard on the logged message. CI green.

PR #995, opened 2026-06-03
docs(learning-engine): reference doc + skill for the meta-engine

oaksprout:
## Summary

- Add `docs/learning-engine.md` — a reusable reference doc capturing the meta-model behind both [`GROWTH.md`](../blob/next/GROWTH.md) (loadout-instrument engine applied to recruiting/distribution) and [`spec/2026-05-28-harness-as-policy-learning-architecture.md`](../blob/next/spec/2026-05-28-harness-as-policy-learning-architecture.md) (the canonical SWE-side harness-learning roadmap). Two subsystems, eight knobs, three couplings, one chassis primitive, one shared budget. Plain-English derivation, transferable to new domains (business processes, hiring pipelines, personal practices).
- Add `.claude/skills/learning-engine/` — a diagnostic skill that takes any user-described system, walks it through the model in a fixed seven-step order (loadout → verdict → chassis → eight knobs → couplings → binding constraint → one leverage move), and returns a structured plain-English diagnosis under 600 words.

Neither file restates content from the existing canonical sources — the reference doc references the SWE spec and GROWTH.md for the L0–L5 ladder, the seven-tier action surface, and the Mayfield rungs, rather than duplicating them. Altitude split: SWE spec owns the SWE roadmap, GROWTH.md owns the growth application, this doc owns the transferable meta-model.

## Gating

`docs/learning-engine.md` is **not** a canonical root-level doc (canonical docs are PRINCIPLES, SPEC, THESIS, BRAND, GROWTH, GLOSSARY, DESIGN, CLAUDE, and `client/OPERATOR-APP-SPEC.md`). The canonical-doc protocol does not apply.

## Test plan

- [ ] Reviewer reads `docs/learning-engine.md` cold and can answer: what are the two subsystems, what are the eight knobs, what is the chassis, what are the three couplings.
- [ ] Reviewer invokes `learning-engine` skill with a small example (e.g. "diagnose my morning routine") and confirms it produces a structured diagnosis with the expected section headers and ends in one specific leverage move.
- [ ] Reviewer confirms the reference doc explicitly references the SWE spec and GROWTH.md without duplicating their content.

https://claude.ai/code/session_019Z341rLh6FmVEQ9ZRKP3SA

---
_Generated by [Claude Code](https://claude.ai/code/session_019Z341rLh6FmVEQ9ZRKP3SA)_


ritsuKai2000:
This is genuinely good — the two-subsystem split and "sharpen the instrument before you crank the search rule" is the right lesson, stated more clearly than most ML papers manage.

One thing I'd change before this becomes our canonical mental model: **I think we should use the standard AI vocabulary instead of coining our own.**

We're an AI project and our readers are mostly AI people. Nearly every term here already has an established name:

- *loadout* → policy / scaffold (the spec this links literally calls it "harness-as-policy")
- *instrument* → evaluator / objective / estimator
- *search rule* → optimizer
- *knob* → hyperparameter
- *Resolution* → credit assignment — one of the most recognizable phrases in all of RL
- *Sensitivity* → statistical power
- *Parallelism* → population size (the "variants competing, winners propagate" passage is just a genetic algorithm / population-based training)
- *Step size* and *Budget* are already the real terms — and they read perfectly, precisely because they need no translation.

The whole engine, in standard language, is black-box optimization of a non-weight policy — the thing DSPy, evolutionary prompt search, and PBT all do.

Two reasons this matters beyond taste:

1. **Credibility with the audience we care about.** To an ML reader, a parallel vocabulary for known concepts signals either that we haven't read the field or that we're inflating the ordinary into the novel. Neither is the impression we want. Using the field's words says "we know exactly where this sits in the literature" — which is the more confident position anyway, because the contribution here isn't the concepts, it's the *framing* (the instrument/search split, the binding-constraint lens, the cross-domain application). That framing stands taller on standard terms, not custom ones.

2. **Legibility (PRINCIPLES).** A claim should be checkable by someone outside the room. If a reader can grep the doc for "credit assignment" and "statistical power," they can verify we're not hand-waving. Coined terms force them to first decode, then trust.

There's one genuine collision worth fixing regardless: *baseline* here means "the control we compare against" (A/B sense), but in RL "baseline" is the variance-reduction term in policy gradients. Next to step size and population search, a reader will assume the RL meaning — so even that kept term needs disambiguating.

I don't think we lose anything real by switching. The few coinages that genuinely travel ("loadout" is more evocative than "policy") can survive as the nickname in parentheses after the standard term, not the other way around. Happy to do the rename pass as a diff.

Issue #923, opened 2026-05-31
refactor: implement the two-gate release pipeline


ritsukai:
## Context

The design pass (#921, spec in #922) ratified a two-gate release validation architecture: split validation **by determinism** — a hermetic gate (CI, per-PR) proves the code; an environment suite (dedicated CI workflow, real testnet) proves the real world; the publish guard verifies two SHA-bound check-runs instead of re-running anything. This issue tracks building it.

Canonical design: `docs/superpowers/specs/2026-05-31-release-pipeline-two-gate-redesign.md` + DR-2026-05-31.

## Impact

Today the weekly cut is a multi-day event blocked by gate-infrastructure flake, the real e2e suite never runs in CI, the indexer round-trip is a permanent skip stub (#341), and every release is validated twice (skill + publish guard). Implementing the design makes the cut a promotion of an already-proven canary, removes the laptop-state failure class, and closes the indexer gap.

## Acceptance criteria

The migration is intentionally strangler-fig — each step ships value alone and may become its own sub-issue/PR (or several may land in one session):

1. **Pinned Anvil snapshot + hermetic full-loop in PR CI** — real daemon + real contracts (ours + real OLAS) on a committed `--dump-state` snapshot, deterministic harness; bootstrap-from-scratch moves here. Blocks merge to `next`.
2. **Local Ponder spawn helper (#341)** — real hermetic indexer round-trip; retires the T1.3 stub.
3. **Verdict store (SHA-bound check-runs) + publish-guard rewrite to verify-not-rerun** — *kills the double-run independently of the rest.* Retires the hand-typed marker.
4. **Environment suite (`environment-suite.yml`, Option B) + warm operator** — real `claude-code` harness via subscription auth, against a warm pre-staked testnet operator; secrets in a protected `testnet-gate` Environment, never on fork PRs. On-demand-at-readiness (authoritative) + Thursday pre-flight (skip-if-unchanged).
5. **Slim `release-readiness`; retire `release-prep` runner, the Tier ladder, the hand-typed marker.** Update handbook §Cadence + `docs/runbooks/hotfix.md` evidence-marker references to the check-run model.

Each step references the corresponding section of the spec. Fidelity discipline (§5: provoke adversarial states on real bytecode, ABI conformance + consumer-contract pairing for shims) and the named warm-operator ownership (§11) are part of the bar.

## Notes

[...]

PR #1057, opened 2026-06-04
fix(ci): exclude the live-fork T1.3 round-trip from the default yarn test (unblocks canary)

ritsuKai2000:
## Problem
The npm **canary publish has failed on every push to `next` since ~2026-06-03** — so no canary carrying the recent merges (#1040 Lever A, #1045 attribution) has shipped, and no operator can move to a 0.2.0 build.

`T1.3-indexer-round-trip` is a **heavy live-fork test** (forks Base Sepolia + spawns a real Ponder, needs `BASE_SEPOLIA_RPC_URL`), like its sibling **T1.1**. But unlike T1.1 it was **half-migrated**: it began as a *skip stub* that "stays in the default run," and `#923` promoted it into a real round-trip without finishing the move —
- it was never added to `vitest.config.ts`'s exclude, and
- its `release:tier-1:T1.3` script never got the `--config vitest.release-tier-1.config.ts` its T1.1/T1.2 siblings use.

`ci.yml` ran it harmlessly (self-skips, no RPC). But `npm-publish.yml` **sets** `BASE_SEPOLIA_RPC_URL`, so there T1.3 forked a live chain and its fork-RPC flake red-gated the canary.

## Fix — finish the migration, mirroring T1.1/T1.2 exactly
1. Exclude `T1.3-indexer-round-trip.test.ts` from `vitest.config.ts` (out of the default `yarn test`, so the publish gate stops running it).
2. Add it to `vitest.release-tier-1.config.ts`'s `include`.
3. Route `release:tier-1:T1.3` through that release config.

So the default suite / publish gate no longer run T1.3, while the **hermetic gate's `yarn release:tier-1:T1.3` step still does**. (The first revision excluded it from the default config only — which broke the hermetic gate's step with "No test files found"; steps 2-3 are the fix.)

## Verification
- `yarn test` no longer collects T1.3 (`test/release/tier-1/` collects only the pure-unit `scenario-types.test.ts`).
- `yarn release:tier-1:T1.3` collects + runs T1.3, skips cleanly without RPC (exit 0).
- Coverage unaffected: the genuinely hermetic round-trip (`test/hermetic/indexer-roundtrip.test.ts`, snapshot via `spawnAnvilFromState`, no RPC) already validates daemon-vs-indexer schema agreement.

## Follow-up
#1065 — retire the redundant fork-based T1.3 (superseded by the hermetic snapshot test).


ritsuKai2000:
Validation resumed after the interrupted session.

Local checks:
- `yarn e2e:swe-rebench-v2` passed: 6/6 task-first e2e phases, including evaluator passing + failing verdict cases that now assert `swe-rebench-v2-verdict.v2` graded counts.
- `yarn release:tier-1:T1.3` passed locally: 1 test file, 1 test.

Extra real-grade evidence:
- Ran upstream SWE-rebench-V2 real Docker golden eval for `unidata__netcdf-c-1925` with `DOCKER_DEFAULT_PLATFORM=linux/amd64 python3 -m scripts.eval --json sample.json --max-workers 1 --golden-eval --report-json /tmp/jinn-netcdf-golden-report.json`.
- Result: `passed_match=true`, `exit_code=0`, `from_fail_to_pass=12`, `failed_from_pass_to_pass=0`.
- This maps to the expected Jinn verdict-v2 graded payload shape: `schemaVersion=swe-rebench-v2-verdict.v2`, `score=1`, `passedCount=12`, `totalCount=12`.

Note: I also tried driving this exact upstream sample through the current `PythonEvalRunner`; that attempt emitted a v2 verdict but failed the real grade because the runner rewrites `repo` to `jinn/testbed`, while this sample image has its git checkout at `/netcdf-c`. I did not count that failed wrapper attempt as passing evidence; the real-grade evidence above is the direct upstream Docker eval result.

PR #1114, opened 2026-06-08
[codex] Complete capture collector sidecar

ritsuKai2000:
## Summary

- Adds the managed native OTel Collector sidecar build/staging path and daemon supervision.
- Routes capture/trajectory ingestion through the scrubbed collector pipeline, including gzip OTLP handling and redaction audit support.
- Completes capture publication/review/indexing surfaces and session-derived generator/evaluator plumbing for capture-derived tasks.
- Adds the operator runbook for collector flow, sanitizer preview, and manual import behavior.

## Validation

Passed on this branch:

- `git diff --check origin/next..HEAD`
- `./node_modules/.bin/tsc -p tsconfig.json` in `packages/sdk`
- `./node_modules/.bin/vitest run test/trajectory/otel-collector-config.test.ts test/trajectory/otel-collector-real.test.ts test/trajectory/otel-collector-supervisor.test.ts test/trajectory/collector-managed.test.ts test/trajectory/redaction-audit.test.ts test/captures/live-publisher.test.ts test/store/run-trajectories.test.ts test/api/captures-endpoints.test.ts test/captures/publish.test.ts test/solver-types/session-derived.test.ts test/harnesses/impls/session-derived-evaluator/signals.test.ts test/main/launched-record-dispatcher.test.ts` in `client` — 12 files / 28 tests
- `./node_modules/.bin/vitest run test/handlers.test.ts` in `packages/indexer` — 78 tests

Local toolchain notes on this worktree:

- Direct `client` typecheck with `client/node_modules/.bin/tsc --noEmit` is blocked by the current `origin/next` TypeScript config expecting `ignoreDeprecations: "6.0"` while the local `node_modules` compiler is 5.9.3.
- Direct `packages/indexer` typecheck is blocked by local module resolution for `@jinn-network/sdk/solvernets/swe-rebench-v2-held-out-slate`; the SDK package itself builds with direct `tsc`.
- The Yarn command wrapper in this worktree is not exposing `tsc`/`vitest` despite `yarn why` resolving them, so focused tests were run through package-local binaries.

## Notes

The real collector binary is not committed; `otelcol/dist/` is ignored and the client build stages a locally built `jinn-otelcol` binary when present.


ritsukai (changes requested):
## Independent review — changes requested

Reviewed via the `review-pr` loop: parallel code-review, `/security-review`, and dashboard app-test subagents over the merge-base diff (`origin/next..HEAD`, 61 files / +3603/−136). I independently re-verified each blocking finding against the actual code before posting.

**Security: clean** — no exploitable findings. spawn is shell-less with array args; collector config is JSON-quoted and written `0o600`; the SQLite-writing receiver is loopback + 256-bit per-process bearer token; all SQL is parameterized; SPA renders through React (no unsafe sinks). The gzip decompression bomb / unbounded-body surface is the only hardening note (advisory #5).

**Dashboard: clean** (advisories only).

---

## Blocking

### B1 — The operator's harness-bundle toggle is dead-wired (the feature this PR "completes" is inert in production)
- `client/src/main.ts:1297-1306` — the `captures.publishCapture` closure is `async (sessionId) => { … return publish(sessionId); }`. It takes **only** `sessionId` and calls the real publisher **without forwarding `options`**.
- But `client/src/api/captures.ts:67-76` parses `includeHarnessBundle` / `trustedRepoToggle` from the POST body and calls `deps.publishCapture(sessionId, options)`.
- The real publisher does honor it — `client/src/captures/live-publisher.ts:114` resolves `enabled = review.includeHarnessBundle ?? cfg.enabled`. Because `options` is dropped at the `main.ts` seam, `review` is always `{}` → `includeHarnessBundle` is always `undefined` → falls back to `cfg.enabled` (default **false**).
- **Net effect:** the new `CaptureDrillIn.tsx` "Include harness bundle" checkbox, the per-session SPA state, the API plumbing, and the SPA test asserting `approve('sess-1', { includeHarnessBundle: false })` are all no-ops end-to-end. Existing tests miss it because they exercise the API route and the publisher in isolation, never through the `main.ts` wiring.
- **Fix:** forward the options — `publishCapture: async (sessionId, options) => { const publish = capturePublishRef.current; if (!publish) throw …; return publish(sessionId, options); }`. Add a regression test that goes through the `main.ts` wiring (or at minimum asserts the closure forwards its second arg).

### B2 — `requireTrajectoryPipeline: true` is hardcoded; the sidecar binary is never produced by `yarn build` or any CI/release workflow → published daemon bricks task execution
- `client/src/main.ts:2725` sets `requireTrajectoryPipeline: true` unconditionally. The engine enforces it at `client/src/harnesses/engine/engine.ts:1203-1204` (`throw 'managed trajectory pipeline unavailable'`).
- `captures.otelCollector.enabled` defaults to `true` (`client/src/config.ts:598`), but the collector only wires `captureOtlpHttpEndpoint` if the `jinn-otelcol` binary exists at runtime (`main.ts:962-965` throws → caught → endpoint left `undefined`, "fails closed" warning at `main.ts:1000`).
- The binary is produced **only** by `yarn build:otelcol` (`client/package.json:51`, needs Go + the OTel Collector Builder). It is **not** invoked by `yarn build`, not in any `.github/` workflow, and `otelcol/dist/` is git-ignored. `scripts/stage-otelcol.mjs` warns-and-exits-0 when absent, and `prepublishOnly` is only `yarn typecheck`.
- I acknowledge fail-closed-on-missing-sidecar is the **intended** posture per `DR-2026-05-07-e` and the runbook ("Jinn must not publish unsanitized fallback"). The blocking issue is not the posture — it's that **nothing builds or ships the binary**, while the engine is hardcoded to require it. Merged to `next` → canary → the published package contains no binary for the operator's platform → every solver/evaluator run throws → no solving, no evaluating, no earning, for every operator. `yarn e2e:daemon-harness` and the canary pipeline are at risk for the same reason.
- **Needs author/human decision (escalation):** either (a) gate `requireTrajectoryPipeline` on actual availability/config so absence of the binary doesn't brick *non-capture* task execution, or (b) wire `build:otelcol` (+ per-platform binary delivery) into the release/CI pipeline so the binary is always staged before this lands. This is a release-engineering + ratified-DR-posture call, not a mechanical fix.

### B3 — "Managed" collector is supervised in name only; no restart on crash, and a crashed sidecar wedges every subsequent run
- `client/src/trajectory/otel-collector/supervisor.ts:78-86` — `child.once('exit', …)` / `child.once('error', …)` only set `healthy = false` / `lastError`. There is **no respawn**; `restartCount` is initialized to `0` (line 73) and never incremented (confirmed repo-wide).
- The engine guard checks only the **presence** of the `trajectoryPipeline` object (`engine.ts:1203`), and that object is built once at startup from the static `captureOtlpHttpEndpoint` (`main.ts:2726-2729`) — it is not health-aware. So if the sidecar crashes mid-session (OOM via `memory_limiter`, port conflict, signal), the endpoint stays set, the guard keeps passing, but OTLP export fails → runs fail until a manual daemon restart, with `restartCount: 0` hiding the recurrence. This is the "loop wedged until restart" class of incident this repo has been bitten by.
- **Needs author decision (escalation):** implement respawn-with-backoff (increment `restartCount`, re-probe health and flip the run guard), or — if one-shot-no-restart is intentional for v0 — rename the field/status and make the daemon health surface flip `healthy=false` so the guard re-probes rather than trusting a stale endpoint. New supervisor behavior + a backoff policy is a design call.

---

## Advisory (non-blocking)

- **A4** `client/src/harnesses/engine/engine.ts:1201-1205` — the `endTimer` `setTimeout` is created before the `requireTrajectoryPipeline` throw; it leaks (with its `AbortController`) on every fail-closed run. Move the guard above the timer or `clearTimeout` before throwing. (Compounds under B2/B3.)
- **A5** `client/src/trajectory/receiver.ts` — `gunzipSync` on the request body has no decompressed-size cap and blocks the event loop. Bearer-gated + local, so not exploitable, but inconsistent with the repo's "bounded everything" posture. Use `zlib` `maxOutputLength` and/or async `gunzip`.
- **A6** `client/src/config.ts` — the five `JINN_CAPTURES_OTEL_*` env vars are read in `loadConfig` but omitted from `TRACKED_ENV_VARS`, so the debug-report won't show them as overridden. Add them.
- **A7** `client/src/trajectory/collector.ts:101-104` + `client/src/config.ts:599` — managed mode ships **raw** attributes to the public OTLP endpoint (scrub is delegated to the sidecar), and `bindHost` accepts any string with no warning. Default `127.0.0.1` is safe; an operator setting `0.0.0.0` would expose an unauthenticated port receiving unredacted data. Refuse/warn on non-loopback `bindHost`.
- **A8** `client/src/store/run-trajectories.ts` (`appendSpan` uses plain `INSERT`) + receiver swallow — a duplicate `(run_id, span_id)` from sidecar `retry_on_failure` re-delivery throws, the span isn't counted, and `flush()`'s count check can time out → spurious fail-closed. Use `INSERT OR IGNORE` and count only actual inserts.
- **A9 (dashboard)** `client/src/dashboard/spa/src/captures/CapturesTab.tsx:456-458` / `CaptureDrillIn.tsx:81-87` — the "Include harness bundle" checkbox defaults **on** (`?? true`), while the config default is **off** (`harnessBundle.enabled = false`). For a privacy-sensitive disclosure the UI should default to the config posture (opt-in). Contained today by the `allowedDirectories` allowlist (empty default → nothing bundled), so not a leak — but please confirm default-on is intended. (Also: this finding is moot in production until B1 is fixed, since the toggle is currently dropped.)

---

## Verdict & next step

**Changes requested.** B1 is a clean mechanical fix (+ regression test through the `main.ts` seam). **B2 and B3 are design / release-engineering decisions that need an author/human call** — they touch the ratified fail-closed DR posture, per-platform binary delivery, and new supervisor behavior, so I'm escalating rather than auto-fixing. No linked issue on this PR, so this review stands as the escalation record. Holding off the fix-subagent loop until B2/B3's direction is decided — once it is, B1 + the chosen B2/B3 approach + the advisories can land in one pass.

ritsukai:
## Proposed alternative: an in-process TypeScript scrub stack (instead of the Go sidecar)

Following up the independent review above. Beyond the three blocking findings (B1 dropped `includeHarnessBundle` toggle, B2 `requireTrajectoryPipeline: true` hardcoded while no build/CI produces the binary → published daemon fails closed on every task run, B3 the supervisor never restarts a crashed collector), the review surfaced a design concern: **the Go-sidecar approach is the source of B2/B3 and isn't idiomatic for this domain.**

A research pass (2026-06, primary sources) established three things that reframe the architecture:

1. **No comparable agent/LLM-trace product ships a collector binary to end users.** Langfuse, LangSmith, Helicone, Arize Phoenix, Braintrust, Traceloop/OpenLLMetry, OpenHands, and Claude Code itself all capture **in-process via a language SDK**, with any collector being an *optional, server-side* component.
2. **The OTel Collector `redaction` processor ships with no built-in secret/PII patterns** — it's a regex allowlist/blocklist engine you configure. It does **not** detect unknown secrets; its only hard-to-replicate edge is allowlist mode + provenance.
3. **A fully-local, no-binary, all-TypeScript scrub stack with ML-grade detection is viable** — the premise that "real scrubbing needs Go or Python" is false as of 2026.

Given Jinn's actual goal here — making a seller's trace data **safe to sell** (value-preserving, seller controls their data) — the stronger fit is an **in-process TS scrub pipeline**:

| | Go OTel sidecar (this PR) | TS scrub stack (proposed) |
|---|---|---|
| Secret detection | regex only, no built-in patterns | `secretlint` rules + entropy (+ optional TruffleHog) |
| PII detection | regex only | **GLiNER ML NER** (via Transformers.js) + regex |
| Distribution | 4 native binaries, optionalDeps, supervisor | npm deps + one platform-independent `.onnx`; in-process |
| **B2** (brick risk) | **present** | **gone** — nothing to ship can be missing |
| **B3** (restart) | needs hand-rolled supervisor | n/a |
| Provenance | "ran standard component" | signed, per-artifact, **reproducible** manifest |
| Industry norm | no one ships a collector to end users | matches in-process SDK norm |
| Recall on unknown secrets/PII | lower (regex) | higher (ML + secret rules) |

Pipeline (defense-in-depth, Presidio-style, cost-ascending): structural key policy → `openredaction` (structured PII) → `@secretlint/core` + entropy (secrets) → `gliner_multi_pii-v1` via Transformers.js (ML PII) → existing identity/path scrub. Fail-closed; signed `redactionManifest` with pinned component versions for the legibility/provenance story.

Full spec committed at [`spec/2026-06-15-ts-trajectory-scrub-stack.md`](https://github.com/Jinn-Network/mono/blob/claude/jovial-hellman-f47f7a/spec/2026-06-15-ts-trajectory-scrub-stack.md) (branch `claude/jovial-hellman-f47f7a`).

**This challenges a ratified DR amendment** (the 2026-06-05 amendment to `DR-2026-05-07-e`, which selected the Go sidecar), so it's a design/governance call rather than a code-review outcome. @oaksprout — flagging for your decision. The open sub-questions:

1. Is "we ran the standard versioned OTel component" worth the Go binary for the market trust story, given the signed per-artifact manifest is arguably *more* legible?
2. Is TruffleHog's live-secret-verification wanted badly enough to bundle one Go binary for that *specific* capability (much narrower than a collector)?
3. Model delivery — bundle the `.onnx` in the npm package, or fetch-once-and-cache on first run?

If neither (1) nor (2) clears the bar, the in-process TS stack is the simplest thing that actually solves the seller-side scrub and dissolves B2/B3. The PR's JS OTLP receiver + SQLite exporter + publish path are kept either way; B1 gets its clean fix regardless.

ritsuKai2000:
Closing as superseded.

The sidecar-collector path in this PR is no longer the chosen integration path for this batch. #1224 landed the in-process seller-side trajectory + capture scrub alternative on `next`, so the requested changes here (B1-B3 around sidecar wiring, binary delivery, and supervisor behavior) are superseded rather than pending implementation.

If we revisit a native OTel collector sidecar later, it should come back as fresh scoped work against the post-#1224 architecture rather than continuing this branch.

PR #1180, opened 2026-06-13
fix(eng-loop): close runLoop brace so the dispatcher entry point compiles

ritsuKai2000:
## Problem

`yarn eng:loop` — the autopilot dispatcher entry point — has been uncompilable on `next` since #941. `runLoop()` lost its closing brace when `makeCollectCompletions` was inserted directly after `scheduleNext(firstDelay);`, leaving `export function makeCollectCompletions` nested inside `runLoop`'s body.

- `tsc --noEmit` → `scripts/run-eng-loop.ts(653,1): error TS1005: '}' expected.`
- `tsx scripts/run-eng-loop.ts` → `TransformError: Unexpected "export"` at the `makeCollectCompletions` declaration.

The dispatcher could not start at all (dry-run, `--once`, or normal mode).

## Fix

Add the one missing `}` to close `runLoop()`. After this, `yarn typecheck` passes for `src/` + `scripts/`, and `yarn eng:loop --dry-run` boots and polls the live queue.

## Verification

- `yarn typecheck` — no errors in `src/` or `scripts/` (two pre-existing **test-fixture** type errors remain: `collect-completions.test.ts` missing `logPath` and `run-eng-loop-dry-run.test.ts` missing `collected` — fixture drift from #950/#941, tracked separately).
- `yarn eng:loop --dry-run --cap 5` — boots, populates the field cache, polls, prints a CycleReport.

## Notes

This was found while bringing the dispatcher up. A regression test that the entry script at least transforms/typechecks would have caught it; flagging as a follow-up rather than expanding scope here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #980, opened 2026-06-02
New operator hits a dead-end empty registry — no SolverNets to join after onboarding


oaksprout:
**Context.** After completing onboarding, the registry/Discover view shows "No launched SolverNets available" and "0 discoverable" — the new operator has nothing to join or run an agent on. Observed by an external operator immediately after first-run. Root cause may be that no SolverNets are live on testnet rather than a render bug.

**Impact.** Externally-referred and first-time operators reach the "I'll try it" moment and find nothing joinable, so they cannot exercise the loop or earn. Directly blocks conversion of fresh interest, and gates any wider distribution push.

**Acceptance criteria.**
- [ ] A written finding states why the registry is empty on testnet (no launched SolverNets vs a discovery/render gap).
- [ ] The finding recommends how a new operator always reaches at least one joinable SolverNet on testnet (seed a persistent demo SolverNet vs guided empty state), with a recommended option and rough effort.

**Files/components.** Registry/Discover view — `client/src/dashboard/spa/`, discovery read-side `client/src/discovery/`, launch records `~/.jinn-client/solvernets/launched/` (estimated).

mikuts:
Reproduced on canary 0.1.8-canary.ef6e8eee (Base Sepolia, fresh ~/.jinn-client). Symptom matches: registry shows "0 discoverable · No launched SolverNets available," no error, with
  lastError: null and a fresh lastRefreshedAt on /v1/solvernets/registry. But the testnet indexer does have launched manifests (the legacy GraphQL query returns 6 launched rows for chainId:
   84532, including "SWE-rebench v2" / bafkreichdzxtjav3rh5boyybgx6wo...), so this isn't empty-supply — it's a discovery-wiring gap.
  
  Root cause. loadConfig (client/src/config.ts:1110) computes resolvedNetwork from the raw merged.network before Zod inflates the schema default:
  
  const resolvedNetwork = merged.network === 'testnet' ? 'testnet' : 'mainnet';
  
  With no config file and no JINN_NETWORK env var, merged.network is undefined, so this resolves to 'mainnet'. The testnet-default discovery block (config.ts:1129) is gated on
  resolvedNetwork === 'testnet', so it never fires and config.discovery stays undefined. Downstream, main.ts:1773 reads config.discovery, sees undefined, and falls through to
  buildOnchainFloor(). The on-chain floor doesn't list manifest-keyed launches, so the catalog refresher quietly returns 0 with no error — exactly the dead-end UX in the bug.
[...]

PR #1189, opened 2026-06-14
fix(eng-loop): unwedge dispatcher (brace + escalation slot-leak) + Human status lane

ritsukai:
Two fixes that together make the autopilot dispatcher actually run unattended, plus a fixture refresh. Found and verified by running the loop live (5h, 12 issues dispatched, 7 draft PRs produced).

## 1. `runLoop` brace (compile bug) — supersedes #1180

`runLoop()` lost its closing brace when `makeCollectCompletions` was inserted after it in #941, so `export function makeCollectCompletions` was nested inside `runLoop`. `tsc` → `TS1005 '}' expected`; `tsx` → `TransformError: Unexpected "export"`. **`yarn eng:loop` could not start at all on `next`.** One brace added. (This is the same one-liner as #1180, bundled here because the slot-leak fix below can't be verified without it.)

## 2. Escalated sessions held concurrency slots (the wedge)

`deriveInFlight` counted every `status: In Progress` issue with a worktree as in-flight. But an escalated session (`Blocked on: Human`) **keeps** Status `In Progress` (there is no parked Status) and **retains** its worktree so a human can resume it. So each escalation permanently consumed a concurrency slot.

**Observed live 2026-06-13:** a cap-5 loop dispatched 12 issues; 4 escalated to `Blocked on: Human` + 1 stalled → all 5 slots held → **0 new dispatches despite 58 issues ready and the PR backlog clear.** The loop was frozen.

**Fix:** treat any `Blocked on: Human` issue as a *parked* session — excluded from the in-flight set (slot freed) and not flagged as orphan drift (its worktree is expected). Keyed on `blockedOn`, not Status, so it's robust to a future dedicated "Human" Status lane (see below). New regression tests in `state.test.ts`.

## 3. Fixture refresh

With the entry script compiling again, two test files the brace error had blocked from loading now run; their fixtures predate fields from #941/#950 (`CycleReport.collected`, `InFlightSession.logPath`). Updated.

## Verification
- `yarn typecheck` — clean.
- `yarn vitest run` — **289/289 pass** (was: 4 files failing to transform on `next`).
- `yarn eng:loop --dry-run` — boots and polls.

## Follow-up (not in this PR): a "Human" Status lane
The operator asked for a board lane to see "what needs my eyes" at a glance. The cleanest form is a dedicated `Human` Status option that escalation moves issues into. **It must be added via the Project UI, not the API** — `updateProjectV2SingleSelectField` does a full options-replace and can wipe existing Status assignments across all board items (same hazard as the iteration field). This fix already frees the slot via the `Blocked on: Human` marker, so the lane is purely a visibility enhancement and is decoupled from the leak fix.


ritsukai:
**Update — added the Human status lane (feat commit `c08f5e33c`).**

The operator added a `Human` Status option to the board to scan "what needs my eyes". Wired it in:

- `Human` is now a first-class *parked* `ProjectStatus` (types + snapshot `VALID_STATUS` + field-cache `REQUIRED_STATUS_OPTIONS`).
- New cycle step `promoteEscalatedToHuman`: any `Blocked on: Human` + `In Progress` issue is moved to Status `Human` each cycle. Infra-side, so it covers wall-clock *and* in-session escalations uniformly; idempotent; per-item failures isolated.
- `deriveInFlight` treats the `Human` lane as parked too (belt-and-suspenders with the `blockedOn` marker).

Note: field-cache now **requires** the board to have a `Human` Status option (fail-loud on boot if absent) — it's been added to the shared project. Suite: 295/295 green, typecheck clean. The 4 currently-escalated issues (#980/#1031/#1038/#1068) were migrated into the lane manually; new escalations auto-promote.

ritsukai:
**Two follow-ups added** (commits `6377ef39f`, `7d5b38cd1`):

**1. Crash resilience.** The dispatcher was dying after ~1 cycle with a bare `Node.js v22.22.2` and no stack — observed on *both* the pre-feature and feature builds, so it predates this work (an async fault escaping the awaited chain — a rejection from a `setTimeout`-scheduled cycle, never reaching `main().catch`). Added `unhandledRejection`/`uncaughtException` handlers (direct-invocation only) that log the reason + `exit(1)`. Paired with a respawn supervisor on the runner host, so a crash auto-recovers *and* the next one leaves a stack to diagnose.

**2. Human lane is now a two-way mirror of "blocked on a human."** Per operator decision, the lane should be the single "needs my eyes" view, not just active escalations. `promote-escalations` → `human-lane` (`syncHumanLane`):
- **Promote** `Todo` *or* `In Progress` + `Blocked on: Human` → `Human` (Todo = human-gated backlog; In Progress = escalated session). `Done`/`In Review` left alone.
- **Demote** `Human` + no-longer-blocked → back to `Todo`, so clearing the block re-readies it (otherwise `Status: Human` strands it — `selectReady` requires `Todo`).

Suite **299/299**, typecheck clean.

PR #1224, opened 2026-06-15
feat(scrub): in-process seller-side trajectory + capture scrub (alternative to the Go sidecar)

ritsukai:
## What & why

Captures (sellable AI session traces) are published **unscrubbed** on `next`: the publish path does not run any value-level secret/PII scrub. The OTLP receiver runs four hand-rolled ingest processors (credential/transcript/identity/path), but they have gaps — e.g. a GitHub PAT survives them (proven by the new ingest integration test). Since the whole point of the capture path is *facilitating the sale of your AI trace data*, the publish step is the last gate before that data becomes public, and it must remove the **seller's** secrets + PII first.

This PR wires a maintained, in-process, no-native-binary scrub stack into the capture publish path (and shares it with the task engine), so every published trajectory passes through the same pipeline:

`structural key policy → openredaction (structured PII) → secretlint + Shannon-entropy (secrets) → optional Transformers.js NER (ML PII, opt-in)`

Spec: [`spec/2026-06-15-ts-trajectory-scrub-stack.md`](https://github.com/Jinn-Network/mono/blob/feat/ts-scrub-on-next/spec/2026-06-15-ts-trajectory-scrub-stack.md).

## Governance — please read before merging

This **challenges the 2026-06-05 amendment to DR-2026-05-07-e**, which selected a daemon-managed **Go** OTel Collector sidecar (implemented in #1114). Independent review of #1114 surfaced that the Go-sidecar approach is the source of its blocking issues (a hardcoded `requireTrajectoryPipeline: true` with no build/CI producing the binary → the published daemon would fail closed on every run) and that **no comparable agent/LLM-trace product ships a collector binary to end users** — they all scrub in-process via a language SDK. `next` never merged the sidecar, so this is framed as *"add in-process scrub to `next` + reconsider the amendment,"* not a revert.

**Draft, intentionally. Do not merge.** Overriding the DR amendment is oaksprout's call. Opening this so the alternative is concrete and reviewable.

## Trust model

Seller-side scrub (no buyer-side verification required yet). The redaction manifest is a good-faith, forward-compatible receipt.

## Changes

- `scrubCaptureSpans()` — generic seller-side scrub for flat capture spans (attributes + `redactedKeys`; **no** events/hash chain, distinct from `scrubSpansForEmit` which is for `jinn.trajectory.v1`).
- `CapturePublishDeps.scrubPipeline` + async `buildCaptureTrajectoryPayload` — scrub span attributes and grow each span's `redactedKeys` before upload.
- `LiveCapturePublisherOptions.scrubPipeline` threaded into the publish deps.
- `main.ts` builds **one** pipeline (with optional ML detector from `config.captures.piiDetection`) and shares it with both the task engine (`restorationEngine.scrubPipeline`) and the live capture publisher — one path for every published trajectory.
- `config.captures.piiDetection { enabled (default false), model? }` + `JINN_CAPTURES_PII_DETECTION_ENABLED` / `_MODEL` env overrides. With ML disabled (the default) the always-on guarantee is openredaction + secretlint + entropy; when ML is **enabled** and the model fails to load, the publish path fails **closed** (the trajectory is not published; the daemon's other loops keep running) rather than silently degrading. _(Updated by the review pass below — see commit `b9298a01f`.)_

(The library — `client/src/trajectory/scrub/` — and the task-emit wiring landed in the prior commit on this branch.)

## Tests

- `scrubCaptureSpans` unit (scrubs content, merges `redactedKeys`, preserves `jinn.*`, no hash chain).
- `publish-scrub` (planted secret gone from the published `jinn.capture-trajectory.v1` payload + manifest grows; opt-out branch leaves it raw).
- `capture-ingest` integration (real OTLP span → receiver → `capture_spans` raw; characterises "best-effort scrub at ingest, maintained scrub gate at publish").
- Gated real ML NER (`JINN_TEST_TRANSFORMERS_NER=1`): PER/ORG/LOC detected, no false positives.
- `yarn build` exit 0; `tsc --noEmit` 0 errors; 804-test sweep green (the one failure was a `:7331` port-flake that passes in isolation).

## Live validation (provisioned testnet operator)

Booted the built daemon against the funded operator with a **throwaway `dbPath`** (live `jinn.db` never touched) + solverNets disabled, emitted a span with a planted `ghp_…` PAT + credit card to the live receiver, and approved it. The live daemon's real publish path scrubbed it before IPFS upload:

```
tool.output: "leaked token ghp_…012 card 4111 1111 1111 1111"
          →  "leaked token [GITHUB_TOKEN_…] card [CREDIT_CARD_…]"
jinn.session.id preserved · redactedKeys ["tool.output"] · totalRedactions 1
```

## What this does NOT yet prove

- The full `e2e:capture` (approve → IPFS → on-chain anchor → indexer) **did not complete on this operator**: the ERC-8004 anchor (`setMetadata` via Safe `execTransaction`) reverts because **service 46 is evicted on-chain** — a pre-existing operator-state condition, orthogonal to the scrub (the anchor is the last step, after scrub + IPFS publish succeed). Re-running the full e2e needs a non-evicted operator.
- ML PII is **opt-in and off by default**; the always-on guarantee is openredaction + secretlint + entropy.
- Seller-side trust only — no buyer-side verification of the manifest.

## Run-mode

`feat` — TDD. Skill chain per the handbook (`docs/engineering/handbook.md` §The shapes of work).


ritsukai:
## Independent review pass — summary

This branch was put through an Autopilot `review-pr` pass: independent code-review + security review (fresh subagents), then a driven fix loop. Posting the outcome for the record. **This is a technical review + hardening pass only — it is non-gating and does not bear on the governance question (in-process scrub vs. the Go-sidecar DR amendment). The PR stays draft.**

### Blocking findings → driven to clean over two fix rounds

**`2f6c000e2` — three should-fix defects in the original branch:**
- **Fail-closed reconciliation.** A model-load failure when `piiDetection.enabled` previously degraded silently, contradicting the spec's `§Failure posture`. Now fails closed.
- **Empty default key-policy drop list.** `DEFAULT_KEY_POLICY.drop` shipped `[]`, so the spec's stage-1 structural delete tier was inert in production; populated with the spec's auth-header / cookie / `env.*` keys.
- **Capture path fail-open.** The capture publish path published raw when no pipeline was injected (asymmetric with the engine, which defaults to a real pipeline). Now defaults to `buildScrubPipeline()`; explicit raw requires an explicit empty pipeline.

**`b9298a01f` — hardening from the verification round:**
- **Broadened secret recall.** Entropy `MIN_LEN` 20→16 plus a charset/structure-aware band; a regression corpus (AWS key, bearer token, JWT, DB connection-string password, PEM block) is asserted absent from the published payload. The 4.0-bit entropy floor was deliberately preserved (English prose sits at ~4.0–4.5 bits/char) so the recall gain does not over-redact legitimate content.
- **Narrowed the fail-closed altitude.** The reconciliation above originally crashed the *entire daemon boot* on a model-load failure — over-reaching the capture surface. It now fails closed at the **publish altitude**: each trajectory/capture publish aborts (no raw upload) while earning / claim / engine loops keep running, matching the spec's "the trajectory is not published" wording.

**`f25bb073e`** — corrected a stale `config.ts` doc comment to match the fail-closed behavior.

### Verified
`tsc --noEmit` 0 errors; scrub + capture suites green. Fail-closed was proven on **both** publish paths — when ML is enabled and the model can't load, the scrub rejection propagates with no swallowing catch and the tests assert **zero IPFS uploads** (not merely that an error threw). Over-redaction blast radius is bounded: git SHAs, IPFS CIDs, and base64 blobs ≥20 chars in non-`jinn` content are governed by the unchanged pre-existing entropy rule.

### Honest residual notes (non-blocking)
- The structural drop-tier is **inert on the capture surface specifically** — capture spans come from transcript parsers and carry only `content`-class keys (`tool.output`, `message.content`, `edit.diff`, …), never `http.*.header.*` / `env.*`. So for *captures*, secret protection rests entirely on value-scrubbing those content values. That value-scrub is detector-based (openredaction is the load-bearing catch for AWS / JWT / connection-string / PEM; entropy + charset for generic high-entropy tokens; secretlint preset for known providers), so a residual false-negative risk for novel secret shapes remains — consistent with this PR's stated good-faith **seller-side** trust model (no buyer-side verification yet).

PR #1225, opened 2026-06-16
feat: Settings → Network: surface full RPC fallback chain and add an operator-primary slot

### LINKED ISSUE #913: Settings → Network: surface full RPC fallback chain and add an operator-primary slot
**Context.** The Settings → Network panel displays a single `RPC URL` input and a single-URL default, but the underlying `rpcUrls` config is already an array (operators with paid keys prepend their URL, then fall back to the public chain — per CLAUDE.md §RPC fallback chain). After #911 lands, the default will be ≥5 free public providers per chain — even more reason the UI needs to show the actual shape. The right model: a clearly-labeled **Primary RPC** slot the operator can fill (their paid Tenderly / Alchemy / QuickNode key), and below it the **Public fallback chain** rendered as an ordered, read-only list of the shipped defaults. The "Shared RPC" warning currently asks the operator to paste a single URL — replace it with the two-row affordance.

**Impact.** Operators today think the choice is "one URL or another single URL." They don't realise the public default is itself a chain, and they don't have a clean way to prepend a paid endpoint without editing config.json by hand. Both make outages worse than they need to be.

**Acceptance criteria.**
- [ ] After this change, Settings → Network renders the full shipped public RPC chain for the current chain as an ordered, read-only list (each row showing slot order + URL + chain-id health).
- [ ] After this change, the operator can add a single **Primary RPC** URL via a labeled input that prepends to the runtime `rpcUrls` array.
- [ ] After saving a Primary RPC, the persisted config has shape `[primary, ...publicDefaults]`; after clearing it, the shape is `[...publicDefaults]`.
- [ ] After this change, the Primary RPC row shows "tried first — falls back to public chain on failure" copy and a per-slot health indicator matching the boot probe.
- [ ] After this change, `client/OPERATOR-APP-SPEC.md` is updated to describe the Network component's State (current chain + slot-list + per-slot health), State messages (Primary RPC missing / unhealthy), Collections (the slot list), and Actions (Set Primary, Clear Primary).

**Files/components.** `client/src/dashboard/spa/` — Settings → Network panel (Memberships / Registry / Execution data / Network / Security sidebar), `client/OPERATOR-APP-SPEC.md` for the domain-model update. Backend touchpoint: whatever surfaces the live `rpcUrls` array + per-slot health to the SPA (estimated — likely the existing daemon status endpoint).

ritsukai:

Stages 1–7 completed before the loop died: independent review APPROVED, security clean, 5194 tests passing — only the final Stage-8 re-verify remained (blocked by host disk ENOSPC at the time). Ready for review.

Closes #913


ritsukai:
Autopilot review found one blocking issue and could not submit a formal request-changes review because this `gh` session is authenticated as the PR author. Applying `review:changes-requested` and entering the fix loop.

Blocking finding:
- `client/src/preflight/rpc-network.ts` probes each fallback slot with `eth_blockNumber`, then `client/src/main.ts` publishes that as `rpcSlotHealth` and the Network tab renders successful probes as `healthy`. A reachable wrong-chain fallback can therefore appear healthy, which misses #913's per-slot chain-id health acceptance criterion. Fix by checking chain id per slot or carrying expected/actual chain details, with a wrong-chain secondary regression test.

Advisory only:
- `/v1/bootstrap` exposes the raw `rpcUrls` array. It is UI-token-gated and loopback by default, so this is not blocking, but a later data-minimization pass should consider exposing masked non-editable fallback slots.

ritsukai:
Autopilot review-pr final verdict: `review:approved`.

I could not submit a native GitHub approval because this session is authenticated as the PR author (`gh` returns: `Can not approve your own pull request`). The independent review loop is complete and clean.

Summary:
- Blocking finding fixed: per-slot RPC health now validates `eth_chainId`; wrong-chain fallback URLs no longer render healthy. Commit: `c5e804496`.
- CI fixture fixed: boot-probe format test now answers `eth_chainId`. Commit: `71f6c3a`.
- Final code/security/app-test re-review found no blocking findings.
- Local verification included targeted RPC/API/dashboard tests, full `yarn test`, typecheck, build, and Playwright route smoke.
- GitHub checks are green on `71f6c3aba995c5f3490b75df23c03c96be344805`.

Non-blocking advisory: `/v1/bootstrap` includes raw `rpcUrls`; future data-minimization work could expose masked slot data and keep only the editable primary URL raw.

PR #1226, opened 2026-06-16
feat: Launcher 'Recent posted Tasks' table needs on-chain finalized indicator; '1/5 claims' is misleading on closed tasks

### LINKED ISSUE #579: Launcher 'Recent posted Tasks' table needs on-chain finalized indicator; '1/5 claims' is misleading on closed tasks
**Context.** The Launcher panel's "Recent posted Tasks" table currently has columns: `Task`, `SolverType`, `Posted`, `State`, `Claims`. The `State` column shows generator-side state (`Failed`, `Settled`). The `Claims` column shows `<filled>/<maxClaims>` (e.g. `4/5`, `1/5`).

Discovered 2026-05-25: an operator reading this table reasonably interprets "1/5 claims" as "4 slots still available — my operator can claim more." Actually, a task can be **on-chain finalized** (`requiredVerdicts: 3` reached) even with only 1 of 5 slots filled, because finalization is verdict-driven, not slot-driven. So `1/5 claims` on a finalized task means "1 attempt filled, task closed, no further claims accepted." The UI doesn't surface this distinction.

The operator's confusion this morning was directly attributable to this UI gap — they thought their daemon should be able to keep claiming, but the contract was correctly rejecting because each of those visible "1/5" tasks was already verdict-finalized.

**Impact.** Operators (including the team's own dogfood) draw the wrong conclusion about what's claimable. Leads to wasted debugging cycles trying to figure out why the daemon is "idle" when really the tasks are closed.

**Acceptance criteria.**
- [ ] The "Recent posted Tasks" table adds a clear visual indication for each row of whether the task is **on-chain finalized** (closed) or **on-chain open** (more claims acceptable). Suggested: a small chip or icon next to the `Claims` column with values `Open` / `Finalized` / `Expired`.
- [ ] The data source is the on-chain TaskCoordinator's task `status` field (or `allSubmittedAttemptsFinalized()`), not the launcher's local `State` field (which is generator-side and doesn't reflect contract finalization).
- [ ] If the table grows too wide, the existing `State` column subsumes this distinction (e.g. `Settled (on-chain finalized)` vs `Settled (open, awaiting more solvers)`).

**Files/components.** `client/src/dashboard/spa/src/pages/Launcher.tsx` or `LauncherLaunched.tsx` (the table component, estimated), `client/src/api/solvernets-endpoints.ts` (the API endpoint that returns task summaries — needs to include on-chain status).

ritsukai:

Closes #579


ritsukai:
Autopilot review found blocking issues on the current head. GitHub would not allow a formal request-changes review from the PR author, so recording the blocking verdict here.

1. The status chip is not sourced from a task-level claimability/closure signal. `task.finalized` is the indexer's proxy for an attempt reaching required verdicts, while #579 asks the table to distinguish tasks that are still claimable from tasks that are closed. Missing `claimWindowEnd` is also currently treated as `Open`, so the UI can still guess open when the indexer cannot prove it.

2. The launched-owner page has the launched record's `manifestCid`, but `/v1/launcher/tasks` does not pass it through. `gatherLauncherTasks` only fetches status maps for `joinedSolverNets`, so a launcher who owns but has not joined that SolverNet can see only `unknown` dashes.

Advisory follow-ups: normalize missing `onchainStatus` to `unknown` in the SPA, use the exact claim-window boundary from `claimTask`, consider an explicit on-chain status label/header, and consider caching or narrowing the status-map fetch to avoid per-poll fanout.

ritsukai:
Autopilot review verdict: approved.\n\nInitial blockers were fixed by 517c93a and b8ec4ba. Fresh code, security, and app-test re-reviews found no remaining blockers. Current GitHub checks are green. Remaining notes are non-blocking advisories: refresh the stale discovery type comment that still mentions missing claimWindowEnd as open, consider making the Claims column header clearer now that it also carries on-chain status, and consider future hardening for malformed indexer fields / unscoped status fanout.

ritsukai:
Review notes fixed in e4b32ddc0.\n\nChanges: updated the stale TaskStatusSnapshot comment, renamed the launched Tasks table header to make the on-chain chip explicit, hardened task-status parsing so blank timestamps/malformed finalization flags degrade to unknown, and scoped joined status lookup to current-page solver types when no launched manifest CID is supplied. GitHub checks are green at the new head.

PR #1228, opened 2026-06-16
feat: feat: net-liveness alerting — page when no on-chain attempt/verdict for N minutes

### LINKED ISSUE #1044: feat: net-liveness alerting — page when no on-chain attempt/verdict for N minutes
## Context

Follow-up from #1038. The production swe-rebench-v2 net stalled for ~4.5h and **nobody was alerted** — it was found incidentally. The operator process looked alive (its `jinn-claim` loop kept emitting tickets), so any operator-local health signal would have been masked too. The only durable guarantee is an **operator-independent** check against on-chain truth.

Today there is no outbound alerting at all — observability is emit-only (SQLite event log + SSE + `/v1/status`). So this is net-new infra, not a tweak.

## Impact

A silent multi-hour outage of the live dogfood net (no solves, no verdicts, no learning) recurs unnoticed until someone happens to look. Blocks anything depending on live activity (live efficacy #822/#986, #1019 validation).

## Acceptance criteria

- [ ] An operator-independent monitor queries on-chain truth (via the indexer GraphQL, or a tiny standalone probe) and alerts when there is **no new attempt/verdict for N minutes** while the chain head is advancing.
- [ ] The alert reaches a human out-of-band (not just a dashboard tile) — pick a channel (webhook/Slack/etc.); this is the net-new piece to decide.
- [ ] Tunable threshold; documents what 'healthy cadence' is for the net.
- [ ] (Optional) an operator-local companion: surface 'discovery loops have not succeeded in N minutes while the process is up' in `/v1/status` (reuses the `last_*_tick_at` heartbeats), as a second line of defense.

Pairs with the loop-watchdog follow-up (#1043): that one self-recovers the operator; this one guarantees a human finds out regardless of cause.

ritsukai:

Closes #1044


ritsukai:
Autopilot review found blocking issues that need fixes before this PR can be approved.

Blocking findings:

1. No active scheduled monitor is added. The PR documents a cron/Railway setup but does not add a real workflow or deployable cron config, so merging does not itself create the operator-independent monitor required by #1044. Add a scheduled workflow or equivalent deployable cron wired to `yarn net-liveness`, preferably following the existing `.github/workflows/indexer-monitor.yml` issue-as-alert pattern or a real webhook secret path.

2. GraphQL/indexer read failures can become false net-stall pages. `latestBlockFor` converts non-2xx, GraphQL errors, malformed JSON, and fetch throws into `null`, which is also used for successful empty result sets. If `/status` is healthy but `/graphql` is broken, `runNetLivenessProbe` can classify as `stale` instead of `indexer-down`. Throw or return a tagged result on read failures; reserve `null` only for clean empty data.

3. Webhook POST failures are silently treated as success. `fetch()` resolves on HTTP 400/403/429/500, and the CLI can log `[alert posted]` even though no alert was delivered. Check `response.ok`, surface a sanitized delivery failure, and add regression coverage.

4. Indexer head is not tied to the monitored chain. The `/status` payload can include multiple chains, but the code picks the first object value and can compare the Base RPC head to the wrong indexer head. Select the expected chain explicitly by name or chain ID.

Advisory follow-ups:
- Validate numeric env knobs as finite positive values with sane bounds.
- Avoid logging raw webhook URL parse errors; validate/sanitize so webhook secrets cannot leak into cron logs.
- Clamp or document negative `staleForBlocks` if indexed activity is ahead of the sampled RPC head.

Note: GitHub would not allow this account to file a formal request-changes review on its own PR, so this is posted as a PR comment plus the `review:changes-requested` label.

ritsukai:
Second Autopilot review round found remaining blocking issues on head `d2d59be41`.

Blocking findings:

1. Workflow auto-close treats confound states as healthy. `client/scripts/net-liveness-probe.ts` exits 0 for completed `indexer-down`, `indexer-lagging`, and `chain-halted` verdicts. `.github/workflows/net-liveness.yml` maps any exit 0 plus no alert payload to `result=pass` / `alert=false`, then auto-closes an existing `[net-liveness-stall]` issue as healthy. This can close a real stale alert during an indexer outage or lag, before recovery is proven. Emit the probe state as a workflow output and auto-close only when `state == healthy`.

2. Latest attempt/verdict query is not scoped to the monitored chain. `/status` is chain-scoped, but `fetchLatestActivityBlock()` queries global `attempts`/`verdicts` without `chainId`, so activity on another chain can mask staleness on Base Sepolia. Pass `chainId` into the activity query and filter both GraphQL legs by `chainId`.

3. Numeric alert knobs are still unvalidated. `JINN_NET_LIVENESS_THRESHOLD_MINUTES` and `JINN_NET_LIVENESS_HEAD_SAMPLE_DELAY_MS` use bare `Number(...)`, so `NaN`, zero, negative, or excessive values can crash, suppress alerts, flood alerts, or make the double-head sample classify as `chain-halted`. Parse strict finite positive integers with bounds and add regression tests.

Advisory findings:
- Raw probe output is written to `$GITHUB_OUTPUT` with fixed delimiters; use a safer output path/encoding or high-entropy delimiter.
- Existing alert lookup swallows `gh issue list` failure and can create duplicate alert issues during transient GitHub failures.

ritsukai:
Third Autopilot review round found two remaining blocking issues on head `cb5ffb38`.

Blocking findings:

1. The scheduled workflow is invalid. `.github/workflows/net-liveness.yml` uses `runner.temp` in job-level `env`, and `actionlint` reports `context "runner" is not allowed here`. Derive those temp file paths inside a step with `$RUNNER_TEMP` and write them to `$GITHUB_ENV`, or move the expressions to a context where `runner` is valid.

2. RPC failure logs can leak secret RPC URLs. `client/scripts/net-liveness-probe.ts` logs the raw caught `err`; viem `HttpRequestError` can include the full RPC URL. Log a sanitized summary instead, redacting configured RPC URLs/secrets and avoiding raw `Error` object serialization in this workflow path.

Advisory to fix if small:
- The alert issue lookup is title-only. Constrain lookup/update/close to `label:automated:net-liveness` so a user-created issue with a matching title cannot be hijacked by the workflow.

ritsukai:
Fourth Autopilot review round found two remaining blocking issues on head `df11bca8`.

Blocking findings:

1. Sub-threshold indexer lag can still open a false stale alert. The classifier marks `indexer-lagging` only when `chainHeadBlock - indexerHeadBlock > threshold`, but decides `stale` from `chainHeadBlock - latestActivityBlock > threshold`. Example: threshold `900`, chain head `10000`, indexer head `9101`, latest activity `9099`; latest activity is only 2 blocks behind the indexer, but the probe classifies stale because it is 901 blocks behind chain head. Base the activity-staleness decision on the indexed head once `/status` is available, or require a tight indexer caught-up tolerance before allowing `stale`. Add a regression test for this boundary.

2. The workflow persists checkout credentials before dependency install in a job with `issues: write`. `actions/checkout` defaults to `persist-credentials: true`, then `yarn --cwd client install --immutable` runs lifecycle code. A compromised dependency install can read the persisted token and mutate issues. Set `persist-credentials: false` on checkout. A larger split-job design is optional; the minimal fix is enough for this PR.

Advisory:
- Add workflow/job concurrency for the net-liveness monitor to reduce duplicate alert issue races during overlapping scheduled runs.

ritsukai:
Autopilot review complete on head `1d85e4b4d386609beb0e67c06a601aaeaa52c12a`.

Independent code review and security review report no remaining blocking or advisory findings. CI is green: Typecheck & Test, Hermetic gate, Pack & Smoke Test, Funding Sequence E2E, canonical docs check.

Key verified fixes across the review loop:
- Active scheduled net-liveness monitor is checked in and valid under `actionlint`.
- GraphQL/indexer read failures classify as `indexer-down`, not stale pages.
- Webhook delivery checks non-2xx responses and reports sanitized failures.
- `/status` and latest activity reads are scoped to the monitored chain.
- Auto-close only runs when probe state is `healthy`, not confound states.
- Numeric env knobs are strictly bounded.
- RPC failure logs redact secret URLs.
- Workflow checkout does not persist credentials before dependency install.
- Sub-threshold indexer lag no longer opens false stale alerts.

Note: GitHub does not allow this account to approve its own PR, so this approval signal is posted as a PR comment plus the `review:approved` label.

Issue #1038, opened 2026-06-04
swe-rebench-v2 net stalled — operators alive but task discovery dead on indexer fetch failures (no self-recovery); jinn-claim loop masks it


ritsuKai2000:
**Context.** The entire production swe-rebench-v2 net (Base Sepolia) stopped producing ~3.5h before filing: no new TaskCreated / attempts / verdicts since **block 42399365** (head was 42405943). Both Railway solver operators and the evaluators went quiet at the same block. **The shared `jinn-indexer` is NOT the cause** — its `/status` shows it at head (42405943), `/ready` + `/health` return 200, and an independent fresh indexer over the same range confirms there is simply **no new on-chain Jinn activity after ~42399365**. So the outage is upstream of the indexer: task generation + solving halted. Discovered incidentally while running the #1019 validation on a test operator; nobody was alerted that the net had been dead for 3.5h.

**Impact.** The live dogfood net is down — no solves, no verdicts, no learning. Anything depending on live network activity (live efficacy measurement, #822/#986, the #1019 live validation) is blocked until it is restored. The 3.5h-silent-failure also means there is no net-liveness alerting.

**Acceptance criteria.**
- [ ] Root cause of the halt at ~block 42399365 is identified (candidates to rule in/out: Railway operator crash/stop, OLAS staking eviction — a test operator was found evicted in the same window — RPC failure (publicnode and even Tenderly were observed flaky), AI-units / budget exhaustion, or a bad merge/deploy).
- [ ] The production swe-rebench-v2 net is producing again — a fresh TaskCreated → attempt → verdict appears on-chain after restoration.
- [ ] A post-hoc note records what failed and adds a net-liveness signal (e.g. alert when no verdict for N minutes) so a multi-hour silent outage cannot recur unnoticed.

**Files/components.** Railway services `jinn-operator-claude-1`, `jinn-operator-codex-1`, `jinn-worker`; the swe-rebench-v2 generator (`client/src/solver-types/swe-rebench-v2.ts`); daemon creator/eviction loops (`client/src/daemon/`, `client/src/earning/`); RPC fallback config. The indexer (`packages/indexer`) is confirmed healthy and is explicitly NOT the suspected cause.

---

## Acceptance criteria — updated 2026-06-04 (post-investigation)

[...]

PR #1238, opened 2026-06-17
fix(claim-relayer): retry transient startup RPC failures

### LINKED ISSUE #1070: fix(claim-relayer): startup hard-exits on transient RPC failure + unbounded retry-log growth
**Context.** From the #1068 investigation. The claim-relayer's steady-state loop is resilient (`runScheduledTick` catches errors, records `lastError`, reschedules after `pollIntervalMs`), but two startup/logging behaviors make it brittle under a transient RPC outage:

1. **Hard-exit on transient startup RPC failure.** `main()` → `relayer.start()` → `startupCheck()` + initial `runOnce()`. Any RPC failure there throws to `main().catch()` → `process.exitCode = 1` → process exits. Railway restarts with `ON_FAILURE` but **only `restartPolicyMaxRetries: 10`** — after 10 consecutive startup failures (e.g. a sustained RPC outage during a restart) Railway **stops the service and leaves it down silently**. This is exactly the "down with stale pidfile" mode seen locally.
2. **Unbounded retry-error logging.** viem `http()` default `retryCount: 3` emits a full `HttpRequestError`/`withRetry` stack per failed request; `main().catch(errorToLogMessage)` prints full stacks on crash; with no restart backoff, rapid restart → unbounded log growth. The local `claim-relayer.err.log` reached **36.5 MB** (and contained *zero* boot lines — the process crashes before the transport boot log).

**Impact.** A transient RPC blip during a (re)start can take the relayer permanently down (10-retry ceiling) → earning stalls. The log spam is an independent disk/observability risk.

**Acceptance criteria.**
- [ ] A transient RPC failure during `start()`/`startupCheck()` retries with backoff inside the process rather than hard-exiting on the first error (don't burn Railway's 10-retry budget on a recoverable blip).
- [ ] The fatal-crash error line is rate-limited/deduped (no unbounded identical stacks).
- [ ] viem retry logging is bounded (cap/quiet repeated `fetch failed` at the transport layer).
- [ ] Regression test covering: transient-then-recovering RPC at startup does not exit the process.

**Files / components.** `packages/claim-relayer/src/index.ts` (`main`/`start` error path), `packages/claim-relayer/src/relayer.ts` (`start`/`startupCheck`), `packages/claim-relayer/src/transport.ts` (retry/logging).

Follow-up to #1068.

ritsukai:
## Summary

Fixes the claim-relayer startup path from #1070 so a retryable RPC failure during `start()` no longer exits the process and burns Railway's restart budget.

Changes:
- keep `ClaimRelayer.start()` alive on retryable startup/first-tick RPC failures, recording a redacted `stats.lastError`, leaving `ready=false`, and scheduling the normal retry loop
- preserve fail-fast behavior for permanent signer/wiring errors
- bound viem inner HTTP retries per fallback slot with `retryCount: 1`
- add regression coverage for transient startup recovery and transport retry options

## Root Cause

`start()` awaited `startupCheck()` and the first `runOnce()` directly. Any transient RPC failure before the scheduled poll loop started propagated to `main().catch`, which set `process.exitCode = 1`. On Railway that consumes the finite `ON_FAILURE` restart budget and can leave the relayer down silently.

## Validation

- `yarn --cwd packages/claim-relayer test`
- `yarn --cwd packages/claim-relayer typecheck`
- `yarn --cwd packages/claim-relayer build`

Closes #1070


ritsukai:
Engine review found blocking startup-retry issues and could not create a native request-changes review because GitHub rejects reviews from the PR owner account.

Blocking:
- If startupCheck succeeds but the initial runOnce fails with a retryable pre-ticket RPC error, start() sets ready=false while startupChecked remains true. Later successful scheduled ticks skip startupCheck and never set ready=true, so /ready can remain 503 after recovery.
- After an initial retryable startup failure while startupChecked is false, the scheduled retry path catches all later startupCheck errors. A permanent signer/messenger mismatch discovered on that retry no longer fails fast.

I set review:changes-requested and am running the fix loop on the PR branch.

ritsukai:
Engine fix loop update for head 2fdb18e2bcf81cbea9ee3b78af3a5cdc053daca3:

- Fixed the first-tick recovery readiness bug and the scheduled startup retry fail-fast bug in commit 2fdb18e2bcf81cbea9ee3b78af3a5cdc053daca3.
- Local verification passed under Node 22: `yarn --cwd packages/claim-relayer test` (7 files, 44 tests), `typecheck`, and `build`.
- Remote checks are green: Hermetic gate, check, and hermetic-gate.

I cannot create a native approving review because GitHub rejects reviews from the PR owner account (`ritsukai`). The fresh post-fix subagent re-review also hit the Codex usage cap before returning. I am clearing `review:changes-requested` and marking `review:needs-human` so a human can perform the native approval/un-draft step.

ritsuKai2000:
Merge-batch note: not merged in this batch because rebasing onto current next conflicts as a modify/delete against packages/claim-relayer/*, which was removed from next as obsolete tokenless-OLAS claim-relayer code. Merging this PR would resurrect that deleted package rather than apply a live fix. Leaving it open for explicit product/maintainer disposition (close as obsolete vs resurrect the package intentionally).

ritsuKai2000 (approved):
## Engine review — approved ✅

Independent review (code-review + security, fresh subagents) plus local validation. No blocking findings.

**Validation (packages/claim-relayer):**
- `yarn test` → **44/44 pass** (7 files)
- `yarn typecheck` → exit 0
- `yarn build` → exit 0

**Correctness.** The fix correctly narrows the process-crash surface to *permanent* startup errors and keeps `start()` alive on retryable RPC failures, closing #1070. Traced the #1038/#1070 silent-wedge class: `runScheduledTick` calls `scheduleNext()` on every non-throwing path; the only path that skips it is the intentional re-throw (`!startupChecked && !isRetryableStartupError`), which becomes an unhandled rejection → Node crash → Railway ON_FAILURE restart → fresh boot fails fast via `main().catch`. That is the intended fail-fast. No path leaves the process alive with the poll loop stopped. The substring-based `isRetryableStartupError` is more robust than it looks: `buildFallbackTransport` funnels transport-level failures into `AllRpcsFailedError`, whose message reliably matches.

**Security.** Traced every error shape reaching the new `stats.lastError` (surfaced via `/status` + `/ready`): raw viem `HttpRequestError` → redacted by `errorToMessage`; `AllRpcsFailedError` → top-level message carries masked hosts only (raw URL lives on `.cause`, never read); nested `cause` chains not serialized into `.message`/`.stack`. Defense-in-depth redaction in `buildStatusPayload` + db layer. No API-key/URL leak. `retryCount: 1` *reduces* request amplification (viem default is 3). Signer key handling untouched.

**Test quality.** The four new tests genuinely exercise the surface — fake timers advanced, assertions on real state transitions (`ready` false→true, checkpoint 9n→12n), and a redaction assertion proving the secret host is stripped from the status payload. `vi.useRealTimers()` cleanup is correct.

**Advisory (non-blocking, optional follow-ups):**
1. `ready` can flip back to `true` if an in-flight `runOnce()` resolves after `stop()` — near-zero impact (stop is SIGINT/SIGTERM only). Could gate line 146 on `!this.stopped`.
2. The permanent-error crash path prints via Node's default unhandled-rejection printer rather than `errorToLogMessage`; today's permanent errors carry only addresses (no secret), but a future URL-bearing permanent error would print unredacted. A `process.on('unhandledRejection', …)` routing through `errorToLogMessage` would close that.
3. Lockstep note: `transport.ts` now adds `http(u, { retryCount: 1 })` while the sibling `client/src/rpc/transport.ts` stays on the viem default — worth a one-line comment explaining the intentional divergence.
4. Consider an alert on prolonged `ready=false` so a permanent-misconfig presenting with a transient signature (dead host → `fetch failed`, revoked key → `HTTP request failed`) doesn't retry indefinitely unnoticed — observable via `/ready` (503), unlike #1070's total silence, but relies on someone watching.

ritsuKai2000:
Closing as obsolete: `packages/claim-relayer` was removed by the tokenless-OLAS pivot (DR-2026-06-30), which deleted the JINN L2→L1 claim path this PR fixes. The target files no longer exist on `next` (hence the unresolvable conflict). If a startup-retry concern applies to a surviving component, please re-file against that component.

PR #1364, opened 2026-07-03
fix(indexer): register the SPA catch-all after all JSON API routes

### LINKED ISSUE #1363: fix(indexer): SPA catch-all shadows all JSON API routes in production (/capture-meta, /distribution-signal, /plugins, /builders/*)
## Context

`packages/indexer/src/api/index.ts` registers the explorer SPA static handler and catch-all fallback (`app.use('*', serveStatic(...))` + `app.get('*', index.html)`) at lines 70–80, **before** the JSON API routes (`/builders/*`, `/plugins`, `/distribution-signal`, `/capture-meta`) registered below it. Hono matches in registration order, so when a SPA build is present the catch-all swallows every one of those routes and serves the explorer HTML shell instead.

The gate is `hasSpaBuild = existsSync('./public/index.html')`:

- In CI and the vitest suite there is no explorer build, so `hasSpaBuild` is false, the catch-all never registers, and everything passes.
- In the production Docker image the explorer build is always present (built in the Dockerfile), so `hasSpaBuild` is true and the routes are shadowed.

Verified live on `jinn-indexer-production.up.railway.app` (2026-07-03): `/capture-meta`, `/distribution-signal`, `/plugins`, and `/builders/<addr>/artifacts` all return `<!doctype html>` (the SPA shell). The Railway deploy of #1347 (commit 07a8bcd57) itself succeeded — GitHub deployment status `success` at 10:42 UTC — so this is a code bug, not an infra failure.

## Impact

- Content-aware corpus search (#1344/#1347) is dead on the deployed indexer: `jinn-layer corpus search` gets HTML, not JSON, so it returns nothing.
- The distribution-signal route (#1314/#1336) is equally dead.
- `/plugins` and `/builders/*` (spec §6.5 REST surface) have been silently shadowed since the explorer SPA landed — no monitor covers response *shape* (indexer-monitor watches data freshness only, a known blind spot noted in `deploy/README.md`).

## Acceptance criteria

- With a SPA build present, `GET /capture-meta?q=…`, `/distribution-signal`, `/plugins`, `/builders/:addr/artifacts` reach their JSON handlers (never the SPA shell).
- The SPA still serves `/`, deep links (e.g. `/solvernet/<cid>`), and `/assets/*` exactly as before.
- A regression test reproduces the shadowing (SPA build present → API route must not return HTML) and fails if a JSON route is ever registered after the catch-all again.
- After merge to `next`, the deployed indexer answers `/capture-meta?q=tdd` with JSON containing seed records, and `jinn-layer corpus search "tdd"` returns results.

oaksprout:
Fixes #1363.

## What broke

`src/api/index.ts` registered the explorer SPA static handler + `*` fallback before the JSON API routes (`/builders/*`, `/plugins`, `/distribution-signal`, `/capture-meta`). Hono matches in registration order, so with a SPA build present the catch-all served the HTML shell for every one of those routes.

CI never caught it: tests and CI run without an explorer build, so `hasSpaBuild` is false and the catch-all never registers there — while the production Docker image always ships the build. Verified live on `jinn-indexer-production.up.railway.app`: all four route families returned `<!doctype html>`. The Railway deploy of #1347 itself succeeded (GitHub deployment status `success`), so content search was shipped but unreachable.

## The fix

- Move the SPA static + fallback block to the last registration in the file, with a comment stating the invariant.
- Regression test `test/api.spa-route-order.test.ts`: fabricates a SPA build in a temp CWD before importing the app (so the catch-all IS registered, unlike the rest of the suite), then asserts all seven JSON routes answer JSON and the SPA still serves `/`, deep links, and `/assets/*`. RED before the fix (7/10 failing), GREEN after.

## Verification

- `yarn test` — 359 passed (25 files).
- `yarn typecheck` — clean.
- Post-merge: Railway auto-deploys (`packages/indexer/**` watch path); will verify `/capture-meta?q=tdd` returns seed records on the deployed endpoint and `jinn-layer corpus search "tdd"` works end to end.


oaksprout:
Post-merge verification (Railway auto-deploys on the packages/indexer watch path):

1. `curl -s 'https://jinn-indexer-production.up.railway.app/capture-meta?q=tdd'` → JSON array containing the tdd seed (not the SPA HTML shell).
2. `curl -s 'https://jinn-indexer-production.up.railway.app/distribution-signal'` → JSON.
3. `jinn-layer corpus search "tdd"` → returns seed records.

Merge is blocked on the required-review rule (no agent self-merge); needs a maintainer approve + merge.

PR #1423, opened 2026-07-06
docs(design): distillation v1 (v0.3) + layer-2 substrate + #1409 fix

ritsukai:
## Distillation v1 — design (v0.3) + the built layer-2 substrate

Design-shape work that realises the layer-1→layer-2 promotion the harness-network spec (`spec/2026-07-02` §5) names but doesn't design. **Opened as a draft for review — this is the Phase-1 (design v0.3) gate; the Phase-2 implementation (bridge, gate, distiller, three-arm measurement) lands as follow-up PRs on this design.**

### What's in this PR

**Design (reviewable):**
- `spec/2026-07-06-distillation-v1.md` **v0.3** — v0.1 design → v0.2 (adversarial-review amendments) → v0.3 (SkillRL / training-substrate framing + failure→lessons axis + explicit build-vs-ship fork).
- `log/decisions/2026-07-06-layer-2-conformant-skill-packages.md` (DR-2026-07-06, proposed) — first-class `jinn.skill.v1` packages + two-altitude scrub + option-C consumption.
- `docs/superpowers/plans/2026-07-06-distillation-v1-plan.md` — Plan A (full TDD) + Plan B/C (interface-level, gated).

**Built & green (Plan A — the layer-2 substrate + the #1409 fix):**
- `jinn.skill.v1` conformant Agent-Skill package + provenance (`skill.ts`), `publishSkill()` (`publish-skill.ts`), the layer-2 secret-only scrub altitude (`scrub/layer2.ts`), consume-path `kind` filter (`consume.ts`).
- **Seeds re-imported as `jinn.skill.v1` at the layer-2 altitude — the root-cause fix for #1409** (`seed-import/execute.ts`). Live-checked: the real `brainstorming` seed + 2 siblings scrub byte-for-byte clean, zero placeholder tokens.
- **Indexer `skill:<cid>` registration** (`packages/indexer/src/types.ts`) — a live check found `parseEnvelopeKey` would have dropped `skill:` events, making distilled skills undiscoverable; fixed and verified (100 handler tests green).

### The v0.3 redirect (for review)
- **Positive prior:** SkillRL (arXiv 2602.08234, verified) — SOTA +15.3% via success→patterns **and** failure→lessons, hierarchical, recursive. It diagnoses why the negatives (SkillsBench/CTIM-Rover/Trace2Skill) nulled: they tested the single-shot + **success-only** + flat cell.
[...]

PR #1416, opened 2026-07-06
design: capability-eval v0 methodology + gate DR

ritsukai:
## What

The v0 **capability-eval methodology** for the Jinn harness network — *how* to credibly measure whether a corpus-connected `jinn-agent` beats the same agent stock (equal quality, lower total cost) on the coding distribution. This is the one number the whole harness-network bet reduces to (`spec/2026-07-02-jinn-harness-network.md` §8).

**Shape: `design`.** Output is the methodology spec, **not the rig**. Building the rig is a follow-on `feat` gated on sign-off of this spec. Reviewable before anything is built.

## Files

- **`spec/2026-07-06-capability-eval-v0.md`** — the methodology.
- **`log/decisions/2026-07-06-capability-eval-v0-gate.md`** — DR-2026-07-06, recording the gate threshold + the reuse-vs-rebuild / contested-band decisions.

## The gate (met/not-met-able)

> On a frozen contested-band coding slate, arm A = stock agent, arm B = same agent + distribution-matched seed loadout pre-installed (no live retrieval), pinned model, R ≥ 3 repeats:
> **PASS** iff at α = 0.05 as an **intersection-union test** — (1) quality **non-inferior** (Δ_quality > −5pp) **AND** (2) cost **strictly lower** on the both-solve set (Δ_cost < 0).
> Else **FAIL** or **INCONCLUSIVE** (underpowered — reported with its MDE; not a pass).

The IUT means **no multiplicity correction** — requiring both component nulls rejected *is* the type-I control (Berger 1982).

## Decisions settled in the design session

| Axis | Decision |
|---|---|
| Distribution | coding (SWE-rebench-V2), aligned with the distillation session |
| Corpus ON | **seeds pre-installed only** (static loadout; no live `corpus_search`) — isolates "skills in context", simplifies contamination control, extends to v1 by swapping seeds→distilled |
| Gate | non-inferior quality (δ=5pp) **AND** strictly-lower cost, IUT @ α=0.05 |
| Slate | **new power-sized contested-band** slate (stock pass-rate ~[0.15,0.85], blind to corpus). Existing N≈10 / N≈9 slates are the *underpowered* DR-2026-06-02-b artifacts — reused as tooling/format precedent only |
| Model | Haiku-class (cheapest; matches the v2 screening base) — low-ceiling external-validity threat named, Sonnet-class replication as mitigation |
| Posture | human-run measurement, **not** a CI gate |

## Load-bearing risk: contamination

[...]

PR #1427, opened 2026-07-06
feat(harness-layer): distillation three-arm measurement — §11 gate + pilot power

ritsukai:
## Phase 2 · Item 4 — three-arm measurement (§11, D9/D11)

Implements the capability gate for `distillation-v1`. Stacked on the design PR #1423 (base = `claude/elated-edison-ccb075`; retarget to `next` after #1423 merges).

**Reuses** capability-eval's rig — `comparePaired` / `mcnemarExact` (exact McNemar) from `client/src/eval/paired.ts` — this is only the analysis layer that composes them into the three-arm gate. The per-arm run (harness × `cap-v0` slate × pre-installed corpus snapshot, graded by the swe-rebench-v2 grader) is wired at run time and feeds this its per-instance results.

- **Three arms:** seeds-only (control), raw-evidence (the shipped v1 product baseline, D11), distilled (the bet).
- **distilled vs seeds** — the capability claim: resolve-rate **superiority** (trustworthy McNemar, distilled-favored) **AND cost non-inferior** on the both-solve set.
- **distilled vs raw-evidence** — the **Bitter-Lesson test (D9)**: the distiller earns its place only if it beats raw-evidence retrieval.
- **Ship verdict:** `ship-distilled` | `ship-retrieval-baseline` (D11 — retrieval ships regardless) | `inconclusive`.
- **Pilot power (§11):** `pilotPower` gates on the minimum discordant count that can reach significance, so an underpowered run returns `inconclusive` — a null is never mistaken for a decisive negative.

6 tests green (ship-distilled when it beats both + powered + cost-OK; ship-retrieval-baseline when raw ties distilled; inconclusive when underpowered; cost guard withholds ship-distilled when distilled is materially pricier); harness-layer `tsc` clean.

Refs #1416


ritsukai:
## Automated code review — three-arm measurement (§11)

Reviewed by a Claude reviewer subagent against `spec/2026-07-06-distillation-v1.md` §11 (D9/D11), range `74982827e..92ec69bcc`. **Verdict: with fixes — zero Critical.**

**Strengths.** Genuine reuse of the capability-eval rig — imports `comparePaired`/`mcnemarExact` from `paired.ts` (reviewer verified the cross-package types actually flow: breaking the call surfaced a real `TS2345`). Pilot floor is derived from exact McNemar (n=6 at α=0.05, recomputed), not hardcoded. Underpowered→`inconclusive` is enforced *first*, so a low-discordant run can't produce a false ship signal. The two-baseline gate matches D9/D11 (distilled must beat *both* seeds and raw-evidence, else the retrieval baseline ships). Honest unscorable handling inherited via `comparePaired`.

**Important findings — addressed in `999c3130e`:**
- ✅ Cost-guard fail-open closed: `baseSum === 0` with a paid treatment now returns `Infinity` (fails the guard) instead of `null` (skipped) — the seeds-only control is precisely the arm most likely to be near-zero cost, so the old branch would silently pass a paid distilled arm as "non-inferior."
- ✅ Documented that a `null` cost ratio means the guard is **skipped / cost UNVERIFIED** (not verified-cheap), so a caller treats a null ratio on `ship-distilled` as "confirm cost tracking was on."
- ✅ Documented why Wilson intervals are deliberately **not** used (matched design → McNemar; marginal Wilson reintroduces the between-instance variance the paired design removes); clarified `powered` semantics (capable of significance, not decisive).

**Process gate (cannot be code-satisfied).** §11.1 flags that the IUT composition — capability = resolve-rate superiority AND cost non-inferiority — inverts cap-v0's quality/cost roles and **must be confirmed with the capability-eval session** before a `ship-distilled` verdict is treated as final. This is now noted in the header; the module encodes the shape, it can't ratify it.

**Note.** The runtime consumer `three-arm-run.ts` lives on `claude/distill-wiring` — merge-order coupling to track.

Issue #994, opened 2026-06-03
Spike: SWE-smith as a task/reward source — new dedicated SolverNet vs. augment swe-rebench-v2


ritsuKai2000:
## Context

While scoping GEPA / `gskill` (the [blog](https://gepa-ai.github.io/gepa/blog/2026/02/18/automatically-learning-skills-for-coding-agents/) referenced at the 3b rung of discussion [#770](https://github.com/Jinn-Network/mono/discussions/770)), we decomposed `gskill` into its two independently-useful halves:

- **GEPA `optimize_anything`** → the *optimizer* (maps onto #770's policy-update / credit-assignment components — the Improve/Consolidate loop).
- **[SWE-smith](https://github.com/SWE-bench/SWE-smith)** → a *task + verifiable-test factory* (maps onto #770's **Environment** (#2) and **Reward** (#3) components — task supply + ground-truth reward), separate from the optimizer.

SWE-smith (SWE-bench team; MIT; Python `pip install swesmith`) "turns any GitHub repository into a SWE-gym" and creates *unlimited* verifiable coding tasks, each with a Docker environment + tests. It ships **250+ prebuilt repo environments** (one Docker image per repo) and **~52k task instances** on HF `SWE-bench/SWE-smith`; arbitrary repos require building a new environment via its documented workflow.

Jinn already runs SWE-bench-family tasks through the `swe-rebench-v2` SolverType (Docker eval infra). SWE-smith is a direct cousin — so the natural question is whether/how to wire it into a SolverNet as a task/reward source.

## The question to spike

How would we **add or combine SWE-smith into a SolverNet** — and specifically, should it be:

- **(a)** a **new dedicated SolverNet** seeded by SWE-smith-generated tasks, or
- **(b)** an **augmentation of the existing `swe-rebench-v2` SolverNet** (SWE-smith as an additional/alternative task source for it)?

## Impact / why it matters

- **Task supply.** Unlimited synthetic verifiable coding tasks → train-mode harness learning *without spending real fees on live tasks*, and more verdict volume for M2 learning measurement (the Haiku verdict pool is accruing from zero per the M2 re-pin, #929).
- **Reward ground truth.** Verifiable tests give a clean pass/fail evaluator signal for SWE-flavored nets.
- **Composability.** This is the "Environment + Reward" half of the GEPA direction in #770; together with GEPA-as-optimizer it is the offline train-mode skill factory ("gskill Shape A"). Worth knowing if the two halves should land separately.

[...]

PR #1482, opened 2026-07-09
design: task-creator v0 — usage-trace mining spine + minting ladder

ritsuKai2000:
## Summary

- `spec/2026-07-08-task-creator-v0.md` — design spec for the Task Creator: a new pool source that mines execution traces into evaluable `swe-rebench-v2.v1` tasks, feeding the existing admission → post → solve → grade → distill machine unchanged.
- **Spine:** usage-trace mining. The rung ladder (instruments → commit-echo → hunk-subset echo → targeted perturbation → env construction) builds the miner's components in the order they can be proven cheaply; they compose into the usage-trace miner when real non-benchmark usage arrives.
- **Answers spike #994:** fork SWE-smith's machinery (mutation, repo profiles, backtranslation), reject its dataset; augment `swe-rebench-v2` rather than a dedicated SolverNet.
- **Sequencing declared plainly:** written ahead of the cap-v0 (#1416) and three-arm distillation gates by explicit choice; only rung 0 (instruments + the mineable-trace contract) builds ahead of them.
- Four decision points for sign-off (§14): escrow funding, the mineable-trace consent tier vs. scrub posture, discrimination flag-vs-reject, rung ordering.

## Run-mode

`design` — see handbook §The shapes of work.

## Test plan

- [ ] Docs-only change; no code paths touched.
- [ ] Review decision points D1–D4 (§14) and the acceptance criteria (§13).

Made with [Cursor](https://cursor.com)


oaksprout:
## Implementation status for this design

The v0 ladder is built through rung 2 (session-echo form) across #1485 (commit-echo plumbing) + #1646 (real-usage pipeline + loop closure). Decisions D1–D6 all landed as specified; the §7 guards are wired into live paths; full AC1–7 status is in #1646's PR body.

**One deliberate deviation from §5.3, recorded here.** v0 mines sessions via **session-echo** — mint the accepted diff at its real `repo@commit` — not hunk-subset echo. The `base ⊕ S ⊖ H` broken state can't ride a `swe-rebench-v2.v1` row, because solvers never receive `test_patch` and so the row can't carry `S ⊖ H`. Session-echo gets the same loop (real usage in → gradeable task out) with no new environment machinery and blindable provenance. Hunk-subset's prepared broken state needs the env-bootstrap spike (#1640); it's deferred there, and rung 4 is renamed accordingly in the #1646 plan.

PR #1542, opened 2026-07-10
feat(harness-layer): distill --progress ndjson — machine-readable run progress events

oaksprout:
Closes #1533. Parent epic #1486. **Stacked on #1541** (the distil→distill rename) — retarget to `next` once #1541 merges.

## What

A `distill` run was a silent multi-minute blocking call — no wrapper could observe it in flight. This adds the progress surface the harness background runner (#1539) consumes:

- **Engine seam**: `distillClusters` gains optional `onPlan` (once, with every cluster + its evidence refs) and `onCluster` (start/end around each cluster's distill call, with outcome `published|rejected|error`, skill name, reason/error, duration). `createLocalDistiller` threads it through. No behaviour change when absent.
- **`--progress ndjson`**: stamped events (`v:1`, `ts`, `runId`) on **stderr**, one JSON object per line: `run_start` → `cluster_plan` → (`cluster_start` → `heartbeat`* → `cluster_end`)* → `run_end`. The heartbeat (15s, emitter-owned, unref'd) gives a watcher liveness while a frontier call is in flight. Cluster events carry the source capture's task summary as `label`.
- **`run_end` is always emitted** — empty, mode short-circuits, consent-declined, resume-nothing, partial failure, success — the single end-of-run signal for a watcher. (The `--where` setter emits nothing: it is a setter, not a run.)
- stdout `--json` is untouched; progress and result compose. Consumers must ignore stderr lines without a `v` field (shared with plain-text capture warnings). Documented in USAGE as non-interactive-only.

## Verification

- TDD: 14 new tests written first (emitter unit incl. heartbeat lifecycle; `distillClusters` seam incl. rejected/error outcomes; CLI integration incl. stdout/stderr separation, empty→`run_end`, error→`partial`, no-flag→no-events, bad format→exit 2). Full harness-layer suite: 412 passed. `tsc --noEmit`: clean.
- Live smoke (no spend): real-process stderr carries the stamped `run_end` on an empty captures dir; stdout stays a single JSON object; `--progress yaml` exits 2.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1486, opened 2026-07-09
Local distillation (rung 1) + distiller seam — MVP subproject (sub-epic)


oaksprout:
**Context.** Sub-epic for rung-1 of the harness-network product, per the 2026-07-09 design conversation. Rung 1 is **local single-player distillation**: distil your own scrubbed captures → install the produced skills → run subsequent tasks cheaper on open-weight models. The distiller sits behind one interface so the bonded-network distiller (rung 3) is a drop-in, not a rebuild (approach A: rung 1 live, rung 3 a stubbed seam).

The umbrella spec `spec/2026-07-02-jinn-harness-network.md` (v0.3) currently routes distillation **only through the network** (§5, "distillation is itself a network task"). This subproject adds the local wedge and the seam. The distillation engine already exists and produces real skills (see #1464); this is wiring a local loop and a seam onto a built engine, not building the engine.

**Impact.** Rung 1 is the single-player entry point — usable before any network / earn / steer machinery — and the seam is what makes rung 3 additive.

**Acceptance criteria.**
- [ ] All child issues closed.
- [ ] A user can distil their own local captures and run a subsequent task against the produced skills, fully locally, with no corpus publish.

**Out of scope.** Capability measurement (three-arm eval #1464 — Ritsu); earn/steer (spec v0.5/v1); the real network distiller backend (#1395 — the seam stub points at it).

**Parent.** #1307.

PR #1543, opened 2026-07-10
feat(harness-layer): per-cluster distiller subprocess timeout

oaksprout:
Closes #1534. Parent epic #1486. **Stacked on #1542** (progress events); retarget as the stack merges.

## What

- `createClaudeDistiller` / `createCodexDistiller` (+ meta variants) accept `timeoutMs` (default 600s). On expiry the child is SIGKILLed and the call rejects with `DistillTimeoutError` — `distillClusters` records it in that cluster's `errors[]` and the run continues. Timers are cleared on normal exit and unref'd, so no lingering handles.
- CLI: `--cluster-timeout <seconds>` on both `distill` (loop) and `distill run` (pipeline); env fallback `JINN_DISTILL_CLUSTER_TIMEOUT_S`; non-positive/non-numeric values exit 2.
- `ChildLike` gains optional `kill()` so pre-existing test fakes remain valid.

## Why

Required before the harness background runner (#1539) ships — an unattended hung child would otherwise burn wall-clock forever with no signal. Pairs with #1533's heartbeat: heartbeat says "still working", timeout guarantees "will end".

## Verification

TDD: 8 new tests first (hung claude/codex child killed + rejects, fast child unaffected, meta ports honour timeout; CLI flag→factory ms, env fallback, factory-default when unset, bad values exit 2). Full harness-layer suite: 420 passed. `tsc --noEmit` clean.


ritsuKai2000 (approved):
Current-head merge-batch review: mechanical conflict composition verified; 108 focused tests, typecheck/build, and all fresh CI checks passed.

PR #1544, opened 2026-07-10
feat(harness-layer): distill run log + status/runs subverbs

oaksprout:
Closes #1535. Parent epic #1486. **Stacked on #1543**; retarget as the stack merges.

## What

- **Run log** (`src/distill-runs.ts`): one JSONL record per run outcome — `ok` / `partial` / `empty` — appended best-effort at every terminal that represents an attempted run. Mode short-circuits (defer/off/consent-declined) log nothing: no run was attempted. Records share the `runId` + start clock with #1533's progress events so a watcher can join stream and history. Corrupt lines are skipped on read.
- **`distill runs [--limit N] [--json]`** — newest-first history.
- **`distill status [--captures][--out] [--json]`** — the loop's whole state in one read: mode, reserved captures (total + not-yet-distilled via the existing `coveredSessionIds` provenance scan), staged/installed counts, resolved provider+model, last run. **Pure reads** — never prompts, spends, or writes (verified: virgin-machine test asserts the mode file is not created). This is the single call the harness plugin's gating (#1537/#1538) reads.
- Subverb routing on the flat verb's first positional; unknown subcommands exit 2.

## Verification

TDD: 12 new tests first (run-log round-trip/newest-first/corrupt-tolerance/limit/absent-file/parent-mkdir; CLI: ok/partial/empty records with field parity to the `--json` result, defer short-circuit logs nothing, `runs --limit`, virgin status zeros + no-write, staged-run status reflects covered captures, unknown subverb). Full harness-layer suite: 432 passed. `tsc --noEmit` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1548, opened 2026-07-10
feat(jinn-agent): background distillation runner with ambient progress lines

oaksprout:
Closes #1539. Parent epic #1486. **Stacked on #1547**; retarget as the stack merges. Runtime dependencies: `--progress ndjson` (#1542) and — before this ships to operators — the per-cluster timeout (#1543), so an unattended hung child cannot burn wall-clock forever.

## What

- **`jinn_layer.spawn(args, stdout_path, stderr_path)`** — detached launch: file-redirected output (no pipe backpressure), `start_new_session=True` (no SIGINT propagation; POSIX guard for win32), stdin `DEVNULL`. Deliberately NOT `_default_runner` (120s cap).
- **Run state** in `$HERMES_HOME/jinn/distill/`: `current.json` while live → `last-run.json` with the terminal outcome (`ok`/`partial`/`empty`/`stopped`/`died`). The layer's run log (#1544) holds the full facts; this is the pointer record.
- **Watcher**: daemon thread tails the events file (~1s poll), filters to `v`-stamped JSON (capture warnings share stderr), maps events to ◇ lines via the plugin's stderr `_user_line` idiom (renders above the TUI input): started / `2/4 — staged "name"` / `cluster failed (kept going)` / `done — N skills staged · /jinn distill review`. Cluster lines throttle to 1 per 5s; heartbeats are liveness only. The drain step is a pure function (`_drain_events(state)`) so tests exercise it without threads.
- **`/jinn distill start`** — two-step (`/jinn consent` precedent): `start` shows the spend facts (N uncovered captures → up to N clusters, model, runs-locally note); `start confirm` records `--where local` on a first run, then spawns `distill --progress ndjson --json --resume --out $HERMES_HOME/skills` (stage-only — installing stays an explicit choice). 0 uncovered → "nothing to distill yet". Concurrent run refused on a live pid.
- **`/jinn distill stop`** — SIGTERM; staged partials kept; `--resume` covers the rest.
- **Re-attach**: session start resumes the tail of a live run from a previous process, or archives a dead one (`died`) with a recovery line.

## Verification

TDD: 13 pytest cases first — spawn args + state lifecycle, refuse-second-run, stop→archive `stopped`, event→line mapping (incl. non-JSON stderr lines and heartbeat suppression), per-type throttling with an injected clock, dead-pid→`died`, re-attach live/dead/no-op, and the full start/confirm/nothing-to-distill/already-running/stop command flows. Full jinn suite: 222 passed.


oaksprout:
Amended: `start confirm` under a recorded `defer` mode used to spawn a run that silently did nothing (the run re-read the persisted mode). It now records `--where local` for any non-local mode first — an explicit two-step start is that consent. New regression test; jinn suite 223 passed. Found during end-to-end verification.

PR #1552, opened 2026-07-10
feat(jinn-agent): payoff surface + distill discoverability

oaksprout:
Closes #1550. Closes #1551. Parent epic #1486. **Stacked on #1549** (top of the stack); retarget as the stack merges.

## What

**Payoff (#1550)** — "see the distilled skills being used":
- Session start snapshots use counts for installed skills whose `.jinn-ref` carries a `local-distill:` ref (join key = skill name, matching the upstream `.usage.json` sidecar the native skill tools bump — read-only here).
- Session end diffs and prints `◇ skill used — <name> · distilled from N of your captures` per used skill (cap 3; silence when nothing was used; independent of capture/consent — usage is the harness's own telemetry; never raises into a session end).
- One-time invitation after the first install (`payoff_invited` seen-flag): run a task like the ones the skills came from.

**Discoverability (#1551)**:
- `/jinn distill skills` — the skillset view: per-skill use count, last used, provenance count, plus the staged count → review pointer.
- `/jinn status` gains a one-line distill summary (mode, uncovered captures, staged, installed).
- One `_JINN_TIPS` entry (tips.py is a JINN.md-owned file).

## Verification

TDD: 9 pytest cases first — payoff line on usage delta, silence on none, non-distilled skills ignored (corpus refs + markerless user skills), the cap, end-to-end session-start→end emission, one-time invitation, skills view with usage, status line, tip presence. Full jinn suite: 237 passed.

### DIFF
--- a/apps/jinn-agent/hermes_cli/tips.py
+++ b/apps/jinn-agent/hermes_cli/tips.py
@@ -521,6 +521,11 @@ _JINN_TIPS = [
     and len(_jinn_rebrand_tip(_tip)) <= 150
 ]
 
+# Jinn-layer features with no upstream tip to rebrand (mono #1551).
+_JINN_TIPS.append(
+    "Distill your recent work into reusable skills, fully on this machine — /jinn distill"
+)
+
 _upstream_get_random_tip = get_random_tip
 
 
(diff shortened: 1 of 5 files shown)


oaksprout:
Amended with an e2e finding: `distill_status()` was called without `--out`, so the hub's coverage/staged/installed counts were computed against the layer's default skills dir rather than `$HERMES_HOME/skills` where the plugin installs — right after a successful run the hub still said "2 of 2 not yet distilled · 0 staged". Now scoped with `--out $(skills_dir)`; regression test added. jinn suite 239 passed. (The fix rides the top of the stack because `distill_status` is introduced in #1546 — flagging so the stack merges bottom-up as intended.)

PR #1676, opened 2026-07-14
fix(cli): only auto-open browser on jinn run's first-ever launch (#804)

### LINKED ISSUE #804: `jinn run` auto-opens a fresh browser tab on every launch (no dedup, accumulates dead tabs)
## What happens

Every `jinn run` invocation unconditionally calls `openBrowser(handshakeUrl)` (`client/src/main.ts:1398-1400`), and the handshake URL has a fresh per-launch `?k=<key>`:

```ts
process.env['JINN_UI_HANDSHAKE_URL'] =
  `http://127.0.0.1:${setupApiServer.port}/?k=${handshakeKey}`;
if (process.env['JINN_NO_UI'] !== '1') {
  openBrowser(process.env['JINN_UI_HANDSHAKE_URL']!);
}
```

`openBrowser` (`client/src/cli/open-browser.ts`) shells out to macOS `open <url>`, which always opens a new tab in the default browser — it does not dedupe against existing tabs already pointing at the dashboard.

Real operator experience after a normal dogfooding session:

> *[screenshot of 8 browser tabs all titled "jinn operator", pointing at stale handshake keys on a port the new daemon no longer owns]*

Each stale tab is silently broken — the cookie attached to the old `?k=` doesn't match the new daemon's `DAEMON_API_TOKEN` (regenerated per launch — `[main] Generated DAEMON_API_TOKEN (prefix=...)`), so the SPA quietly 401s. The operator only sees a dashboard that looks correct on the *newest* tab and stale-looking dead ones on the others. There is no signal to "use this tab, close the others."

## Why it bites M1 / dogfooding hard right now

Every restart during a debugging session (eviction loop, codex quota, indexer flakes, rebuild-and-relaunch — all the things M1 dogfooding involves) opens a fresh tab. Within an hour you have 8+ tabs that all *look* live and you can't tell which one is real. This is corroded operator confidence in a surface that's supposed to be the one boring thing.

## Fix shape (smallest)

Before calling `openBrowser`, probe whether any local browser tab is already pointing at `http://127.0.0.1:<port>/`. Two cheap heuristics:

1. **Touch-file in `earningDir`.** On first-ever launch, set `~/.jinn-client/earning/.ui-opened`. Subsequent launches skip auto-open unless the operator passes `--ui` explicitly or the touch-file is missing. Trade-off: doesn't handle the case where the operator closed all tabs.
2. **AppleScript / Chrome remote-debug poll.** Ask the system whether any browser tab matches the URL; only open if not. macOS-only; correct, but heavy.

Recommended: **(1) plus a `jinn ui` CLI** (already exists — `client/src/cli/commands/ui.ts`) for the explicit "open me a fresh tab" case. `jinn run` then defaults to suppressed once first-launch is done; `jinn ui` is the operator-driven reopen.

Alternative: short-circuit if the SPA's old cookie is valid. On launch, persist the `DAEMON_API_TOKEN` instead of regenerating per-run, so old tabs stay alive across restarts. That removes the "stale tab silently 401s" half of the problem — operators see all tabs working and can close the duplicates.

## Repro

1. `JINN_NO_UI=` unset (default).
2. `jinn run` four times in a row, killing in between.
3. Observe: four browser tabs, three with stale handshake keys. Open them — they look identical to the live one until you hit a `/v1/*` endpoint that 401s.

## Pointers

- `client/src/main.ts:1392-1400` — the unconditional `openBrowser` call
- `client/src/cli/open-browser.ts` — the launcher (no dedup)
- `client/src/cli/commands/run.ts:286-290` — the `--no-ui` → `JINN_NO_UI` translation
- `client/src/cli/commands/ui.ts` — the explicit `jinn ui` reopen command (good — keep using this)

oaksprout:
## Summary

`jinn run` unconditionally called `openBrowser(handshakeUrl)` on every launch, so every daemon restart during a dogfooding session opened another browser tab; stale tabs (dead per-launch `?k=` handshake keys) accumulate and silently 401 with no signal which tab is live (#804).

This change gates auto-open behind a first-launch marker:

- New pure decision function `client/src/cli/ui-auto-open-gate.ts` — `decideUiAutoOpen({ noUi, forceUi, markerExists })`. `JINN_NO_UI=1` wins outright; otherwise open on first-ever launch (no marker) or when forced.
- `client/src/main.ts` writes marker `<earningDir>/.ui-opened` on first launch (try/catch-guarded — a failed write never blocks startup) and, when auto-open is suppressed, logs a one-line hint pointing at `jinn ui` / `--ui`.
- New `jinn run --ui` flag (`JINN_FORCE_UI=1`) force-opens regardless of the marker. `--no-ui` still overrides everything.
- `jinn ui`, `open-browser.ts`, and UI-token persistence are untouched — the persisted `jinn_ui_token` cookie is why existing tabs keep working across restarts, and `jinn ui` remains the deliberate reopen path.

Trade-offs accepted (per issue's recommended smallest fix): if the operator closes all tabs, auto-open does not resume — `jinn ui` or `--ui` is the reopen path. Marker is scoped per `earningDir` (per operator identity). Known narrow edge: two concurrent first launches sharing an `earningDir` can each open a tab (TOCTOU on the marker); accepted as out of scope.

## Test plan

- Regression test first (TDD): `client/test/cli/ui-auto-open-gate.test.ts` — full `noUi × forceUi × markerExists` truth table (8/8 combinations), including the exact bug scenario (marker present → no auto-open).
- `client/test/cli/run-no-ui.test.ts` — `--ui` → `JINN_FORCE_UI=1` flag translation, alongside existing `--no-ui` coverage.
- `yarn typecheck`, full `yarn test`, and `yarn build` green locally.
- Independent review: APPROVE, all 7 acceptance criteria verified. Security review: no findings. Non-blocking follow-up noted by review: no integration test exercises the `main.ts` wiring block itself (pure gate + flag translation are covered).

Closes #804

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -10,9 +10,13 @@ Jinn protocol client. Runs a headless daemon that participates in the Jinn train
 Running `jinn run` opens a browser to `http://127.0.0.1:7331`
 with the operator panel — Status, Visibility (live event stream), Setup
 (bootstrap state machine), and an embedded Claude Code session in Auto Mode
-(when available).
+(when available). Auto-open only happens on the first-ever launch (tracked
+by a marker file), so restarting the daemon doesn't accumulate fresh
+browser tabs.
 
-To suppress auto-open: `jinn run --no-ui`.
+To force the panel open on a later launch: `jinn run --ui`.
+
+To suppress auto-open entirely: `jinn run --no-ui` (wins over `--ui`).
 
 To open the panel manually for an already-running daemon: `jinn ui`.
 
--- a/client/src/cli/commands/run.ts
+++ b/client/src/cli/commands/run.ts
@@ -127,7 +127,7 @@ export function createRunCommand(deps: RunDeps = PRODUCTION_DEPS): CommandModule
     name: 'run',
     summary: 'Start the daemon in the foreground; stops on SIGINT/SIGTERM',
     helpText: `Usage: jinn run [--human] [--config <path>] [--password-fd <fd>] [--no-ui]
-                [--no-daemon] [--funding-timeout <duration>] [--json-progress]
+                [--ui] [--no-daemon] [--funding-timeout <duration>] [--json-progress]
 
 Long-running. Starts the creator, harness, and delivery-watcher
 loops and runs until the process receives SIGINT or SIGTERM. Before
@@ -138,12 +138,18 @@ By default, stdout emits a single machine-readable startup record and
 all progress / runtime logs go to stderr. Use \`--human\` for a concise
 terminal summary instead.
 
-By default, the operator panel is opened in your default browser once the
-daemon's API server is up. Pass \`--no-ui\` to suppress this.
+The operator panel is only auto-opened in your default browser on the
+first-ever launch (tracked by a marker file), so restarting the daemon
+doesn't accumulate fresh browser tabs. Use \`jinn ui\` to reopen it any
+time. Pass \`--ui\` to force a browser open on this launch even after the
+first; pass \`--no-ui\` to suppress auto-open entirely — \`--no-ui\` wins
+if both are given.
 
 Flags:
   --human                Print a concise terminal summary instead of JSON.
-  --no-ui                Suppress automatic browser open (default: open the operator panel).
+  --no-ui                Suppress automatic browser open. Overrides --ui.
+  --ui                   Force the operator panel open even after the first
+                         launch (normally auto-open only happens once).
   --no-daemon            Stop after bootstrap completes; do not start daemon
                          loops. Emits a JSON summary on stdout and exits 0.
   --funding-timeout <d>  Bound the wait when the wallet needs funding. Accepts
@@ -156,6 +162,7 @@ Examples:
   jinn run
   jinn run --human
   jinn run --no-ui
+  jinn run --ui
   jinn run --no-daemon --json
   jinn run --funding-timeout 30m --json-progress
   printf '%s\n' secret | jinn run --password-fd 0
@@ -174,6 +181,7 @@ Failure example (funding gate):
           options: {
             ...COMMON_FLAGS,
             'no-ui': { type: 'boolean', default: false },
+            ui: { type: 'boolean', default: false },
             'no-daemon': { type: 'boolean', default: false },
             'funding-timeout': { type: 'string' },
             'json-progress': { type: 'boolean', default: false },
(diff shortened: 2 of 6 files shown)


ritsuKai2000:
The red here ("Task Creator AC1 (amd64 gold proof)") is not this PR's fault — it's the HuggingFace-outage CI flake fixed in #1683 (merged to next as ba8062431, committed HF fixture). This branch's last commit (14:38) predates that merge. Rebase onto origin/next to inherit the fixture and the check will clear; the actual #804 change (browser-tab dedup) looks unaffected.

ritsuKai2000 (approved):
## Engine review: APPROVE

Independent review of PR #1676 (`fix/804` — first-launch UI auto-open gate). Two review subagents (code-review + security), both clean; no blocking findings.

**Code review — nothing blocks.** Verified independently:
- `decideUiAutoOpen` gate logic correct — `noUi` short-circuits first (`--no-ui` wins over `--ui`); `forceUi || !markerExists` / `!markerExists` are the right predicates; full 8-combination truth table covered by tests.
- `main.ts` wiring (lines 994–1029): `JINN_UI_HANDSHAKE_URL` set before `openBrowser(...)`, so the non-null assertion is safe; marker-write is try/catch-guarded so a read-only/full `earningDir` degrades to "reopens every launch" rather than crashing startup; `mkdirSync(dirname, {recursive})` handles a not-yet-created `earningDir`; the `else if (!noUi)` hint branch fires only on a suppressed non-headless restart.
- New hint `console.log` routes to stderr via `run.ts`'s console override, so the machine-readable `--json` startup record is not polluted.
- `run.ts` `--ui` → `JINN_FORCE_UI=1` translation is symmetric with the existing `JINN_NO_UI` block; tests snapshot/restore the env key.
- `yarn typecheck` exit 0; `ui-auto-open-gate.test.ts` + `run-no-ui.test.ts` → 21/21 pass. American English, surgical, simplicity-first respected.

**Security — no findings.** Marker path is `join(earningDir, '.ui-opened')` (trusted config/env + hardcoded basename, no traversal); content is a bare ISO timestamp; the per-launch `?k=` handshake key handling is byte-identical (freshly `cryptoRandomBytes(16)` each run, never persisted/reused — the marker holds only a timestamp); no injection/SSRF/local-priv surface (opened URL is a fixed `127.0.0.1:<port>` loopback literal).

Advisory/nits only (non-blocking): `process.env` translation is process-global like the pre-existing `JINN_NO_UI` pattern; TOCTOU on concurrent first-launches sharing an `earningDir` is acknowledged out-of-scope in the PR body.

Approving and un-drafting.

ritsuKai2000:
Merge-batch checkpoint (2026-07-15): conflict-free rebase from `f9b5d398` onto current `origin/next` `32cf314d`; new head `6670a165`. All three commits are `=` in range-diff. Focused first-launch/browser tests passed 21/21; client typecheck and build passed under Node 22.22.2. Exact force-with-lease push succeeded and remote/local heads match. Fresh CI is running; the prior approval is stale under this batch's current-head review rule.

PR #1680, opened 2026-07-14
fix(daemon): successful on-chain deliveries persisted as FAILED due to desired_state_id NOT NULL constraint

### LINKED ISSUE #506: fix(daemon): successful on-chain deliveries persisted as FAILED due to desired_state_id NOT NULL constraint
## Context

The operator dashboard activity table shows four solve runs from today as **FAILED** even though their delivery txs landed on-chain with status \`0x1\`:

- \`0x5aa25e2a\`
- \`0x0006bc8f\`
- \`0xbbfa8346\`
- \`0x36ea8435\`

These are real successes. The state badge is lying.

## Root cause (per parallel analysis)

The daemon's SQLite persistence path for completed runs has a \`desired_state_id NOT NULL\` constraint that fails when the run has no associated desired-state row (e.g. SWE-rebench v2 solve runs whose desired-state record was never persisted, or was cleared before the post-delivery write). The persist call throws, the run's state never transitions to \`COMPLETE\`, and the engine fallback flips it to \`FAILED\`.

## Impact

- Operators cannot tell which runs actually succeeded from the dashboard.
- Inflates the perceived failure rate and obscures real failures.
- Likely also breaks any downstream consumer that reads run state (reward attribution sanity checks, success-rate metrics, eventual outcome column from #502).

## Acceptance

- [ ] Repro: write a Vitest test that drives a solve run through delivery without a desired-state row and asserts the persisted state is \`COMPLETE\`, not \`FAILED\`.
- [ ] Fix: either (a) make \`desired_state_id\` nullable on the relevant SQLite table and update the schema migration, or (b) ensure the desired-state row exists before the post-delivery write — whichever matches the underlying data model.
- [ ] Backfill: a one-shot script that re-classifies historical rows where \`state = FAILED\` but a successful delivery tx exists on-chain for the same \`requestId\`. (Or document that historical rows stay wrong and only new runs are correct.)
- [ ] Manual verify against the running operator: the four txs listed above show as \`succeeded\` after re-bootstrap.

## Out of scope

- The over-broad \`FAILED\` bucket for evaluate-already-taken / solve-killed-by-restart rows — separate issue.
- The outcome column (did the patch pass SWE tests) — tracked in #502.
- Reward-claim liveness — unrelated, tracked in #505.

## Reference

- Operator session 2026-05-22 (Oak's running daemon on Base Sepolia, SWE-rebench v2).
- Persistence path: \`client/src/store/store.ts\` (the schema with \`desired_state_id NOT NULL\`) — entry point likely the post-delivery write in \`client/src/daemon/delivery-watcher.ts\` or the harness engine completion handler.

oaksprout:
## Summary

Issue #506: solve runs whose delivery landed on-chain (status 0x1) showed as FAILED in the operator dashboard. Root cause: on SQLite databases created before the artifacts-schema change, `insertArtifact()` threw `NOT NULL constraint failed: artifacts.desired_state_id` during the post-delivery write, and the engine's error fallback flipped the just-completed run to FAILED.

The schema-level production fix already landed on `next` in 65de316e8 (PR #511): `Store.hasLegacyDesiredStateId` detection + mirroring `task_id` into the legacy column. This PR delivers the remaining acceptance criteria:

- **Regression tests** (previously missing): a store-level test that `insertArtifact()` succeeds against a legacy `desired_state_id NOT NULL` schema, and an engine-level test driving the full `deliver()` walk on a legacy-schema DB asserting the run lands COMPLETE, not FAILED. Both were empirically confirmed to fail with the exact original error when the 65de316e8 fix is reverted.
- **Backfill**: `jinn backfill-failed-deliveries` — a one-shot, idempotent CLI command that re-classifies historical FAILED rows to COMPLETE. Scoped tightly to this bug's signature: a row is only a candidate when its `failure_reason` contains `artifacts.desired_state_id` (covers the raw and `recovery: `-prefixed forms), and its `deliveryTxHash` receipt is independently re-verified as `status: success`. The scoping matters because the delivery flow persists the tx hash after `deliverToMarketplace` but before `claimDelivery` — a receipt-only condition would wrongly reclassify genuine claim-step failures. Supports `--dry-run`; reports `reclassified` (with each row's original failure reason, preserving the audit trail), `skipped` (with reasons), and `failed` (per-row RPC errors, isolated).
- **Known caveat (documented in CLI help + module docblock)**: reclassified rows have no `artifacts` row and no ERC-8004 anchor — the artifact write is exactly what failed for them; the backfill does not reconstruct either.

## Test plan

- `client/test/store/store.test.ts` — insertArtifact regression on legacy schema.
- `client/test/harnesses/engine/engine-packaging.test.ts` — end-to-end deliver() regression on legacy schema (COMPLETE, delivery tx hash set, restoration-result artifact present).
- `client/test/harnesses/engine/backfill-failed-deliveries.test.ts` — 10 cases: success reclassify, no-hash skip, reverted-receipt skip, non-matching-signature skip (claim-step failure shape), recovery-prefix match, RPC-error isolation with batch continuation, COMPLETE rows untouched, dry-run no-write, idempotency, concurrent-state-change skip.
- `client/test/cli/commands/backfill-failed-deliveries.test.ts` — JSON envelope, counts, --human rendering, fatal envelope, --dry-run forwarding.
- Full suite: `yarn typecheck` clean; `yarn test` green (715 files / 6189 tests); `yarn build` clean.

## Remaining acceptance item (operator-side)

Manual verify against the running operator (the four txs from the issue showing `succeeded` after re-bootstrap) requires the operator's live daemon and DB — run `jinn backfill-failed-deliveries --dry-run` first, review the candidate list, then run without the flag and re-check the dashboard. Not automatable from this PR.

Closes #506

### DIFF
--- a/client/src/cli/commands/backfill-failed-deliveries.ts
+++ b/client/src/cli/commands/backfill-failed-deliveries.ts
@@ -7,16 +7,23 @@
  * the legacy `artifacts.desired_state_id NOT NULL` column (fixed in #511,
  * commit 65de316e8). The engine's error-classification catch turned that
  * throw into a FAILED run even though the on-chain delivery had already
- * landed. This command scans historical FAILED rows, checks each recorded
- * `deliveryTxHash`'s receipt, and reclassifies the ones whose delivery
- * succeeded. No wallet/password is required — this only reads chain state
- * and writes to the local SQLite DB.
+ * landed. This command scans historical FAILED rows whose `failureReason`
+ * carries the `artifacts.desired_state_id` constraint signature, checks
+ * each candidate's recorded `deliveryTxHash` receipt as secondary
+ * confirmation, and reclassifies the ones whose delivery succeeded.
+ * `deliveryTxHash` + a successful receipt alone is NOT sufficient grounds —
+ * see the scoping note in the harness module below. No wallet/password is
+ * required — this only reads chain state and writes to the local SQLite DB.
+ *
+ * Caveat: reclassified rows have no `artifacts` row and no ERC-8004
+ * metadata anchor (that write is exactly what failed) — this command does
+ * not reconstruct either, it only corrects `task_runs.state`.
  *
  * See `client/src/harnesses/engine/backfill-failed-deliveries.ts` and
  * `client/src/harnesses/engine/persistence.ts#reclassifyFailedAsComplete`.
  */
 
-import type { CommandContext, CommandModule } from '../command.js';
+import type { BaseCommandDeps, CommandContext, CommandModule } from '../command.js';
 import { COMMON_FLAGS, parseCommandArgs } from '../command.js';
 import { emitResult } from '../output.js';
 import { emitEnvelope } from '../../errors/envelope.js';
@@ -41,8 +48,8 @@ function humanResult(payload: BackfillFailedDeliveriesResult & { dryRun: boolean
     payload.dryRun ? 'Failed-delivery backfill (dry run) complete.' : 'Failed-delivery backfill complete.',
   ];
   lines.push(`  reclassified: ${payload.reclassified.length}`);
-  for (const requestId of payload.reclassified) {
-    lines.push(`    ${requestId}`);
+  for (const r of payload.reclassified) {
+    lines.push(`    ${r.requestId} — was: ${r.originalFailureReason ?? '(no failure reason recorded)'}`);
   }
   lines.push(`  skipped:      ${payload.skipped.length}`);
   for (const s of payload.skipped) {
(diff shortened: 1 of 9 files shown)


ritsuKai2000:
Merge-batch checkpoint (2026-07-15): conflict-free rebase from `69979647` onto current `origin/next` `32cf314d`; new head `abdcaa97`. All four commits are `=` in range-diff. Focused delivery persistence/backfill tests passed 51/51; client typecheck and build passed under Node 22 after an immutable dependency refresh. Exact force-with-lease push succeeded; remote/local heads match and the worktree is clean. Fresh CI and current-head review are required before draft conversion.

Issue #649, opened 2026-05-26
Concurrent jinn run invocations corrupt pidfile + daemon state before port-bind mutex fires


ritsukai:
**Context.** A second `jinn run` invocation that races against an already-running daemon currently pollutes shared on-disk state before being blocked by the API port bind. Two contributing bugs:

1. `client/src/main.ts:2557-2559` writes `~/.jinn-client/earning/daemon.pid` **unconditionally** with `writeFileSync(pidPath, String(process.pid), 'utf-8')` — no liveness check on any existing pidfile. So the racing process clobbers the original daemon's PID in the pidfile.
2. `client/src/daemon/daemon.ts:259-286` `Daemon.start()` mutates shared store state — `setShutdownState('running')`, `setDaemonStartedAt(now)`, and emits the `startup` activity event — **before** calling `startApiServer({ port })`. The port bind (`EADDRINUSE`) is the real mutex, but by the time it fires the bookkeeping is already corrupted.

Observed in production overnight 2026-05-25 → 2026-05-26: original daemon PID 79798 was running; at 2026-05-26T03:36:54 UTC something invoked a second `jinn run`; the racing process wrote PID 31255 to `daemon.pid`, emitted activity event id 2157 (`kind: 'startup'`), overwrote `daemon_started_at`, then failed to bind port 7332 (`EADDRINUSE`) and exited without cleaning the pidfile. PID 79798 was unaware and kept running. `MechAdapter.initialize()` runs before the port bind too but is read-only (verified — `getBlockNumber`, cursor sets in memory, `loadPendingEvaluationSolutions`, `recoverPendingState` — no on-chain writes), so no funds/tx side effect from the race today.

**Impact.**

- **`jinn stop` is silently broken when this race has occurred.** `stop.ts` reads `daemon.pid`, SIGTERMs the PID, catches ESRCH if the PID is dead, and reports `state: 'stopped'` while cleaning the pidfile (`stop.ts:131-138`, `140-156`). After the race, `daemon.pid` points to the dead racing PID, not the live daemon — so `jinn stop` SIGTERMs a corpse, reports success, and leaves the real daemon running. Operator must `kill <pid>` directly.
- **Observability lies.** `/v1/status` reports `daemon.startedAt = <racing process timestamp>` instead of the real daemon's start time. Spurious `startup` activity row distorts uptime/restart analyses.
[...]

PR #1685, opened 2026-07-14
fix(cli): recover from orphaned daemons — reclaim recycled pids, jinn stop fallback, new jinn kill

### LINKED ISSUE #805: No `jinn kill` / `jinn run` doesn't recover from orphaned daemons; stale pidfiles silently overwritten
## What happens

There are three holes in the daemon lifecycle controls that bite the operator together when something goes wrong mid-session:

### 1. `jinn run` never checks for an existing instance

`client/src/main.ts:2576-2578` overwrites `~/.jinn-client/earning/daemon.pid` unconditionally:

```ts
const pidPath = join(config.earningDir, 'daemon.pid');
writeFileSync(pidPath, String(process.pid) + '\n', 'utf-8');
```

No precheck for "is there a process with the old pid still alive." If the prior daemon crashed without removing the pidfile, the new pid is written, the orphan is now untrackable via `jinn stop`, and the operator has to `ps aux | grep jinn` and kill by hand.

(Port-binding *is* checked — `EADDRINUSE` aborts the launch — but only if the orphan still owns the port. If the orphan crashed mid-shutdown, port may be free but the process may still be alive doing nothing useful, e.g. blocked on a `waitForTransactionReceipt` call.)

### 2. `jinn stop` only stops the daemon recorded in the pidfile

Useful for the common case, useless when (a) the pidfile is stale, (b) multiple daemons accidentally launched (different ports, different `JINN_EARNING_DIR`), or (c) a previous `jinn run` from a different worktree wrote its own pidfile and the operator forgot.

### 3. No `jinn kill` / `jinn kill --all` surface

There is no CLI command to "kill any/all jinn processes on this machine." The operator's escape hatch is:

```
ps aux | grep -E "[j]inn|dist/bin/jinn" | awk '{print $2}' | xargs kill
```

Which is exactly what the daemon's own lifecycle tooling should provide.

## Why it bites M1 / dogfooding right now

Dogfooding loops (eviction loop debugging, codex quota issues, rebuild-relaunch cycles) hit all three holes routinely:

- Daemon hangs mid-shutdown (e.g. waiting on an RPC) → SIGTERM doesn't take, operator force-quits the terminal, pidfile orphaned
- Operator runs `jinn run` from a different shell to test, doesn't realise old one is still alive, port-binding error → looks like a config problem, not a lifecycle problem
- Multiple worktrees, each launching daemons against the same earningDir → pidfile races

## Fix shape

1. **`jinn run` precheck.** Read pidfile; if pid is alive AND it's a `node dist/bin/jinn.js run` (cmdline check), refuse to start with a clear message and `jinn stop` / `jinn kill` hint. If pid is alive but not jinn, clear the pidfile and proceed (defensive).
2. **`jinn kill` CLI.** New subcommand: enumerate jinn processes by cmdline match (`ps -eo pid,command | grep "dist/bin/jinn.js run"`), SIGTERM with 10s timeout, then SIGKILL. `--all` flag for the multi-instance case.
3. **`jinn stop` fallback.** If pidfile is stale or missing, fall through to the `jinn kill` enumeration before giving up.

## Repro

Force the orphan path:

1. `jinn run` in shell A.
2. In shell B: `kill -STOP $(cat ~/.jinn-client/earning/daemon.pid)` (suspend without removing pidfile).
3. `jinn run` in shell C: writes its own pid to the pidfile, port-binding fails, orphan still alive and suspended.
4. `jinn stop`: stops shell C's pid (the new one), can't find the suspended orphan.

## Pointers

- `client/src/main.ts:2575-2586` — pidfile write (unconditional)
- `client/src/cli/commands/stop.ts:21-28` — `processAlive(pid)` helper (good — reuse for precheck)
- `client/src/cli/commands/run.ts` — where the precheck should live
- `client/src/main.ts:2629-2647` — EADDRINUSE error envelope (only partial coverage)

oaksprout:
## Summary
- `jinn run` precheck (`client/src/preflight/pidfile-liveness.ts`) now verifies the recorded live pid's cmdline via a tri-state probe: a live pid that is definitively not a jinn daemon (recycled pid) reclaims the pidfile and proceeds; a live jinn daemon refuses with a hint to `jinn stop` / `jinn kill`; an undeterminable cmdline fails closed (refuses), preserving the #649 double-daemon protection. Covers the tsx dev-mode invocation (`yarn jinn run`) as well as the built `dist/bin/jinn.js` and published `jinn` shim shapes.
- `jinn stop` (`client/src/cli/commands/stop.ts`) identity-checks the recorded pid before SIGTERM, and when the pidfile is missing/malformed/stale falls through to host-wide cmdline enumeration: exactly one discovered jinn process is SIGTERMed (reported in additive `discoveredPids`); multiple matches are reported as `ambiguousPids` with a `jinn kill --all` hint instead of blind mass-signaling (multi-worktree hosts are a supported pattern).
- New `jinn kill [--all]` (`client/src/cli/commands/kill.ts`): enumerates jinn daemon processes by cmdline, SIGTERM, polls up to 10s, re-verifies identity, then SIGKILLs survivors. Refuses more than one match without `--all`. DI factory pattern for testability. Registered in the CLI, documented in helpText, `spec/2026-04-14-client-surface.md` §2.1, `client/README.md`, and regenerated `SKILL.md`.
- Shared module `client/src/lifecycle/process-discovery.ts` (`enumerateJinnProcesses`, `pidMatchesJinn`, `processAlive`) backs all three; `ps -ww` prevents cmdline truncation; injectable execSync for deterministic tests.
- Implementation plan: `docs/superpowers/plans/2026-07-14-jinn-lifecycle-orphan-recovery.md`.

## Test plan
- Regression-test-first throughout (fix shape): new suites `client/test/lifecycle/process-discovery.test.ts`, `client/test/cli/commands/kill.test.ts`; extended `client/test/cli/commands/stop.test.ts`, `client/test/preflight/pidfile-liveness.test.ts`, `client/test/main/concurrent-jinn-run-refused.test.ts`.
- `yarn typecheck` — exit 0, zero errors.
- `yarn test` — 715 test files passed, 7 skipped (722 total); 6201 tests passed, 24 skipped (6225 total). No EADDRINUSE flake on this run.
- `yarn build` — exit 0.
- `yarn skill:check` — "SKILL.md is up to date", exit 0.
- Manual smoke: `node dist/bin/jinn.js kill --help` / `stop --help` render the new help text.

## Review notes
- Pipeline: code-review pass (9 findings, 7 fixed in-branch, see commits), independent review (1 blocking finding — tsx dev-mode cmdline gap — fixed and re-approved), security review (clean).
- Known follow-up (pre-existing, out of scope): `client/src/mcp/operator-server.ts` `startDetachedDaemon` has the same recycled-pid bug class this PR fixes elsewhere; a follow-up issue will be filed.
- `jinn kill --all` terminates matches sequentially (~N x 10s worst case); accepted trade-off, N is tiny in practice.

Closes #805

### DIFF
--- a/client/README.md
+++ b/client/README.md
@@ -268,6 +268,7 @@ docker run --rm ghcr.io/jinn-network/client:latest version --json
 | `jinn bootstrap` | Advance toward a running fleet | Yes |
 | `jinn fund-requirements` | List what needs funding | Yes |
 | `jinn stop` | Signal a running daemon to stop | Yes |
+| `jinn kill` | Force-terminate jinn daemon processes by cmdline enumeration (`--all` for multiple) | Yes |
 | `jinn version` | Version, phase, deployment digest | Yes |
 | `jinn update` | Update the client package + refresh integrations | Yes |
 | `jinn integrations install` | Wire Jinn into Claude Code / Codex / other AI tools | Yes |
--- a/client/skills/jinn-operator/SKILL.md
+++ b/client/skills/jinn-operator/SKILL.md
@@ -54,28 +54,37 @@ This gives you the `jinn` operator CLI. The built-in MCP server is invoked via `
 | `jinn version` | Print client version, protocol phase, and resolved token map |
 | `jinn doctor` | Preflight checks: answers "would jinn run work?" without running it |
 | `jinn init` | Generate the master wallet and write the encrypted keystore |
+| `jinn quickstart` | Legacy compatibility alias for zero-to-running setup; prefer jinn run |
+| `jinn auth` | Legacy compatibility: check Claude authentication and persist daemon runtime mode |
 | `jinn bootstrap` | Advance the fleet state machine toward a running daemon |
 | `jinn fund-requirements` | List addresses that need funding before the next bootstrap step |
-| `jinn run` | Zero-to-running in one command: init, fund, bootstrap, then start the daemon in the foreground (stops on SIGINT/SIGTERM) |
+| `jinn run` | Start the daemon in the foreground; stops on SIGINT/SIGTERM |
 | `jinn stop` | Signal a running jinn daemon to shut down gracefully |
+| `jinn kill` | Force-terminate jinn daemon processes discovered by cmdline enumeration |
 | `jinn status` | Daemon liveness + roll-up (poll this for monitoring; pull detail separately) |
 | `jinn fleet` | Per-service fleet detail (wallets, staking, rewards, attention) |
 | `jinn balance` | Flat per-wallet balance map across master and service wallets |
 | `jinn history` | Recent protocol activity (Tasks, claims, deliveries, evaluations, rewards) |
-| `jinn rewards` | Earned vs claimed per service, per asset; next checkpoint time |
+| `jinn rewards` | Staking collector queue per service; next checkpoint time |
 | `jinn logs` | Structured event log (one JSON object per line) |
-| `jinn tasks` | Submit, list, and inspect Tasks |
-| `jinn solver-nets` | Enable SolverNets, select Harnesses, and attach SolverNet-scoped SolverPlugins |
-| `jinn harnesses` | List, inspect, add, and remove Harnesses |
-| `jinn solver-plugins` | Validate, inspect, and pack SolverPlugin packages |
-| `jinn integrations` | Configure host AI tools to use the Jinn MCP server and operator skill |
 | `jinn claim-rewards` | Pull pending protocol rewards to the fleet multisigs |
 | `jinn withdraw` | Sweep master / agents per withdraw flags |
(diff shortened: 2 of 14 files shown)


ritsuKai2000:
Merge-batch checkpoint (2026-07-15): conflict-free rebase from `a79d1aee` onto current `origin/next` `32cf314d`; new head `bf5b913b`. Seven commits remain and all seven are `=` in range-diff. Focused lifecycle/CLI tests passed 45/45; client typecheck and build passed under Node 22.22.2. Exact force-with-lease push succeeded and fresh CI/hermetic runs are active; downstream AC1 must also pass. Current-head review is required before draft conversion.

PR #1703, opened 2026-07-14
fix(daemon): gate evaluation-opportunity scan on evaluator role

### LINKED ISSUE #547: Daemon iterates the full historical evaluation-opportunity backlog every cycle, even when not joined as evaluator
## Context

While debugging T3.1 op-b on 2026-05-23 we noticed op-b's log shows a long sequence of:

```
[mech] skipping evaluation opportunity 0x… for task 156/0: TCAttemptAlreadyFinalized(156, 0) (terminal — pruned)
[mech] skipping evaluation opportunity 0x… for task 180/0: TCEvaluationDeadlinePassed(180) (terminal — pruned)
…
```

op-b's joined-SolverNet entry has `roles: ["solver"]` — it should never look at evaluation opportunities. But on every cycle the daemon iterates the full historical list of finalized/expired evaluation tasks just to call `canEvaluate` (or equivalent) and prune them. The pruning is correct, but the work is wasted: the operator's role doesn't permit it to evaluate, so the iteration produces nothing actionable.

## Impact

- Wasted RPC budget (`canEvaluate` and friends fan out to chain reads per task).
- Slower poll cycles, especially as the historical backlog grows.
- Contributes to substrate RPC pressure — exactly the kind of load that took the indexer down on 2026-05-20 (see `docs/release/v2026.05.25/handoff.md` §Session 2 bug-hunt).

## Acceptance

- Before fanning out evaluation-opportunity scans, the daemon filters by the operator's role in the SolverNet. If `evaluator` isn't in `roles`, skip the scan entirely for that SolverNet.
- A small unit test fixture sets `roles: ["solver"]` and asserts `fetchEvaluationOpportunities` is never called.

## Provenance

Surfaced during v2026.05.25 release-readiness drive (2026-05-23). See `docs/release/v2026.05.25/handoff.md` §Deferrable follow-up 6.

ritsukai:
## Summary

The daemon iterated the full historical evaluation-opportunity backlog on **every** poll cycle even for operators whose joined-SolverNet roles are `["solver"]` and never `evaluator`. Each opportunity fanned out chain reads (`canEvaluate` and friends) that could never produce an actionable claim — wasted RPC budget and added substrate RPC pressure (the class of load that took the indexer down on 2026-05-20).

This gates all three evaluation surfaces behind a single `evaluatorEnabled` flag on `MechAdapter`:

- **Boot rehydrate** — `loadPendingEvaluationSolutions()` is skipped for non-evaluators. `recoverPendingState` (own in-flight *restoration* claims via router cursor / TaskCreated scan) stays unguarded.
- **Per-cycle ingest** — delivery-claimed logs are only decoded into the pending-evaluation set when `evaluatorEnabled`. Restoration discovery is untouched.
- **Per-cycle scan** — `retryPendingEvaluationSolutions()` returns early for non-evaluators (covers both `watchForTasks` call sites via the one guard).

`MechAdapterConfig.evaluatorEnabled` is **opt-out** (undefined ⇒ enabled) so bare/CLI construction sites keep historical behaviour; production callers (`main.ts`) pass an explicit boolean derived from `joinedSolverNets` roles. A live evaluator-role join flips the gate on via `setEvaluatorEnabled(true)`, wired through the join-applier's new `enableEvaluator` dep. A join never removes a role, so no "off" path exists.

## Test plan

- `client/test/adapters/mech/adapter.test.ts` — solver-only (`evaluatorEnabled: false`) asserts `canClaimEvaluation` is never called and `pendingEvaluationSolutions` stays empty (the acceptance criterion); a paired evaluator (`true`) test asserts a real `evaluation:…` announcement is yielded so the gate is load-bearing, not a no-op; a boot-rehydrate test asserts persisted opportunities are not rehydrated for non-evaluators.
- `client/test/daemon/join-applier.test.ts` — asserts `enableEvaluator` fires for an evaluator-role join and never for a solver-only join (closes a review-flagged coverage gap on the live-flip mechanism).
- `yarn typecheck` green; both test files pass (58 + 5).

## Reviewer notes

- Independent review: **approve**, nothing blocking. Traced the second `new MechAdapter` site (`cli/execution-context.ts`) — it drives `jinn tasks submit` only and never calls `watchForTasks`, so its default-enabled flag is inert.
- Security review: **no security-relevant issues** — this is an internal behavior gate, not an auth boundary; self-eval prevention and claim eligibility remain enforced on-chain.
- Known functional edge (out of scope, fails **closed**): the request-block cursor advances while the gate is disabled, so an operator that turns evaluator on *live* mid-run will not rehydrate solutions delivered during the disabled window. It fails closed (fewer opportunities, never unauthorized) and the common case (evaluator role present at boot) is unaffected.

Closes #547

### DIFF
--- a/client/src/adapters/mech/adapter.ts
+++ b/client/src/adapters/mech/adapter.ts
@@ -248,6 +248,23 @@ export class MechAdapter implements ExecutionAdapter {
     this.store = store;
   }
 
+  /**
+   * Whether this operator participates as an evaluator. Undefined ⇒ enabled
+   * (opt-out default). Gates evaluation-opportunity ingest, boot rehydrate, and
+   * per-cycle scan (#547).
+   */
+  private get evaluatorEnabled(): boolean {
+    return this.config.evaluatorEnabled !== false;
+  }
+
+  /**
+   * Enable the evaluator role at runtime after a live SolverNet join (a join
+   * never removes a role, so turning it back off is never needed). #547.
+   */
+  public setEvaluatorEnabled(enabled: boolean): void {
+    this.config.evaluatorEnabled = enabled;
+  }
+
   async initialize(): Promise<void> {
     const chain = this.config.chainId === 84532 ? baseSepolia : base;
     const clients = createClients(
@@ -278,7 +295,13 @@ export class MechAdapter implements ExecutionAdapter {
 
     // Recover pending state from on-chain events
     if (this.store) {
-      this.loadPendingEvaluationSolutions();
+      // #547: only rehydrate the evaluation-opportunity set for evaluators.
+      // recoverPendingState recovers this operator's own in-flight restoration
+      // claims (router cursor + TaskCreated scan), not the evaluation set, so it
+      // stays unguarded.
+      if (this.evaluatorEnabled) {
+        this.loadPendingEvaluationSolutions();
+      }
       await this.recoverPendingState(blockNumber);
     } else {
       const fromBlock = this.onchainScanFromBlock(blockNumber);
@@ -1046,6 +1069,7 @@ export class MechAdapter implements ExecutionAdapter {
   }
 
   private async *retryPendingEvaluationSolutions(): AsyncIterable<TaskAnnouncement> {
+    if (!this.evaluatorEnabled) return; // #547: non-evaluators never scan.
     let processed = 0;
     for (const [requestId, solution] of Array.from(this.pendingEvaluationSolutions)) {
       // Yield to the event loop periodically so a large backlog of pending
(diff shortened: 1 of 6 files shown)


ritsuKai2000:
Merge-batch checkpoint (2026-07-15): conflict-free rebase from `576e626a` onto current `origin/next` `32cf314d`; new head `39595847`. Both commits are `=` in range-diff. Focused daemon/evaluator-role tests passed 63/63; client typecheck and build passed under Node 22.22.2. Exact force-with-lease push succeeded and fresh CI/hermetic runs are active. Current-head review is required before draft conversion.

PR #1715, opened 2026-07-14
feat(harness-layer): adapter shims for the five plugin ports (#1660)

ritsukai:
## Summary
Stage 1 of the Jinn Plugin port (S1-E1): five additive adapter shims in `client/packages/harness-layer/src/adapters/` implementing the five `@jinn-network/plugin` ports over existing harness-layer machinery — no files moved, dependency arrow one-way (harness-layer imports the plugin contracts, never the reverse).

- **CorpusPort** → `consume.ts` `createHarnessLayer().corpus.search/get`, mapping `CorpusSearchHit`/`CorpusRecord` → `KnowledgeHit`.
- **EvidencePort** → a captures-dir file store: reads legacy `CapturedTask` (up-mapped via `capturedTaskToEpisode`, fail-closed `EpisodeV1Schema.parse`) and writes/reads `EpisodeV1` (`.episode.json` suffix so the two never collide). `episodeId` is guarded against path traversal at the filename boundary.
- **ContributionPort** → an adapter-owned file-backed `ContributionStatusStore` (queued→vetoed lifecycle); no `MineableTraceStore`/`client/src` import (the Stage-1 package split forbids the backwards dependency).
- **LocalLearningPort** → `createLocalDistiller` (rung-1 distiller pipeline).
- **SkillsPort** → a skills-install index.

Wiring mirrors the `@jinn-network/sdk` precedent: `client/package.json` gains `"@jinn-network/plugin": "portal:../packages/plugin"` and a `build:plugin` script chained into `typecheck`/`test`/`build`. `client/vitest.config.ts` already collects `packages/*/test/**`, so no vitest/CI change is needed.

## Test plan
- **AC1** — all five adapters pass their `@jinn-network/plugin/testing` contract kits in harness-layer's vitest suite with network mocked (injected fake `HarnessLayer`/stub distill; fs on `mkdtemp` temp dirs).
- **AC2** — Evidence adapter reads a legacy `CapturedTask` and up-maps it to a valid `EpisodeV1`, and `put`→`get` round-trips a schema-canonical `EpisodeV1`.
- **AC3** — no existing module edited (diff is only `package.json` + `yarn.lock` + new `adapters/`/`test/adapters/` files); the full pre-existing harness-layer + client suite stays green.
- Reviews: independent code review (APPROVE-WITH-NITS; both SHOULD-FIX items fixed — dead ledger projection dropped, round-trip doc corrected) and `/security-review` (PASS; the one latent LOW path-traversal note fixed with a boundary guard + regression test).

Closes #1660


ritsuKai2000:
Review verdict: **request-changes.** Four of five adapters are clean (dependency direction one-way, no behavior change, unified one-span-sequence trajectory, byte-exact round-trip). But the central surface — the `CapturedTask → EpisodeV1` up-map — has a Critical, silent mis-typing bug that CI does not catch because the fixtures encode the same wrong convention as the code.

**Critical — `stepKind()` mis-types every real capture step** (`client/packages/harness-layer/src/adapters/evidence-adapter.ts:88-92`). It labels a step `jinn.agent_turn` only if `attributes['role'] === 'user'|'assistant'` or `name === 'turn'` — but **no real `CapturedTask` step uses a bare `role`**. The actual producers (`client/src/cli/commands/capture.ts:107,327-343`, `client/src/captures/distil-export.ts`) set `name = 'jinn.transcript.<kind>'` and `attributes['jinn.capture.event.kind'] ∈ {user-message, assistant-message, tool-call, tool-result, edit}`. So **every** step — including user/assistant turns — resolves to `jinn.tool_call`; a zero-tool conversation up-maps to an all-`jinn.tool_call` episode, the semantic inverse. Schema-valid (both kinds legal) → silent. Defeats the point of the DR-2026-07-14 unified trajectory.
- **Fix:** derive kind from the source convention, matching this repo's own reader `client/packages/harness-layer/src/distill-traces.ts:101-105` — `eventKind = attributes['jinn.capture.event.kind'] ?? name`; treat `user-message`/`assistant-message` (or `name` prefix `jinn.transcript.user-message`/`assistant-message`) as `jinn.agent_turn`, else `jinn.tool_call`. Then swap the synthetic `makeLegacyCapturedTask()` fixture (`test/adapters/evidence-adapter.test.ts:842-876`) for a real-capture-shaped one — it will fail against the current code, proving the bug.

**Important — Condition A (attribute-key convention) unresolved.** `stepKind` reads the bare `role` convention while `attributes` (line 66) passes the namespaced keys (`jinn.capture.event.kind`, `message.content`, `tool.name`, `tool.args`, `tool.result`) through verbatim — so the discriminant and the attributes speak different languages, and an up-mapped `jinn.agent_turn` would NOT carry the `role`/`content` that `packages/plugin/src/schemas/episode.ts:22-26` documents. Pick one convention, normalize on write, and document it (the DR + existing readers use the namespaced keys — recommend normalizing to those).

**Minor:** `corpus-adapter.ts:57` hardcodes `kind:'trace'` (skill refs would mislabel; only null path tested) — deferrable with the existing consume.ts follow-up note.

Do not merge until `stepKind` reads `jinn.capture.event.kind` and the fixtures use real capture shape. The trajectory *structure* is correct; only the `kind` derivation + attribute convention need the fix.

Issue #815, opened 2026-05-28
Default 10% reference-ceiling throttle for paid harnesses (M1 cost-protection UX)


oaksprout:
**Context.** The Jinn daemon should self-cap so any credential (`provider:authMethod`) never burns more than **100 AI units per 6h UTC-aligned block** (00:00 / 06:00 / 12:00 / 18:00) — where one AI unit is calibrated so that 100 units = 10% of a GPT-5.4-mini Codex Plus 6h-block equivalent. The same 100-unit cap applies to every harness/model. A Haiku 4.5 run that costs ~1 AI unit debits 1 unit; an Opus 4.7 run that costs ~50 AI units debits 50 units; once the credential's running 6h-block sum exceeds 100, claims for that credential pause until the next block boundary. A 7d UTC-aligned cap layers on top as the weekly safety net.

The mechanism: at claim time, the daemon converts the run's estimated cost into AI units (USD/GPT-5.4-mini-USD-per-block × 100), sums per-credential over the current 6h block + the current 7d window, and skips the claim if either sum + the projected debit would exceed the matching cap. The activity row is written at claim time with the AI-unit estimate and updated at completion with actuals as telemetry — estimates are the gate input, never recomputed.

Onboarding and harness/model selection surface the projected per-block + per-week AI-unit footprint upfront so operators commit to a 48h burn-in knowing exactly what their node will consume.

This **subsumes #346** as the live M1 cost-protection workstream. PR #345 shipped per-task USD surfacing + the $1/task gate; this issue layers the AI-unit ceiling on top, reusing `estimateModelCost`.

**Impact.** P0. Operators on Claude Max / Codex Plus subscriptions currently can't commit to a 48h burn-in because they fear losing their subscription quota to Jinn. Directly gates M1 (#605), whose calibration is the matching "~10% of a baseline AI subscription's weekly cap" framing.

**Acceptance criteria.**
- [ ] A `REFERENCE_CEILING = { units_per_block: 100, units_per_week }` constant is defined in `client/`, with `units_per_block` representing 10% of a GPT-5.4-mini 6h-block equivalent and `units_per_week` representing 10% of a GPT-5.4-mini weekly equivalent.
- [ ] A `GPT_5_4_MINI_USD_PER_BLOCK` constant is defined as the conversion key between any model's USD cost and AI units.
[...]

PR #1722, opened 2026-07-15
feat(815): post-merge follow-ups for AI-units ceiling

### LINKED ISSUE #830: feat(815): post-merge follow-ups for AI-units ceiling
Follow-up bundle from the post-merge review of #815 (AI-units ceiling). Each item is small enough to land independently; bundled here to avoid issue sprawl.

### 1. `weekResetsAt` should reflect the sliding window, not a fixed wall-clock instant

The status surface reports `weekResetsAt` as `now + 7d` (`weekResetsAtUtc`). For a rolling 7-day window this is misleading — the cap doesn't reset all at once; the oldest unit in the window simply falls off. Either:
- Compute the instant when the *binding* unit (the one that pushes the credential over the cap) will age out of the 7d window — that's the true "claims resume at" instant.
- Or rename the field to make the semantics explicit (e.g. `weekWindowEndsAt`).

Related: `client/src/spend/ai-units.ts` `weekResetsAtUtc`, status payload schema, `AiUnitsPauseAlert`.

### 2. UI window picker can disagree with the gate

`AiUnitsPauseAlert` picks the binding window via `unitsThisBlock >= capPerBlock`, but the gate considers the *projected* claim against the cap. The UI's picker can therefore claim the block window is binding when the gate paused on the week window (or vice versa) in edge cases. Surface the binding window directly from the gate decision via `/v1/status` so the UI doesn't re-derive it.

### 3. Worked-example regression test for Haiku model in MODEL_COST_TABLE

The cost-table tests cover GPT-5.4 mini (the peg) and a handful of others but not Haiku. Add a small worked-example assertion in `test/spend/ai-units.test.ts` to catch peg drift.

## Notes

These are nits, not blockers. Filed from the #815 post-merge review to keep them visible.

ritsukai:
## Summary
Three small, independently-landable follow-ups from the #815 post-merge review (AI-units ceiling):

1. **`weekResetsAt` reflects the rolling window, not a fixed `now + 7d`.** New `Store.weekWindowResumeAt()` computes the true "claims resume at" instant — the moment the oldest in-window rows age out far enough that the 7d spend sum drops back under the weekly cap (binding row's `ts + 7d`). `gather-status.ts` uses it when the credential is paused on the week window; falls back to the coarse `now + 7d` otherwise. Field name unchanged (semantics corrected) to avoid SPA/test churn.
2. **The UI binding-window picker can no longer disagree with the gate.** New `pausedWindow: 'block' | 'week' | null` field on the `/v1/status` AI-units row, computed with the daemon gate's block-preferred precedence (`overBlock ? 'block' : overWeek ? 'week' : null`). `AiUnitsPauseAlert` reads it directly (`row.pausedWindow ?? local-derivation` for daemon backcompat) instead of re-deriving from `unitsThisBlock >= capPerBlock`.
3. **Haiku worked-example regression test** in `test/spend/ai-units.test.ts` pinning `estimateModelCost('claude-haiku-4-5-20251001').usd ≈ 0.15` to catch cost-table peg/rate drift.

Reviewed independently (approve, no blockers) and security-reviewed (no findings). `/simplify` applied (shared `SEVEN_DAY_MS`/`AiUnitsPausedWindow` reuse, gated the extra query behind `overWeek`).

Two structural findings surfaced during review were deliberately deferred as out-of-scope for this Low-effort follow-up: (a) extracting the block-preferred precedence into a shared predicate reused by both `gather-status.ts` and `daemon/ai-units-gate.ts`; (b) importing `AiUnitsPausedWindow` into the SPA to replace its inlined union. Worth a future issue.

## Test plan
- New store unit tests (`test/store/ai-units-week-resume.test.ts`): under-cap → null; single oldest row over cap → `ts+7d`; multi-row requiring multiple drops.
- Updated `AiUnitsPauseAlert.test.tsx`: gate-sourced `pausedWindow: 'week'` wins over a local `unitsThisBlock >= capPerBlock` derivation (the disagreement case).
- New Haiku worked-example assertion.
- `yarn typecheck`, full `yarn test`, and `yarn build` green locally.

Closes #830

### DIFF
--- a/client/src/api/gather-status.ts
+++ b/client/src/api/gather-status.ts
@@ -25,6 +25,7 @@ import type { FleetState } from '../earning/types.js';
 import { displayFleetServiceIndex } from '../earning/fleet-display-index.js';
 import {
   assembleStatusV1,
+  type AiUnitsPausedWindow,
   type GatheredStatusRaw,
   type ServiceBalanceErrorEntry,
   type StatusV1Response,
@@ -906,14 +907,16 @@ export async function gatherStatusForApi(
         // Block-preferred precedence — mirrors the daemon gate
         // (`src/daemon/ai-units-gate.ts`): when both windows are over, the
         // binding window is the block (issue #830, item 2).
-        const pausedWindow: 'block' | 'week' | null = overBlock ? 'block' : overWeek ? 'week' : null;
+        const pausedWindow: AiUnitsPausedWindow = overBlock ? 'block' : overWeek ? 'week' : null;
         // The true "claims resume at" instant for the week window (issue
-        // #830, item 1) — null when the week window is under cap, in which
-        // case the fixed `now + 7d` fallback is used since the value isn't
-        // operator-relevant.
-        const weekResetsAt =
-          store.weekWindowResumeAt(credentialId, aiUnitsCfg.capPerWeekUsdMicros, now) ??
-          weekResetsAtFallback;
+        // #830, item 1) — only computed when actually paused on the week
+        // window; otherwise the fixed `now + 7d` fallback is used since the
+        // value isn't operator-relevant (and querying it would cost an
+        // unconditional per-credential row scan on every status poll).
+        const weekResetsAt = overWeek
+          ? (store.weekWindowResumeAt(credentialId, aiUnitsCfg.capPerWeekUsdMicros, now) ??
+            weekResetsAtFallback)
+          : weekResetsAtFallback;
         return {
           credentialId,
           // #1006: legacy unit fields, derived from USD via the peg.
--- a/client/src/api/gather-status.ts
+++ b/client/src/api/gather-status.ts
@@ -888,7 +888,7 @@ export async function gatherStatusForApi(
   if (aiUnitsCfg) {
     const now = new Date();
     const blockResetsAt = blockResetsAtUtc(now).toISOString();
-    const weekResetsAt = weekResetsAtUtc(now).toISOString();
+    const weekResetsAtFallback = weekResetsAtUtc(now).toISOString();
     // Dedupe credentials across manifests — multiple SolverNets on the same
     // credential share one row.
     const uniqueCredentials = new Set(Object.values(aiUnitsCfg.manifestCredentials));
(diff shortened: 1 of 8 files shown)


ritsuKai2000:
Merge-batch checkpoint (2026-07-15): conflict-free rebase from `9b76dda6` onto current `origin/next` `32cf314d`; new head `e0b33881`. Both commits are `=` in range-diff. Focused operator tests passed 39/39; client typecheck and build passed under Node 22.22.2 after an immutable dependency refresh. Exact force-with-lease push succeeded. Fresh CI and mandatory non-author CODEOWNERS approval are required before draft conversion.

ritsuKai2000 (commented):
## Engine review — clean, no blocking findings (human code-owner approval required)

Independent review across three passes (code-review, security, app-test). This PR touches code-owned paths (`client/src/dashboard/spa/`, `client/src/api`, `client/src/store`), so per DR-2026-06-03 the engine posts a comment and hands off — a human code owner must approve and merge.

### Code review — no blockers
- **Filter consistency (the main risk):** `Store.weekWindowResumeAt`'s SQL sums the *same* row set and cost column as `usdMicrosThisWeek → sumUsdMicros` — identical `COALESCE(actual, estimated, 0)`, `claim_status IN ('claimed','delivered')`, same `weekStart` (`SEVEN_DAY_MS`). The only added predicate is `ts < now`, and every claim/deliver row is `ts`-stamped at write time (past relative to a later poll), so no counted row is excluded. `overWeek == true` therefore cannot silently fall through to the coarse `now + 7d`.
- **Precedence:** `overBlock ? 'block' : overWeek ? 'week' : null` mirrors the daemon gate (`ai-units-gate.ts` prefers block when both windows are over cap).
- **`throw 'unreachable'`:** genuinely unreachable — the loop is entered only when `remaining >= cap`, `remaining` strictly decreases to `0 < cap`, so a return always fires (positive cap from the peg). The strict `< cap` correctly matches the pause predicate.
- **SPA fallback:** `row.pausedWindow ?? (unitsThisBlock >= capPerBlock ? 'block' : 'week')` — a current daemon only emits `null` for unpaused rows, which are filtered out before the map, so `??` fires only on `undefined` from pre-field daemons. Correct backcompat.
- **Tests** are real: multi-row-drop path with two caps asserting exact ISO instants; the SPA disagreement case (`unitsThisBlock >= capPerBlock` but `pausedWindow: 'week'` → must show 7d, not 6h); the Haiku cost pin. 26/26 client tests + 7/7 SPA tests pass locally; typecheck clean.

### Security — no findings
Parameterized `better-sqlite3` bind params throughout (no interpolation); `credentialId` is env-resolved operator config; `/v1/status` is a pre-existing unauthenticated read route and this adds no new auth or disclosure surface. The row-scan is gated behind `overWeek` and bounded to operator-local 7-day data.

### App-test — no regression
Rendering verified for `pausedWindow: 'week'` / `'block'` / absent; SPA suite green; sole consumer (`Overview.tsx`) passes `status.aiUnits` straight through.

### Advisory only (non-blocking, do not gate merge)
- `weekWindowResumeAt` does a `reduce` then re-walk (double pass) — trivial for a 7-day per-credential window.
- A non-positive week cap (`cap <= 0`) would reach the `throw` rather than degrade gracefully; no evidence such config is producible today. An early `cap <= 0 → return null` guard would be more defensive.
- `Overview.tsx`'s inline `aiUnits` type mirror omits `pausedWindow` (and already omits `active`); harmless given the field is optional and the runtime payload carries it, but could be tightened.
- The two structural items the PR body already deferred (shared precedence predicate; importing `AiUnitsPausedWindow` into the SPA) are reasonable future-issue material.

Engine verdict: **clean.** Awaiting human code-owner approval + merge.

ritsukai:
Autopilot parked this item for Human review.

PR label: review:needs-human

PR #1728, opened 2026-07-15
feat(hermes): make provider first-class in model catalog and joined SolverNet config

### LINKED ISSUE #1243: feat(hermes): make provider first-class in model catalog and joined SolverNet config
Context: follow-up implementation issue from the Hermes provider catalog schema investigation (#295) and PR #1231.

The spike found that Hermes is provider-pluralistic, but `HERMES_MODELS` and joined SolverNet config currently persist only `model`. The v0.1.6 regex inference patch should be replaced by a first-class provider route that is stored per joined SolverNet.

Scope:
- Add a shared `ProviderRef` type/schema: string for named providers, object form for custom endpoints (`name`, optional `baseUrl`, optional `authVar`).
- Add `provider: 'openrouter'` to current `HERMES_MODELS` entries.
- Thread provider through JoinFlow request, join endpoint validation/persistence, `joinedSolverNets` config, `SolverNetConfig`, `TaskSessionInputs`, and engine stamping.
- Update the Hermes adapter to prefer per-task provider, then daemon global provider, then legacy inference only as a migration bridge.
- Add a load-time migration/backfill that stamps `provider: 'openrouter'` onto legacy joined SolverNet entries with OpenRouter-shaped model IDs.
- Remove runtime reliance on `inferProviderFromModelId` after migration coverage is in place.

Acceptance criteria:
- A non-OpenRouter Hermes catalog entry routes through its declared provider.
- A v0.1.6 `{ model }`-only joined SolverNet config backfills to `openrouter` without breaking load.
- A custom provider object produces the expected Hermes config/base URL/env behavior.
- Tests prevent shipping a Hermes catalog entry with no provider route.

Related: #295, #293, PR #1231.

ritsukai:
## Summary
Makes `provider` a first-class route alongside `model` across the Hermes harness stack, replacing the v0.1.6 runtime `inferProviderFromModelId` patch (which only worked because every shipped catalog entry was OpenRouter-routed).

- New shared `ProviderRef` type (`client/src/harnesses/provider-ref.ts`, zero-dep, no zod, imported by both server and SPA): string named provider, or `{ name, baseUrl?, authVar? }` for custom endpoints. Also owns the shared `OPENROUTER_MODEL_FORMAT` regex + `isOpenRouterModelId` (deduped from adapter/config).
- `provider: 'openrouter'` stamped on all `HERMES_MODELS` catalog entries; new `providerForModel()`.
- Provider threaded end-to-end: JoinFlow request → join endpoint validation/persist → `joinedSolverNets` config (zod union) → `registerJoinedNet` / `SolverNetConfig` / `LoadedSolverNet` → engine `TaskSessionInputs` stamping → Hermes adapter.
- Adapter resolution precedence: local `hermesBaseUrl` → per-task `inputs.provider` → daemon-global `hermesProvider` → legacy `inferProviderFromModelId`. Object form writes Hermes `model.base_url` and injects `authVar` into the spawned child env (with a warn when the named var is unset).
- Load-time backfill stamps `provider: 'openrouter'` onto legacy `{ model }`-only joined entries with OpenRouter-shaped model ids (covers real + synthetic legacy-migrated entries), so existing configs migrate silently.

**Scope note:** `inferProviderFromModelId` is intentionally retained as the documented final migration bridge — it still covers the hand-typed top-level `config.hermesModel` (no `hermesProvider`) path, which the joinedSolverNets backfill does not reach. Full removal is deferred until that path is also migrated. The custom-provider object form is reachable via the join API / config (catalog entries are all OpenRouter today); no new join-UI control was added.

## Test plan
- `provider-ref` unit tests (accessors + regex predicate).
- Catalog guard: every `HERMES_MODELS` entry must declare a provider (fails the build if one is dropped).
- Config: `{ model }`-only joined entry loads via `loadConfig` and backfills `openrouter`; non-OpenRouter ids and already-set providers untouched.
- Adapter: per-task provider string routes `--provider <name>`; object form writes `base_url` + injects a non-allowlisted `authVar`; precedence (per-task > daemon-global > inference); warn on unset `authVar`.
- Join endpoint: string + object provider validated and persisted; malformed rejected.
- SPA `JoinFlow`: join request carries `provider: 'openrouter'` for a Hermes model; omits it for Claude Code.

Closes #1243

### DIFF
--- a/client/src/api/setup-endpoints.ts
+++ b/client/src/api/setup-endpoints.ts
@@ -660,6 +660,7 @@ export function addSetupRoutes(app: Hono, config: SetupRoutesConfig = {}): void 
       roles?: unknown;
       harness?: unknown;
       model?: unknown;
+      provider?: unknown;
       plugins?: unknown;
       disabledDefaultPlugins?: unknown;
       contract?: unknown;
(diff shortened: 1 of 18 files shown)


ritsukai:
Autopilot parked this item for Human review.

PR label: review:needs-human

PR #1750, opened 2026-07-15
feat(plugin): render canonical contribution states

ritsuKai2000:
## Product outcome

`/jinn ledger` now renders the real shared contribution store rather than a synthetic or process-local projection. A captured session can be inspected through recorded, preview-required, minted, queued, published, vetoed, or rejected states even when the daemon sidecar is absent.

## User-facing state table

| Canonical state | User label | Public refs shown |
|---|---|---|
| recorded | recorded | none |
| preview-required | preview required | none |
| minted | minted | none |
| queued | queued | none |
| published | tier chip | mint + publication refs |
| vetoed | vetoed (local only) | none |
| rejected | rejected (local) | none |

## Boundaries and privacy

- Adds `JinnPlugin.contributionLedger()` and `jinn-layer contribution ledger --json`.
- Projects only timestamp, public repository/base commit, tier, derived state, and published refs.
- Never exposes local record/source IDs, trajectory, accepted diff/gold, failures, skills, or holdouts.
- Keeps the legacy receipt ledger as an old-binary fallback only.

## Verification

- Plugin: typecheck, 22 files / 114 tests, build.
- Harness layer: typecheck plus 3 focused files / 28 tests.
- Python plugin: 4 files / 65 tests.
- Real shared-store integration covers session-end write, daemon-readable record, preview acknowledgement, all state transitions, absent sidecar, and privacy projection.

Closes #1664

### DIFF
--- a/apps/jinn-agent/plugins/jinn/__init__.py
+++ b/apps/jinn-agent/plugins/jinn/__init__.py
@@ -572,9 +572,31 @@ def _handle_jinn(command_args: str = "", session_id: str = "", task_id: str = ""
         return f"preview failed:\n{out}"
 
     if sub == "ledger":
-        # Prefer structured rows (design 1b columns + tier chips + retry
-        # sub-line + exact empty state). Degrade to the layer's raw text when
-        # the installed layer predates `ledger --json`.
+        # Prefer canonical contribution-store rows. A layer that predates the
+        # command falls back to the legacy publication-receipt ledger.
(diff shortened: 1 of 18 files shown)


ritsuKai2000:
Self-review complete: no actionable findings. I checked the canonical store join and derived-state precedence, old-binary fallback, unavailable/degraded behavior, published-only ref exposure, and the terminal projection boundary. The focused cross-language tests cover every state and assert that local IDs, diffs, trajectory, failures, skills, and holdouts are absent from JSON output.

PR #1769, opened 2026-07-16
design(jinn-plugin): Stage 1 rescope — evidence retrieval and the complete contribution loop (R0)

ritsuKai2000:
## Run-mode

design — see the handbook SOP for the design shape.

## What

R0 of the Stage 1 rescope. The 2026-07-16 #1654 walkthrough recorded **iterate**: the capture/contribution half passed; the consumption half drifted into the Stage 3 skills surface (crude search terms, duplicated seed skill records, metadata-only injection). This PR:

- adds the rescope plan: `docs/superpowers/plans/2026-07-16-jinn-plugin-stage-1-rescope-plan.md` (corrected product definition, audit matrix, retrieval design, seed/acceptance design, work units R0–R7, acceptance gate);
- amends the Stage 1 product design (§4.2/§4.5/§7/§8/§9) to evidence retrieval with `searched → provided` states;
- amends the completion plan's knowledge-state bullets and status;
- adds package-architecture §11a (SkillsPort quarantined later-stage; content-bearing `CorpusPort.get`; shipped-API reconciliation) — **Closes #1755**;
- adds the harness-network §7 Stage 1 note (evidence seeds required; skill seeds excluded from Stage 1 pickup; seed idempotency).

## Why

Restores alignment with the roadmap (skills are Stage 3) and `spec/2026-07-06-distillation-v1.md` D11 (retrieval over anchored evidence is the ratified v1 product baseline). Stage 1 must close cleanly so Stage 2 (the refactor stage) can start.

Implementation lands as sub-issues R1–R7 of #1654 (filed alongside this PR). Refs #1654.

Human-surface docs PR — human review required; no auto-merge.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1780, opened 2026-07-16
feat(plugin): evidence-first pickup core + content adapters (Stage 1 rescope R1+R2)

ritsuKai2000:
## Summary

Implements R1 and R2 of the [2026-07-16 Jinn Plugin Stage 1 rescope](https://github.com/Jinn-Network/mono/blob/claude/jinn-plugin-stage-1-rescope-315d0e/docs/superpowers/plans/2026-07-16-jinn-plugin-stage-1-rescope-plan.md) (plan PR #1769, tracking #1654): rebuilds the auto-pickup retrieval policy once in the product core and wires the real harness-layer adapters to serve full evidence content instead of metadata.

**R1 — `packages/plugin` (the core)**
- `deriveSearchTerms(firstMessage, repositorySlug?)` replaces the 2-token first-line heuristic with up to 6 deterministic lowercase terms, priority-ordered: backticked/quoted tokens → identifier-shaped tokens (`_-./`, digits, camelCase) → the session's repository slug → longest remaining non-stopword tokens. Operates over the whole message.
- New selection policy (`selectKnowledgeHits`): drops every skill-classified hit, dedups by ref and by `(taskSummary, origin)` content key, scores by matched-term count over summary+tags (repo-slug match counts 2), applies a relevance floor of 2 (else honest nothing-found), ranks score → tier → recency, takes the top 2.
- New `jinn.knowledge-packet.v1` schema + `projectKnowledgePacket(record, budget)`: a pure, deterministic projection selecting failure/fix/command/diff/note excerpts from a record's steps — never paraphrases. Enforces a 3,500-char per-packet budget (two packets ≤ 7,000 total) with line-boundary-aware truncation ending in an explicit `corpus_fetch <ref>` pointer.
- `CorpusPort.get()` now returns a content-bearing `CorpusRecord` (task summary, outcome, synthesis, steps, tags, provenance, origin, capturedAt) instead of a metadata-only `KnowledgeHit`. `InMemoryCorpusPort` and the port contract kit updated accordingly.
- `firstTurnPickup` returns `{ contextBlock: string | null, packets: KnowledgePacket[], searchedTerms: string[], degraded?: string }` and composes the injected `[jinn corpus] Prior evidence relevant to this task:` block from real packet content. The skill-suggest/install/auto-adopt path is deleted entirely from pickup — **closes #1729 by removal** (the auto-adopt honesty gap it was filed against no longer exists). `SkillsPort` itself stays in the package, quarantined for the local-distill install path.
- `Episode.activity` gains `searchedTerms`/`providedRefs`; the pre-rescope `surfacedRefs`/`fetchedRefs`/`installedSkillRefs` remain in the schema (defaulted) so old episode files and hosts that haven't migrated still parse. `SessionSummary` gains `searchedTerms`/`providedPackets` and keeps the legacy fields actively populated (a new-shape `providedRefs` wins; a pre-rescope caller's own `surfacedRefs` echoes through unchanged) for downstream harness-layer/cli/Python compatibility until the host switch (R3) lands. History counts fall back to the legacy fields for episodes captured before this change.

**R2 — `client/packages/harness-layer` (real adapters + process contract)**
- `corpus-adapter.ts`'s `get()` decodes and sha256-verifies the record's `jinn.trace-envelope.v0` artifact — mirroring the Python `skills_install._extract_trace` path — and returns the full `CorpusRecord`; a hash mismatch degrades rather than serving unverified content. `search()`'s `KnowledgeHit` mapping now carries `tags`, `origin` (on-chain `agentId` preferred over the forgeable `safeAddress`, matching `consume.ts`'s existing `resolveHeads` precedent), and `publishedAt`.
- `process-contract.ts`'s `SessionMetaSchema` gains an optional `repositorySlug` so the request can carry it end to end (additive; `contractVersion` stays 1) — completing the plumbing for R1's `SessionMeta.repositorySlug`.
- `evidence-adapter.ts` now enforces the declared newest-200 episode retention (`EvidenceRetentionPolicy.maxEpisodes` was read but never acted on): `put()` prunes the oldest-by-`capturedAt` `.episode.json` files above the configured cap after every successful write.
- `process-contract.test.ts` rewritten for the additive `session pickup` response shape and to prove pickup never touches the skills port now that the adopt/install path is gone.

**The Python host path is deliberately untouched.** `apps/jinn-agent/plugins/jinn/pickup.py` today does its own term-derivation/selection/rendering directly against the low-level `corpus search`/`corpus get` CLI verbs — it does not call `session pickup` at all, so this PR's changes are additive/parallel, not yet wired into the live product surface. The blocking `cold-stock-e2e` gate (which exercises the unmodified Python path) stays green. Switching the host to delegate to `session pickup` is R3, a separate issue.

Closes #1770
Closes #1772
Closes #1729
Refs #1654

## Test plan
- [x] `cd packages/plugin && yarn install && yarn typecheck && yarn test && yarn build` — 159/159 tests, clean typecheck, clean build
- [x] `cd client && yarn install && yarn typecheck` — clean (builds `packages/plugin` + `packages/sdk` + harness-layer as part of the pipeline)
- [x] `cd client && yarn build` — clean, produces `dist/bin/jinn-layer.js`; smoke-checked `node dist/bin/jinn-layer.js contract --json` → `{"contractVersion":1}`
- [x] `cd client && yarn vitest run packages/harness-layer/test/contribution-store.test.ts packages/harness-layer/test/process-contract.test.ts test/solver-types/task-creator-session-echo.test.ts` (the exact CI-gated set) — 60/60
- [x] `cd client && yarn vitest run packages/harness-layer/test/` (full suite) — 621/621
- [ ] `cold-stock-e2e.sh` full run — needs network; left to CI per instructions


ritsukai (changes requested):
Recovered independent review, revalidated after patch-equivalent rebase to head 73c19b37a969075c4d87cbf74cab390581b0a240, found blocking correctness issues:

1. Enforce the 3,500-character knowledge-packet budget even when remaining capacity is smaller than the truncation tail; add tiny-capacity boundary tests.
2. `foldExplain`/`SessionExplanation` must expose canonical `searchedTerms`/`providedRefs` with legacy fallback.
3. Corpus content dedup must not trust forgeable `safeAddress`; use trustworthy record identity/agent identity and keep safeAddress only for display attribution.
4. Evidence pruning needs a deterministic filename tie-break for equal timestamps.
5. Explicitly prove evaluator-verified skill records remain excluded.
6. Correct transitional comments and stale schema citations.

Please fix test-first and request re-review on the new head.

ritsuKai2000:
## Independent review record

Reviewed by an independent agent (not the author) per the operator's delegation for this train. **Verdict: approve-with-nits — no blockers.** Reviewer independently reproduced every claimed test count (plugin 159/159; harness-layer 621/621; CI-gated set 60/60; client typecheck/build clean; contract v1 unchanged), verified design fidelity to the rescope plan §3 (term priority order, skill exclusion, ref+content-key dedup, relevance floor ≥2 tested at the boundary, score→tier→recency, sha-256 retention in the adapter, retention prune), and confirmed the #1729 closure is structural (adopt/install path deleted; a throwing SkillsPort leaves pickup ok).

**Findings (3 should-fix, 4 nits) — carried into the R3+R5 PR** (the implementing agent's fix pass was interrupted by a session limit; findings are non-blocking and several intersect R3+R5's scope anyway; recorded on #1773):
1. Knowledge-packet char budget can overshoot by ≤~94 chars when the truncation tail doesn't fit the remaining budget (skip-the-excerpt fix + boundary test).
2. `foldExplain`/`SessionExplanation` not migrated to `searchedTerms`/`providedRefs` (latent — no live callers).
3. SessionSummary legacy fields kept vs AC "drop" — **reviewer-approved deviation**: keep until the host flips, drop in R3+R5.
4. Dedup content-key should use agentId only (not the forgeable safeAddress); soften the resolveHeads comment.
5. Comment the unconsulted `autoAdopt`/`autoAdoptTier`/`maxCandidates` config fields.
6. Stale §4.2 citation in knowledge-hit.ts header.
7. Add a tier-pinned skill-exclusion test (evaluator-verified skill still excluded) + filename tie-break on the retention sort.

All CI green including the blocking Cold-stock Stage 1 product gate (Python path deliberately untouched until R3).

PR #1788, opened 2026-07-17
fix(plugin): strip trailing punctuation from search terms; parse jinn-layer stdout only (R6 walkthrough defects)

ritsuKai2000:
## Summary

Fixes the two defects the R6 live walkthrough (#1775) found, plus the gate's fixture-recognition blind spot that let #1786 ship green.

- **#1786** — `cleanWord` (`packages/plugin/src/pickup.ts`) kept `_-./` anywhere in a token, including trailing, so any sentence-final prose word (`done.`, `flag.`, `else.`) read as identifier-shaped and took a priority-2 slot ahead of real content words in the 6-term search budget.
- **#1787** — `_default_runner` (`apps/jinn-agent/plugins/jinn/jinn_layer.py`) merged the child process's stdout and stderr into one string before returning it, so any stderr diagnostic (e.g. the evidence-adapter's "skipping malformed legacy capture" warning) corrupted an otherwise-valid stdout JSON response before `json.loads`. `/jinn history` was dead on any machine carrying a legacy capture the current schema rejects; the rescope had additionally routed `session pickup` / `session end` through the same strict parser.
- **Gate blind spot** — `stage1-stock-product.py`'s scenario-1 message happened to carry exactly one sentence-final period, so it passed the gate regardless of the #1786 bug. Replaced with representative multi-sentence prose that depends on the fix.

## Fix 1 — #1786 term hygiene

`cleanWord` now strips **leading and trailing** `_-./` while preserving internal separators. A leading `/` is stripped too (`/v1/status` → `v1/status`) — deliberate: it stays a useful, greppable term, and treating every separator position the same way keeps the rule one rule instead of a special case for paths.

Before/after on the walkthrough's real prose message (built `dist/pickup.js`, `repositorySlug='Jinn-Network/mono'`):

```
BEFORE: ["ai-units","flag.","one.","done.","jinn-network/mono","distinguishes"]
AFTER:  ["ai-units","jinn-network/mono","distinguishes","dashboard","estimated","operator"]
```

`dashboard` — the one term that matches the seeded evidence record's tags — now survives instead of being evicted by three punctuation artifacts.

`Say OK.` → `[]` (was `["ok."]`).

On the `2+2` fold (non-kept chars dropped, fragments joined into `22`): left as-is per the issue's own "do not gold-plate" guidance — it's orthogonal to the leading/trailing-strip fix (`+` was never a kept separator, before or after) and splitting instead of joining is a separate, larger change to `cleanWord`'s character-filtering model.

## Fix 2 — #1787 stream separation

`_default_runner` now returns `(code, stdout, stderr)` instead of `(code, merged)`. JSON parsing consumes stdout only; the FileNotFoundError/timeout synthetic failures put their message on stderr so the install-hint guidance path keeps working (verified: `test_jinn_layer_not_found_points_at_canary_tag`).

This changed the shared `run()`/`session_pickup()` contract, so I inspected and updated every caller (~24 call sites across `plugins/jinn/__init__.py`, `pickup.py`, `onboarding.py`, `skills_install.py`, `distill.py`, `hermes_cli/banner.py`, plus the test doubles in `test_jinn_layer.py` / `test_jinn_layer_verbs.py` / `test_jinn_plugin.py` that monkeypatch the runner). A custom `Runner` (test doubles, `jinn._runner` overrides) still returns the simple `(code, out)` 2-tuple it always did — `_normalize_result` pads it to `(code, out, "")` so every caller sees one consistent 3-tuple regardless of which runner produced it. The handful of failure paths that render diagnostic text now prefer `err`, falling back to `out` only when `err` is empty (protects existing tests/behavior where a test double never had a stderr channel to begin with).

Regression coverage (a runner stub emitting stderr + valid stdout, the way the walkthrough hit it):
- `test_default_runner_separates_stdout_and_stderr_so_json_parsing_stays_clean` — a real subprocess (`python -c ...`) writing `[evidence] skipping malformed legacy capture ...` to stderr and a valid v1 envelope to stdout; `_default_runner` reports them separately and `parse_process_response` parses the stdout only.
- `test_default_runner_still_reports_stderr_on_a_real_non_zero_exit` / `test_run_normalizes_a_two_tuple_injected_runner_to_a_three_tuple` — round out the contract.
- `test_jinn_history_renders_normally_when_stderr_carries_a_warning`, `test_pickup_parses_normally_when_stderr_carries_a_warning`, `test_delegate_parses_normally_when_stderr_carries_a_warning` — the three AC-named surfaces (`/jinn history`, `session pickup`, `session end`) each proven against a fake `_default_runner` with the same stderr-warning/stdout-envelope shape.

A true end-to-end repro (real `jinn-layer` binary + a real legacy capture file on disk) is what the R6 walkthrough did manually and what `cold-stock-e2e.sh` exercises in CI; it's out of reach for a fast Python unit test, so the above is the closest cheap equivalent — same mechanism (a real subprocess, real separate streams), synthetic content.

## Fix 3 — the gate's blind spot

New scenario-1 message (`apps/jinn-agent/scripts/stage1-stock-product.py`):

> The dashboard vitest suite is flaky again. It keeps racing the version-status check and timing out. Track down the root cause and confirm the fix holds. Report back once the suite is green.

Four sentences, four sentence-final periods (`again.`, `out.`, `holds.`, `green.`) — representative prose, not a single run-on sentence with one incidental artifact.

Verified empirically against the real built `deriveSearchTerms`/`scoreKnowledgeHit` (`packages/plugin/dist`, `repositorySlug='acme/widget'` — the gate's actual `work_repo` git remote, not a stand-in):

| | derived terms | source score | distractor (same-repo) | distractor (other-domain) |
|---|---|---|---|---|
| **with the #1786 fix** | `["version-status","acme/widget","dashboard","confirm","vitest","racing"]` | **3** (floor is 2) | 0 | 0 |
| **pre-#1786-fix `cleanWord`** (checked out in isolation, rebuilt, tested, not committed) | `["again.","version-status","out.","holds.","green.","acme/widget"]` | **1** (< floor) | 0 | 0 |

Under the pre-fix `cleanWord`, the four punctuation artifacts plus `version-status` plus the repository slug fill the 6-term budget entirely in the identifier-shaped pass — `dashboard` never reaches the remainder bucket. Source drops from 3 to 1, below the relevance floor, and the gate's `assert pickup is not None` / `assert target_context is not None` would fail. The gate now depends on the #1786 fix instead of merely coexisting with the bug.

Both skill distractors (`distractor-skill-tdd`, `distractor-skill-tdd-dup`) are excluded from selection categorically (`selectKnowledgeHits` drops every skill-kind hit before scoring), independent of message wording.

No 4+ consecutive-word run is shared with the source fixture's `taskSummary` (`"Fix flaky dashboard update_available banner test — assert after the version status fetch resolves"`) or `synthesis` — checked programmatically (4-gram set intersection over both texts): empty.

`cold-stock-e2e.sh` was **not** run locally (network) per instructions — reasoned through every assertion in the `finish_session(... message=TARGET_TASK_MESSAGE ...)` block instead; none depend on the message's specific wording, only on which record gets selected (source, confirmed above) and structural boundaries (skill-install language absent, sharing-off POST boundary, etc.), all of which are unaffected by the wording change. `stage1-task-creator-acceptance.mjs` (outside this PR's boundary) only reads the structural `stock-product-result.json` output (episode/session/record IDs), not the message text.

## Verification (all green)

- `packages/plugin`: `yarn install && yarn typecheck && yarn build && yarn test` — **171 tests, 23 files, all pass**. Typecheck clean.
- `client`: `yarn install && yarn typecheck` — clean. Harness-layer vitest suite (`packages/harness-layer/test/`) — **736 tests, 51 files, all pass**.
- `apps/jinn-agent`: full `test_jinn_*.py` set via `scripts/run_tests.sh` (per-file isolation) — **349 tests, 33 files, all pass**. `ruff check` clean on every touched file.
- Manually verified (git-stash isolation, rebuilt, not committed) that both the walkthrough repro and the new gate message score as claimed under the pre-#1786-fix code, confirming the regression-canary property.

Closes #1786
Closes #1787
Refs #1654
Refs #1775

### DIFF
--- a/apps/jinn-agent/hermes_cli/banner.py
+++ b/apps/jinn-agent/hermes_cli/banner.py
@@ -918,7 +918,7 @@ def _read_contribution_count() -> Optional[int]:
     """
     try:
         from plugins.jinn import jinn_layer, ledger_view
-        code, out = jinn_layer.ledger_json()
+        code, out, _err = jinn_layer.ledger_json()
         if code != 0:
             return None
         rows = ledger_view.rows_from_json(json.loads(out))
@@ -940,7 +940,7 @@ def _read_corpus() -> Dict[str, object]:
     """
     try:
         from plugins.jinn import jinn_layer
-        code, out = jinn_layer.corpus_search("", limit=500, as_json=True)
+        code, out, _err = jinn_layer.corpus_search("", limit=500, as_json=True)
         if code == 0:
             hits = json.loads(out)
             if isinstance(hits, list):
--- a/apps/jinn-agent/plugins/jinn/__init__.py
+++ b/apps/jinn-agent/plugins/jinn/__init__.py
@@ -84,7 +84,7 @@ def _check_contract() -> None:
             return
         _contract_checked = True
         try:
-            code, out = jinn_layer.contract(runner=_runner)
+            code, out, _err = jinn_layer.contract(runner=_runner)
         except Exception:
             _degraded = "jinn-layer handshake failed"
             return
@@ -387,7 +387,7 @@ def _delegate_session_end(
     if contribution_candidate is not None:
         request["contributionCandidate"] = contribution_candidate
     try:
-        code, out = jinn_layer.session_end(request, runner=_runner)
+        code, out, _err = jinn_layer.session_end(request, runner=_runner)
         if code != 0:
             return None
         envelope = jinn_layer.parse_session_end_response(out)
@@ -621,7 +621,7 @@ def _handle_jinn(command_args: str = "", session_id: str = "", task_id: str = ""
         if action == "decline":
             if not confirmed:
                 return consent.confirm_decline_command()
-            code, out = jinn_layer.contribution_disable(runner=_runner)
+            code, out, err = jinn_layer.contribution_disable(runner=_runner)
             # Let the core serialize revocation with any publication already
             # crossing its boundary. The direct marker is the fail-closed host
             # fallback and mirrors the successful core write for old stubs.
@@ -631,11 +631,11 @@ def _handle_jinn(command_args: str = "", session_id: str = "", task_id: str = ""
                 return message
             if marker_ok:
                 return message + "\nexisting publication queue is fail-closed locally; jinn-layer cleanup is degraded"
-            return message + f"\nwarning: could not disable the existing publication queue: {out}"
+            return message + f"\nwarning: could not disable the existing publication queue: {err or out}"
         return consent.render_explainer(consent.COMMANDS_LINE)
 
     if sub == "preview":
-        code, out = jinn_layer.contribution_preview(acknowledge=True, runner=_runner)
+        code, out, _err = jinn_layer.contribution_preview(acknowledge=True, runner=_runner)
         if code == 0:
             try:
                 reply = jinn_layer.parse_process_response(out)
@@ -662,11 +662,11 @@ def _handle_jinn(command_args: str = "", session_id: str = "", task_id: str = ""
             # an empty screen. Does not mark previewed — the real gate stays
             # on a real trace.
             return consent.render_preview_example()
-        code, out = jinn_layer.capture_preview(pending, runner=_runner)
+        code, out, err = jinn_layer.capture_preview(pending, runner=_runner)
         if code == 0:
             consent.mark_previewed()
             return out + "\n\nlegacy preview only — this retained file will not auto-publish."
-        return f"preview failed:\n{out}"
+        return f"preview failed:\n{err or out}"
 
     if sub == "session":
         state = _peek_state(session_id)
@@ -677,9 +677,9 @@ def _handle_jinn(command_args: str = "", session_id: str = "", task_id: str = ""
         )
 
     if sub == "history":
-        code, out = jinn_layer.history_json(runner=_runner)
+        code, out, err = jinn_layer.history_json(runner=_runner)
         if code != 0:
-            return f"history unavailable:\n{out}"
+            return f"history unavailable:\n{err or out}"
         try:
             reply = jinn_layer.parse_process_response(out)
         except ValueError as exc:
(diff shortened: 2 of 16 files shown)


ritsuKai2000:
## Independent review record

Reviewed by an independent agent (not the author) per the operator's delegation for this train. **Verdict: approve** — zero blockers, zero should-fix.

Everything was reproduced from source rather than trusted from the description:
- **Fix 1**: the walkthrough's exact prose now derives `["ai-units","jinn-network/mono","distinguishes","dashboard","estimated","operator"]` — `dashboard` survives, no `word.` artifacts. Internal shapes preserved (`update_available`, `version-status`, `client/src/dashboard`, `AI-units`). Edge sweep clean and Unicode-safe (`e.g.`→`e.g`, `--flag`→`flag`, `co-op.`→`co-op`, `...`→`[]`, `café-menu.`→`café-menu`), no empty-string terms or crash paths. The `/v1/status`→`v1/status` call was judged correct: multi-segment paths keep identifier shape, and `scoreKnowledgeHit`'s substring match still hits a corpus tag spelled `/v1/status`.
- **Fix 2**: whole-tree grep (not just diff files) confirms all ~24 unpack sites on the 3-tuple; test doubles in seven untouched files still return 2-tuples and pass via `_normalize_result` padding. Every discarded-`err` site traced individually — all pre-existing never-surfaced paths; every user-facing site keeps `err or out`. Reproduced end-to-end through a **real** `subprocess` boundary with a fake `jinn-layer` writing a stderr warning + valid stdout: `/jinn history`, `session pickup`, `session end` all parse post-fix; the pre-fix runner body (extracted from `7e6a729ae`) genuinely fails with `JSONDecodeError: Extra data`.
- **Fix 3**: the canary is real — reconstructing the actual seed fixtures and scoring with the built functions gives source **3** post-fix vs **1** pre-fix (below the floor of 2), so `selectKnowledgeHits` returns empty and the gate's assertion fails if #1786 regresses. Zero n-gram overlap with the source fixture at 3/4/5-grams (stronger than the required 4+). No gate assertion depends on the message wording; the no-result and 503 scenarios are unperturbed.
- Suites verified independently: plugin 171/23, harness-layer 736/51, Python 349/33 (plus the 2008-test superset), ruff clean. CI green on HEAD `ca6b77650` including the blocking Cold-stock Stage 1 product gate.

Nit carried forward, not blocking (filed as a follow-up): `deriveSearchTerms`' quoted/backtick-span path (`pickup.ts:64-67`) bypasses `cleanWord`, so `` `npm test.` `` still yields the literal term `npm test.` — the same class as #1786 in an adjacent code path, outside this issue's stated AC.

Issue #1006, opened 2026-06-03
Operator dashboard AI-units surfaces show "units", not USD actual spend / estimated-vs-metered


oaksprout:
**Context.** Issue #1004 switches the daemon's claim-gate accumulator and ceiling from projected "AI units" to USD *actual* token spend, and adds an `estimated` flag to the `/v1/status` payload (true when no observed-usage rows back the figure, e.g. the Hermes harness). The operator dashboard still renders the gate figure in "units" (`AiUnitsPauseAlert`, `HarnessFootprintPanel`, `Overview`) and presents every figure as metered. After #1004 lands, those labels mislabel a USD-denominated, sometimes-estimated value.

**Impact.** Operators read a misleading unit and cannot tell a metered figure (backed by harvested token usage) from an estimated one (heuristic projection). The dashboard would contradict the gate it visualises.

**Acceptance criteria.**
- [ ] After a claim settles, the AI-units block on Overview displays the accumulator and ceiling in USD, not "units".
- [ ] `AiUnitsPauseAlert` and `HarnessFootprintPanel` display USD, consistent with the `/v1/status` USD fields.
- [ ] When the `/v1/status` `estimated` flag is true for a credential, that figure is labelled "estimated"; when false, it is labelled "metered" (or equivalent).
- [ ] No regression: figures remain correct after a page reload.

**Files/components.** `client/src/dashboard/spa/` — `AiUnitsPauseAlert.tsx`, `configuration/HarnessFootprintPanel.tsx`, `Overview.tsx`, `JoinedNetCard.tsx` (prop types); consumes the `/v1/status` `aiUnits` block (`client/src/api/status-build.ts`, `client/src/api/gather-status.ts`); `client/OPERATOR-APP-SPEC.md`. Prerequisite: #1004.

## Human-surface change

**Domain-model delta.** `client/OPERATOR-APP-SPEC.md`, AI-units credential component.
- *Static (State):* unit of the accumulator/ceiling changes units → USD; add `estimated: boolean`.
- *Streams:* the `/v1/status` `aiUnits` row gains USD fields + `estimated` (delivered by #1004).
- *Actions:* none added or changed.
- *State messages:* the existing "paused — AI-units ceiling reached" §2.10 banner is reworded to USD; add an "estimated figure" qualifier where `estimated` is true. Empty / loading / error states unchanged.

**Design artifact.** None yet. A design pass is required before implementation (hence `Blocked on: Human`).

[...]

PR #1793, opened 2026-07-17
fix(plugin): lexical-v2 term selection — message-order terms, path segments, repo-name vote, plural fold (R6 retrieval)

### LINKED ISSUE #1790: Pickup: the repository-slug term is dead weight — it can never match, yet consumes 1 of 6 term slots and its +2 bonus never fires
## Context
Found in the R6 live walkthrough re-run (2026-07-17) against the live testnet corpus, on `next @ 99eb16af4` (i.e. after #1786 was fixed).

The rescope design (`docs/superpowers/plans/2026-07-16-jinn-plugin-stage-1-rescope-plan.md` §3.3) specifies the session's repository slug as a search term and says scoring counts "a repository-slug match counting 2" — intended to boost same-repo evidence. It was implemented faithfully (`packages/plugin/src/pickup.ts`, `deriveSearchTerms` step 3 + `scoreKnowledgeHit`'s `term === repoTerm ? 2 : 1`). **The design is wrong: neither half can ever fire.**

## Evidence (live corpus, real seeded record)
Terms derived for a real dashboard task: `["ai-units","jinn-network/mono","distinguishes","dashboard","estimated","operator"]`.

Scored against the real seeded evidence record (`bafkreie32jr5og…`, tags `seed-import, mono, dashboard, vitest, version-status, async, flake`):

| term | matches? |
|---|---|
| `jinn-network/mono` | **no** |
| `dashboard` | yes (+1) |
| all others | no |

→ **score 1**, below the floor of 2. And:

```
record text contains "jinn-network/mono": false
```

Two failures, one cause — **a record's text never contains its full `owner/repo` slug**:
1. **As a search term** it is dead: the indexer matches `haystack.includes(needle)` over `taskSummary + tags`; no record carries the literal `jinn-network/mono`, so this term returns nothing from every query. It still costs one of the six slots (~15% of the budget), evicting a term that could match.
2. **As a scoring bonus** it is dead: `scoreKnowledgeHit` only awards the +2 when `haystack.includes(term) && term === repoTerm` — the first condition is never true, so same-repo evidence receives no boost at all. `KnowledgeHit` carries no structured repo field, so there is nothing else to compare against.

Net effect: same-repo evidence is *harder* to retrieve than the design intends, because the slug both fails to boost and steals a slot.

## Options (pick one; the first is minimal)
- Derive the **repo name** (the segment after `/`, e.g. `mono`) as the term instead of the full slug — it matches the seed lane's own `mono` tag, so both the search and the +2 fire as designed. Cheap, but a short repo name may be a noisy substring (`mono` matches `monorepo`, `monotonic`).
- Add a structured `repositorySlug` to `KnowledgeHit`/`CorpusRecord` (plumbed from the record's own metadata) and award the +2 on a structured compare rather than a text match; drop the slug as a search term entirely, freeing a slot.
- Drop the repo-slug concept from §3.3 and let plain term overlap decide.

The second is the honest one and matches "the record knows its own repo"; the third is the smallest. Whichever is chosen, **update rescope plan §3.3** so the design and the code agree.

## Acceptance criteria
- The repo slug either measurably contributes (test: the real seeded record scores ≥2 for a same-repo task sharing one vocabulary word) or is removed from term derivation (test: six slots go to terms that can match).
- Plan §3.3 amended to match.
- A test pins the chosen behavior against a record shaped like the real seed (tags carrying the short repo name, no full slug anywhere in the text).

Refs #1654, #1770, #1786. Found by the R6 walkthrough (#1775).

ritsuKai2000:
## Summary

Faithful implementation of the ratified "lexical v2" design from #1791's decision comment (empirically validated by the issue author against the three real Stage 1 seed fixtures and five representative messages before landing). Closes three related term-selection defects in `packages/plugin/src/pickup.ts` as one cohesive change:

- **#1791** — term selection optimized for length/identifier-shape, not retrievability. A tight 6-term budget, longest-first remainder ordering, and over-specific path tokens starved real, on-topic tasks below the relevance floor.
- **#1790** — the repository-slug term was dead weight: it could never match a record's text (no record's text contains the literal `owner/repo` string), so it both stole a term slot and its `+2` scoring bonus never fired.
- **#1789** — quoted/backtick spans bypassed `cleanWord`, keeping trailing punctuation (e.g. `` `npm test.` `` kept its period and could never match).

### What changed in `deriveSearchTerms` / `scoreKnowledgeHit` / `selectKnowledgeHits`

1. `MAX_TERMS` 6 → 10.
2. Remainder bucket is now **message order** (order of first appearance), not longest-first.
3. Any identifier-shaped token containing `/` also contributes its **path segments** (cleaned, ≥3 chars, stopword-filtered, deduplicated), right after the full token.
4. The repository's **name** (segment after the last `/` of `repositorySlug`, ≥3 chars) replaces the full slug as a normal term. The `term === repoTerm ? 2 : 1` special case in `scoreKnowledgeHit` is deleted — every matched term counts 1. The now-dead `repositorySlug` parameter is removed from `scoreKnowledgeHit`/`selectKnowledgeHits` (`plugin.ts` still passes `repositorySlug` to `deriveSearchTerms`, which is where it now does all its work).
5. **Plural fold** at match time: a term matches verbatim, or — for a term ending in a simple plural `s` with length > 3 — its singular form matches.
6. Quoted/backtick spans now run through `cleanWord`'s leading/trailing `_-./` strip before being pushed. `cleanWord` gains space to its character whitelist so this is safe on multi-word spans (the two existing single-word call sites never see a raw space, so this is a no-op there).
7. Floor stays 2; top-2 selection, skill exclusion, and dedup are unchanged.

### `plugin.ts`

No behavior change to the wiring intent: `firstTurnPickup` still passes `repositorySlug` to `deriveSearchTerms` (derivation). The now-dead 3rd argument to `selectKnowledgeHits` (the old repo-bonus plumbing) is removed, since the repo-name signal now flows entirely through the derived `terms` array like any other term.

### Tests

`packages/plugin/test/pickup.test.ts` rebuilds the five-message × three-fixture validation table from #1791's decision comment as unit tests, using hit shapes that mirror the real Stage 1 seed fixtures verbatim (`client/packages/harness-layer/fixtures/stage1-seeds/*.episode.json` — `taskSummary` → `snippet`, `tags` → `tags`, the only two fields `haystackFor()` reads):

- Run B (the real on-topic walkthrough task, verbatim from #1791) → source clears the floor and is selected; the same-repo distractor scores repo-name-only (1, excluded); the other-domain distractor scores 0.
- Run A (#1006, dashboard USD labels — reconstructed from the exact pre-lexical-v2 derived-terms fingerprint recorded verbatim in #1654's R6 walkthrough #3 comment, and now cross-verified against the real fixtures) → both source and the same-repo distractor clear the floor, distractor ranks first — pinned as documented, accepted behavior citing #1791's accepted-trade-off paragraph.
- The gate's own no-result message (verbatim) and a constructed same-repo/zero-content-overlap control → nothing clears the floor in either case, pinning #1791's AC ("a same-repo record with zero content overlap does not clear the floor").
- New `deriveSearchTerms` unit tests: message-order remainder (a 5-char early word beats a 13-char later one), path segmentation (`client/src/dashboard/spa/src` → full token + `dashboard`/`spa`/`client`/`src`, `src` included at exactly the 3-char minimum), repo-name derivation (`Jinn-Network/mono` → `mono`, short names <3 chars excluded, no-slash slugs used whole), plural fold (`tests` matches `test`; `its` does not fold), quoted-span stripping (`` `npm test.` `` → `npm test`, no trailing period; multi-word and path-shaped spans keep their shape).
- Existing tests that pinned the old `+2` repo bonus, the old 6-term cap, or longest-first order are rewritten to pin the new behavior, citing the issue numbers. Every other existing test stays green **unmodified** — notably the existing `#1786` term-hygiene test using the real Run A message text needed no changes at all (its loose `toContain`/no-trailing-period assertions hold under both the old and new algorithm).

### Reproduced validation table (built `dist/pickup.js`, `repositorySlug = 'Jinn-Network/mono'`)

Matches #1791's decision comment **exactly**:

| message | source | D1 operator-claims | D2 sympy | selected (top-2, ranked) |
|---|---|---|---|---|
| Run B (flaky notification tests — the real on-topic task) | **3** | 1 | 0 | source-dashboard-flake |
| Run A (USD labels) | 2 | **3** | 0 | distractor-operator-claims, source-dashboard-flake |
| gate scenario-1 | **5** | 1 | 0 | source-dashboard-flake |
| gate no-result | 1 | 1 | 0 | (none) |
| unrelated control (same-repo/zero-content AC) | 1 | 1 | 0 | (none) |

### Gate-margin verification (`apps/jinn-agent/scripts/stage1-stock-product.py`) — driver **not** edited

The decision-comment table above (and the pinned unit tests) use `repositorySlug = 'Jinn-Network/mono'` — the real corpus's actual repository, matching the issue author's validation methodology. The **live gate driver** itself derives `repositorySlug` from its fixture git repo's remote (`https://github.com/acme/widget.git` → `acme/widget`, via `session_bridge._repository_slug`/`snapshot_repository` in `apps/jinn-agent/plugins/jinn/session_bridge.py:75-108`, exercised end-to-end by `apps/jinn-agent/tests/plugins/test_jinn_session_evidence.py::test_snapshot_and_accepted_diff_cover_every_change_without_mutation`, which asserts a `git@github.com:Jinn-Network/example.git` remote yields `repository_slug == "Jinn-Network/example"`), which is unrelated to the corpus fixtures' real repo. Computed with that **actual** driver-path slug:

| message | source | D1 | D2 | floor |
|---|---|---|---|---|
| gate scenario-1 (`TARGET_TASK_MESSAGE`) | **4** | 0 | 0 | 2 |
| gate no-result (`NO_RESULT_MESSAGE`) | 0 | 0 | 0 | 2 |

Scenario-1 clears the floor with a margin of 2 and is the sole candidate above it (distractors both at 0); the no-result message stays at 0 against every fixture. Both driver assertions (`expected_pickup=True` for scenario 1, `expected_pickup=False` for the no-result scenario) hold under the new scorer. Per the task's instructions, since both hold, **the driver is not edited** — this section is that stated verification. (Note: `TARGET_TASK_MESSAGE`'s inline comment block in the driver, lines 68–85, still describes the pre-lexical-v2 score of 3; it is now stale documentation, not a functional problem, and was left alone per the "do not edit the driver" instruction — worth a follow-up doc-only touch-up if desired.)

### Verification

- `cd packages/plugin && yarn install && yarn typecheck && yarn build && yarn test` — 186/186 tests green, 23 test files.
- `cd client && yarn install && yarn typecheck` — green.
- `cd client && npx vitest run packages/harness-layer/test` (after `yarn build:sdk && yarn build:plugin && yarn build:harness-layer-store`, so the adapters compile against the freshly built plugin) — 736/736 tests green, 51 test files.
- Decision table and gate-margin numbers above reproduced against the built `packages/plugin/dist/pickup.js`.

Closes #1791
Closes #1790
Closes #1789
Refs #1654
Refs #1775
Refs #1792

### DIFF
--- a/docs/superpowers/plans/2026-07-16-jinn-plugin-stage-1-rescope-plan.md
+++ b/docs/superpowers/plans/2026-07-16-jinn-plugin-stage-1-rescope-plan.md
@@ -187,11 +187,18 @@ one `corpus_fetch` away; nothing is model-generated in the hot path.
 
 ### 3.3 Search and ranking
 
-- **Terms:** `deriveSearchTerms(firstMessage, repositorySlug?)` replaces the 2-token heuristic.
-  Up to 6 terms, in priority order: backticked/quoted tokens; identifier-shaped tokens (contains
-  `_`, `-`, `.`, `/`, a digit, or CamelCase); the session's repository slug (known at session
-  start from `session_bridge.snapshot_repository`); remaining longest non-stopword tokens.
-  Whole first message, not just line one. Deterministic.
+- **Terms (lexical v2 — #1791, #1790, #1789):** `deriveSearchTerms(firstMessage, repositorySlug?)`
+  replaces the 2-token heuristic. Up to 10 terms, in priority order: backticked/quoted tokens
+  (edge-stripped through the same leading/trailing `_-./` strip as other terms, #1789);
+  identifier-shaped tokens (contains `_`, `-`, `.`, `/`, a digit, or CamelCase) — a `/`-bearing
+  token also contributes its path segments (each cleaned, ≥3 chars, stopword-filtered,
+  deduplicated), right after the full token; the session repository's **name** — the segment
+  after the last `/` of `repositorySlug` (known at session start from
+  `session_bridge.snapshot_repository`), ≥3 chars — not the full slug, which can never appear in
+  a record's text (#1790); the remaining non-stopword tokens (≥4 chars) in **message order**
+  (order of first appearance, not longest-first — length is not a retrievability signal against a
+  corpus of short summaries and tags, #1791). Whole first message, not just line one.
+  Deterministic.
 - **Search:** per-term `CorpusPort.search` (capture-meta first, manifest floor fallback — existing
   rails), results merged.
 - **Selection:**
@@ -200,10 +207,17 @@ one `corpus_fetch` away; nothing is model-generated in the hot path.
      they are simply not what pickup serves.)
   2. Dedup by ref **and** by content key `(taskSummary, origin)` — kills the duplicated-seed
      symptom at the consumer even before store hygiene lands.
-  3. Score = count of matched terms across `taskSummary + tags`, with a repository-slug match
-     counting 2. **Relevance floor: score ≥ 2**, otherwise honest nothing-found. Tier is a
-     ranking preference, not a filter: sort score desc → tier desc (`TIER_ORDER`) → recency desc.
+  3. Score = count of matched terms across `taskSummary + tags`; **every match counts 1** — the
+     old repository-slug +2 bonus is gone (#1790, #1791) now that the repository's name is a
+     normal derived term. A term matches verbatim or, for a term ending in a simple plural `s`
+     (length > 3), by its singular form (plural fold, #1791). **Relevance floor: score ≥ 2**,
+     otherwise honest nothing-found. Tier is a ranking preference, not a filter: sort score desc
+     → tier desc (`TIER_ORDER`) → recency desc.
   4. Take the top 2; fetch full content; project packets.
+
+Escalation ladder if lexical v2 proves insufficient: lexical v2 (above) → two-stage content
+re-scoring for near-miss candidates against synthesis/steps (#1792, deterministic, model-free) →
+embeddings only with Stage 2 evidence.
 
 ### 3.4 Injection
 
--- a/packages/plugin/src/pickup.ts
+++ b/packages/plugin/src/pickup.ts
@@ -1,6 +1,7 @@
-/** Evidence-first pickup policy (rescope design §3.3). No I/O: term
- *  derivation + selection only. Orchestration (search/get over ports,
- *  packet projection, rendering) lives in plugin.ts. */
+/** Evidence-first pickup policy (rescope design §3.3; lexical v2 term
+ *  selection — #1791, #1790, #1789). No I/O: term derivation + selection
+ *  only. Orchestration (search/get over ports, packet projection, rendering)
+ *  lives in plugin.ts. */
 import type { KnowledgeHit } from './schemas/knowledge-hit.js';
 import { TIER_ORDER } from './schemas/pickup-config.js';
 
@@ -13,8 +14,14 @@ export const STOPWORDS: ReadonlySet<string> = new Set([
   'using', 'about', 'have', 'has', 'had', 'you', 'your', 'our', 'not',
 ]);
 
-const MAX_TERMS = 6;
+// Lexical v2 (#1791): a tight 6-term budget forced real trade-offs between
+// terms that could plausibly match a record's summary/tags — 10 gives
+// selection room to include more of the message before the floor decides
+// relevance. Search is one cheap indexer call per term.
+const MAX_TERMS = 10;
 const MIN_REMAINDER_LENGTH = 4;
+const MIN_PATH_SEGMENT_LENGTH = 3;
+const MIN_REPO_NAME_LENGTH = 3;
 
 const QUOTED_SPAN_RE = /`([^`]+)`|"([^"]+)"|'([^']+)'/g;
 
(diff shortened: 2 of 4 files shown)


ritsuKai2000:
## Independent review record

Reviewed by an independent agent (not the author) per the operator's delegation for this train. **Verdict: approve-with-nits** — zero blockers.

Faithfulness to the #1791 decision spec verified file:line on every item (budget 10, message-order remainder, path segments, repo-name +1 with the +2 case deleted, plural-fold formula byte-exact, quoted-span strip, floor/top-2/selection blocks byte-identical to base). Both above-spec implementer choices judged sound: the space in `cleanWord`'s keep-set is a no-op for whitespace-split tokens (verified by reading both call sites) and makes multi-word quoted spans genuinely matchable (empirically confirmed against the real seed haystack); the dropped `selectKnowledgeHits` parameter has no dangling caller repo-wide. The validation table was reproduced from the built dist against the real fixture JSON files — exact match on all five rows — and the gate-margin claim was re-derived including hand-tracing `session_bridge.py`'s slug derivation against the driver's `make_repo()` (`acme/widget` confirmed; source 4, distractors 0). Suites independently reproduced: plugin 186/186, harness-layer 736/736; all CI green including the blocking cold-stock gate.

Worth carrying forward (informational): under lexical v2, Run B's retrieval succeeds via `dashboard` + `mono` + plural-fold `tests` — the term `flaky` is still evicted by the budget (path segments + repo name consume 6 slots). The fix's robustness rests on redundancy across three signals, not on any single term surviving — noted on #1792 as context for the content-rescore step. Nits (non-blocking): `pathSegments`' stopword branch lacks a dedicated unit test; multi-space quoted spans degrade to silently-unmatchable terms; the gate driver's inline score comment is stale (self-disclosed in the PR body, doc-only follow-up).

Issue #1394, opened 2026-07-03
feat(harness-layer): first-class skill artifact type with provenance


oaksprout:
Part of #1392 (learning-loop epic), stage: knowledge representation.

**Context.** Envelopes today carry trajectory captures (jinn.capture-trajectory.v1), harness bundles, and solver-type payloads. Spec/2026-07-02 §5 defines Layer 2 — consumable skills in SKILL.md-compatible form with provenance links back to evidence — but no skill artifactType exists; the seed import publishes skills as synthetic capture traces, and /jinn skills install works by convention on that shape.

**Impact.** Skills are second-class: not distinguishable in queries, no provenance to the evidence that motivated them, no surface for the verification gate (Layer 1 → Layer 2 promotion) to act on.

**Acceptance criteria.**
- [ ] A skill artifactType (e.g. jinn.skill.v1) in the envelope schema: SKILL.md content (+ companion files, cf. #1381), machine-readable provenance (source envelope CIDs, operator, solverType) and the standard access/pricing fields.
- [ ] jinn-layer publish + corpus get round-trip it; /capture-meta indexes its tags/summary; the harness skills-install path recognises it (backwards-compatible with the seeded shape).
- [ ] Schema change reviewed against the frozen envelope-v0 caps (additive).
Entry points: client/src/types/envelope.ts, client/packages/harness-layer/ (capture/publish), client/src/captures/publish.ts.

PR #1794, opened 2026-07-17
fix(plugin): exclude content-classified skill records and empty packets from pickup (#1782)

ritsuKai2000:
## Summary

Evidence-first pickup's kind filter excludes wire `kind: 'skill'` hits, but a legacy pre-#1394 seed-imported skill record (SKILL.md as a step attribute, no first-class `jinn.skill.v1` artifact) classifies as `kind: 'trace'` on the real indexer wire shape and slips the filter. This manifested live on 2026-07-17 (R6 walkthrough, `next @ 1ae580dd3`): a natural on-topic message returned two packets — the seeded evidence record, and the legacy RunComfy skill seed, which also projected empty (zero excerpts, no synthesis — a skill payload has no seed steps), wasting one of the two injection slots.

Two composable guards, both applied after a candidate's content is fetched (per the issue's ratified scope):

1. **Content-level skill classification.** The corpus adapter (`client/packages/harness-layer/src/adapters/corpus-adapter.ts`) now derives `isSkillPayload` on `CorpusRecord` — true when the record carries a `jinn.skill.v1` artifact, or any step carries a string `skill.md` attribute — independent of the wire `kind`. The core (`packages/plugin/src/plugin.ts` `firstTurnPickup`) enforces exclusion, keeping the adapter the fact provider and the core the policy owner.
2. **Empty-packet honesty.** A projection with zero excerpts and no synthesis is dropped rather than occupying a packet slot.

Both guards need to promote the next-ranked candidate when they disqualify the current one, which `selectKnowledgeHits` (pre-sliced to `MAX_SELECTED_PACKETS`) can't support. Added `rankKnowledgeHits` — the full ranked-but-unsliced candidate pool; `selectKnowledgeHits` is now its top-slice convenience wrapper, unchanged in behavior/tests. `firstTurnPickup` walks the ranked list, applying both guards, until it has `MAX_SELECTED_PACKETS` valid packets or candidates are exhausted.

Both guards fail open: a detection or projection error degrades that one record to the pre-guard behavior, never a crash.

## Test plan

- `mono #1782 pin: ...` (`packages/plugin/test/plugin/first-turn-pickup.test.ts`) reproduces the live escalation's exact "two packets" symptom (fails pre-fix returning both `bafyCheckoutEvidence` and `bafyRunComfySkillLeak`; passes post-fix returning only the real evidence) — **this is the pinning test**.
- `mono #1782 guard 1: a legacy skill.md-attribute record that would otherwise project a non-empty packet is still excluded by content (isolates guard 1 from guard 2)` — proves guard 1 is independently necessary, not just piggybacking on guard 2's empty-packet exclusion (added after independent review flagged the original pin as guard-2-confoundable for that fixture shape).
- `mono #1782 guard 1: a first-class jinn.skill.v1-backed record is excluded by content even when the wire kind says trace` — the defensive/first-class variant from the issue's AC.
- `mono #1782 guard 2: a zero-excerpt, no-synthesis packet is dropped and the next above-floor candidate is promoted into its slot` + the "nothing to promote" honest-nothing-found variant.
- `mono #1782: promotion skips past multiple consecutive disqualified candidates (guard 1 then guard 2) to fill both slots` — 4-candidate compound scenario (added after independent review).
- `client/packages/harness-layer/test/adapters/corpus-adapter.test.ts` — new `describe('CorpusAdapter.get() — content-level skill classification (mono #1782)')` block: legacy step-attribute detection (including on a step not named `seed:skill-md`), first-class artifact detection, and two negative/regression cases (ordinary trace record, empty/non-string `skill.md` attribute).
- All new tests independently verified to fail pre-fix and pass post-fix (via `git checkout HEAD~1 -- <src files>` against the test files, then restored) — not just asserted.
- Regression: all pre-existing pickup/adapter tests (seeded evidence fixtures projecting/injecting as before) pass unmodified.
- Both guards' fail-open behavior verified via code inspection and try/catch coverage in the adapter's detection path.

Verification run:
- [x] `cd packages/plugin && yarn install && yarn typecheck && yarn build && yarn test` — 198/198 passing.
- [x] `cd client && yarn install && yarn typecheck` — clean (builds sdk/plugin/harness-layer-store, then `tsc --noEmit` across the client tree plus a dedicated `tsc -p packages/harness-layer/tsconfig.json`).
- [x] `cd client && yarn vitest run packages/harness-layer/` — 741/741 passing (18/18 in `corpus-adapter.test.ts` specifically).
- [x] Independent review (fresh subagent, read-only checkout, independently re-verified TDD red/green on the real production decode path) — verdict: ready to merge; two test-precision recommendations, both since implemented in the second commit.

Closes #1782
Refs #1654
Refs #1775

### DIFF
--- a/client/packages/harness-layer/src/adapters/corpus-adapter.ts
+++ b/client/packages/harness-layer/src/adapters/corpus-adapter.ts
@@ -18,8 +18,9 @@ import {
   type HarnessLayer,
   type HarnessLayerConfig,
 } from '../consume.js';
-import { parseTraceEnvelopeV0, type TraceStep } from '../envelope.js';
+import { parseTraceEnvelopeV0, type TraceEnvelopeV0, type TraceStep } from '../envelope.js';
 import { TRACE_ENVELOPE_ARTIFACT_TYPE } from '../publish.js';
+import { SKILL_ARTIFACT_TYPE } from '../../../../src/types/skill-artifact.js';
 
 export interface CorpusAdapterDeps {
   /** A ready HarnessLayer (test seam: a fake `{ corpus: { search, get } }`). */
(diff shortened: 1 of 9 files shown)


ritsuKai2000:
## Review record

Two layers of review on this PR: the implementing agent commissioned its own independent read-only reviewer (verdict: ready to merge; its two findings — guard-1/guard-2 isolation in the pin test, and multi-candidate promotion coverage — were fixed in 50d125da4 before the PR was reported), and the coordinator's final gate is the live re-probe below. CI green on all checks including the blocking cold-stock gate.

**Live verification (the defect's own scenario, against the live testnet corpus):** the pickup probe that leaked the RunComfy skill seed as an empty second packet on `next @ 1ae580dd3` is re-run post-merge; the merge is confirmed only if it returns exactly the one evidence packet. Result posted on #1654 with the final walkthrough record.

PR #1813, opened 2026-07-17
fix(harness-layer): episode store read-tolerance for null optional fields

oaksprout:
## Summary
- Real `EpisodeV1` files on disk carry literal JSON `null` in optional fields (`outcome.summary`, `cost.tokens`, `cost.usdEstimate`, `lineage`). `EpisodeV1Schema` declares these `.optional()` on `z.strictObject` (absent-OK, null-fatal), so every such file failed parse and `evidence.list()` silently dropped it — `/jinn history` and every `EvidencePort` consumer saw zero episodes.
- Adds `parseEpisodeV1Read` (`packages/plugin/src/schemas/episode.ts`): strips literal `null` from the known optional slots (`outcome.summary`, `cost.tokens`, `cost.usdEstimate`, `lineage`, `lineage.mintRef`, `activity`, `eligibility`) then delegates to the unchanged strict `EpisodeV1Schema.parse` — canonical type unchanged, callers can never observe a null. Exported from the package barrel.
- Switches the evidence adapter's three read sites (`list`, `get`, `classifyExisting`) to the read-tolerant parser. Write paths stay strict: `put()` and `SessionEndRequestV1Schema` still use `EpisodeV1Schema.parse` directly, so writers remain forbidden from emitting nulls.
- `list()` now emits `[evidence] list: N readable, M unreadable episode file(s)` whenever anything was skipped, on top of the existing per-file warn — silent total loss is no longer possible.
- Defense in depth in the Python fallback writer: `write_episode_fallback` (`apps/jinn-agent/plugins/jinn/distill.py`) drops `None`-valued keys from the known-optional slots before `json.dumps`, and normalizes the existing on-disk file through the same strip before its identity check (so a post-upgrade retry over a pre-fix null-bearing file resolves as identical instead of a spurious collision).

## Root cause
The current-source writers cannot produce these nulls: `_build_episode` in `capture_buffer.py` conditionally omits the optional keys entirely (e.g. `cost["tokens"]` is only set `if tokens:`), and the TS core parse-gates writes (`JSON.stringify` drops `undefined`). The 18 affected files were written by an older installed wheel whose episode dict carried explicit `None`s: the strict `SessionEndRequestV1Schema` delegation to the TS core rejected each one, and `write_episode_fallback` then persisted the dict unfiltered via `json.dumps` — minting `null`-bearing files the strict reader could never load again. `mineable-traces.json` kept gaining candidates throughout because `tee_capture` writes the legacy `CapturedTask` shape best-effort BEFORE the strict session-end delegation and independently of its outcome (see `distill.py` `tee_capture`); that legacy path has no strict-schema gate, so trace mining kept succeeding while episode persistence was failing over. A session-end on the current wheel+layer persists via the TS core with no fallback write (AC4). Remaining live-machine verification (fresh Hermes session appearing in `/jinn history` on the dogfooding machine) is tracked as a post-merge dogfood check.

## Acceptance criteria
- AC1 (existing null-bearing files returned by `evidence.list()`): read-tolerant parser + adapter wiring; regression tests use a fixture mirroring the exact on-disk shape.
- AC2 (new sessions still strict-parse): write paths untouched and still strict; existing strict-schema tests unchanged and green.
- AC3 (failures surfaced): per-file warn (existing) + new readable/unreadable summary line, tested.
- AC4 (root cause confirmed): see Root cause above.

## Review notes (follow-ups, non-blocking — from the pipeline's review stages)
- The per-file warn logs the basename; logging the full path would meet AC3 even more literally.
- The seven-slot strip list mirrors the schema's `.optional()` fields but nothing enforces the mirror; a zod shape-introspection lockstep test would future-proof it.
- The null-bearing test fixture is hand-encoded in two test files; the plugin's `./testing` subpath is a natural shared home.
- The Python write-side strip covers the four observed slots only (deliberate, documented); TS read tolerance covers the rest.

## Test plan
- [x] `packages/plugin` vitest — read-tolerant parse round-trips the null-bearing shape through the strict schema; corrupt/unknown-field payloads still throw
- [x] `client` harness-layer vitest — `list()`/`get()` return the on-disk null shape; corrupt file still skipped; summary log line asserted
- [x] `apps/jinn-agent` pytest — fallback writer drops None-valued optional keys; pre-fix null-bearing file treated as identical on retry
- [x] Typecheck green (`packages/plugin`, `client` root)

Closes #1811

### DIFF
--- a/apps/jinn-agent/plugins/jinn/distill.py
+++ b/apps/jinn-agent/plugins/jinn/distill.py
@@ -163,6 +163,33 @@ def tee_capture(task: Dict[str, Any], session_id: str, runner: Optional[Runner] 
         return None
 
 
+def _drop_none_optional_keys(episode: Dict[str, Any]) -> Dict[str, Any]:
+    """Strip None-valued KNOWN optional slots before serializing (mono #1811).
+
+    EpisodeV1's optional fields are absent-OK but null-fatal under the strict
+    zod schema, so a persisted literal ``null`` makes the whole file
+    unreadable. Defence-in-depth, deliberately scoped to the known optional
+    slots — not a generic None sweep, which could mask a producer bug in a
+    required field.
+    """
+    out = dict(episode)
+    outcome = out.get("outcome")
+    if isinstance(outcome, dict) and outcome.get("summary") is None and "summary" in outcome:
+        out["outcome"] = {k: v for k, v in outcome.items() if k != "summary"}
+    cost = out.get("cost")
+    if isinstance(cost, dict):
+        pruned = {
+            k: v
+            for k, v in cost.items()
+            if not (k in ("tokens", "usdEstimate") and v is None)
+        }
+        if pruned != cost:
+            out["cost"] = pruned
+    if "lineage" in out and out["lineage"] is None:
+        del out["lineage"]
+    return out
+
+
 def write_episode_fallback(episode: Dict[str, Any]) -> Optional[Path]:
     """Persist one complete EpisodeV1 when the core did not.
 
@@ -184,7 +211,7 @@ def write_episode_fallback(episode: Dict[str, Any]) -> Optional[Path]:
             else f"episode-{hashlib.sha256(episode_id.encode('utf-8')).hexdigest()}"
         )
         serialized = json.dumps(
-            episode,
+            _drop_none_optional_keys(episode),
             ensure_ascii=False,
             allow_nan=False,
             sort_keys=True,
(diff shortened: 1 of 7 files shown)


ritsuKai2000:
Superseded by the merged evidence-contract and C4 index/repair train: PR #1868 established the additive EpisodeV1 reader normalization and strict write boundary; PR #1869 added readable/unreadable indexing, reindex/repair/rescue, surfaced doctor counts, and real-store operator commands. Current `next` contains those implementations at `4bb1c1a21b9cc8966fa29ba67b3211eca3a676fa`. This old-base branch is now conflicting and must not be independently merged. Issue #1811 remains open only for the shared Human fresh-Hermes-session/history proof; closing this PR does not claim that live step occurred.

PR #1844, opened 2026-07-17
fix(autopilot): run triage from self-contained package

ritsukai:
## Summary

- run the mandatory reality-check from `packages/autopilot`, which is a self-contained Yarn project
- remove the invalid root-workspace invocation that blocked #1822 before implementation
- guard the command contract with a focused skill-content regression test

## Root cause

The repository intentionally has no root `package.json`, but the implementation skill invoked `yarn workspace @jinn-network/autopilot` from the repository root. Yarn therefore could not resolve the workspace and every implementation session stopped at the mandatory fail-loud triage gate.

## Verification

- `yarn test test/implement-issue-skill.test.ts` — 4 passed
- `yarn test` — 62 files, 558 tests passed
- `yarn triage:check 1822` under the Autopilot implementer identity — `classification: clear`
- `yarn typecheck` remains red on unchanged `next` because the self-contained package does not declare `@types/node`; this PR does not alter TypeScript sources or dependencies

Refs #1822

### DIFF
--- a/.claude/skills/implement-issue/SKILL.md
+++ b/.claude/skills/implement-issue/SKILL.md
@@ -72,18 +72,19 @@ Before any worktree is created and before `Status` is flipped to `In Progress`, 
 
 ### Invoke the CLI
 
-Run the reality-check from the primary checkout (not the worktree — the worktree doesn't exist yet):
+Run the reality-check from the self-contained Autopilot package in the primary checkout
+(not the worktree — the worktree doesn't exist yet). The repository root has no Yarn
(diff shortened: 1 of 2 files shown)


ritsuKai2000 (approved):
Independent code and security review complete. The package-local triage invocation preserves fail-loud behavior, quotes the full checkout path, and is guarded against both the obsolete workspace command and unquoted regressions. Focused and full Autopilot tests pass; no blocking or advisory findings remain.

